PR #41, opened 2014-11-28
Blackfire integration

GuGuss:


pjcdawkins:
I have called the command `integration:set` rather than `integration:add` (because it can add an integration or update an existing one)

pjcdawkins:
Both these companies have camel-case names: GitHub and HipChat

PR #54, opened 2015-01-05
Add reference to mods-available

Berdir:
Added a reference on how to get the complete list of supported php extensions.


damz:
That makes sense. Thanks for the addition!

Berdir:
Aw, just noticed that "not be" is in there twice :(

PR #59, opened 2015-01-15
Small corrections + add direct link to github page

OriPekelman:
Added more precise docs


OriPekelman:
Added generic webhook

GuGuss:
Will be handled here: https://github.com/platformsh/platformsh-docs/pull/68

PR #60, opened 2015-01-18
Clarify a MAMP oddity regarding the build path resolution

webkenny:
When you specific the symbolic link for MAMP, it auto-follows the path and substitutes the fully qualified build path. So, when you rebuild, you must then specify it again and restart your web server.

Smart folks will learn how to automate this (I'm still working on how) but for now, let's tell developers about this weird behavior in MAMP.


robertDouglass:
Thanks, Kenny!

pjcdawkins:
@webkenny what exactly do you need to do after rebuilding? just restart the web server? or do you need to edit config somewhere?

webkenny:
That may have been unclear. These are the steps:

- In the MAMP Control Panel, change the "Document root:" under your local project host to @www. This will cause MAMP to refollow the symbolic link to the newly created build folder.
- Restart your web server. MAMP will typically prompt for this after a configuration change is made.

robertDouglass:
Fortunately you only need to do this after updating the .make file, and not after normal code edits.

webkenny:
I'm not sure actually. Given the build folder doesn't contain symlinks to the repository folder, it would appear anytime you make a code edit - you would need to rebuild to have the files picked up by a new build. I could be completely misreading how this all works, however, and that certainly is for another issue somewhere. :)

pjcdawkins:
The build folder does symlink back to the repository folder for "project" or "profile" builds. And it symlinks anything in the `shared` folder to sites/default.

webkenny:
See? I knew I was missing something awesome about this whole thing. :+1:

PR #68, opened 2015-01-30
Document HipChat and GitHub integration.

GuGuss:


integration/webhooks.rst
```diff
@@ -0,0 +1,80 @@
...
+Webhooks
+========
+
+Platform.sh can easily be integrated with external services using webhooks.
+
+With the CLI, you can list all your active integrations:
+
+.. code-block:: console
+
+    $ platform integration:list
```
pjcdawkins: I think we shouldn't include the `$` sign in code examples, it makes it harder to copy/paste

PR #83, opened 2015-03-10
Build hooks should use relative path(s)

pjcdawkins:
From ticket 2401


pjcdawkins:
This also needs to be fixed in platformsh-examples

acervulus:
Created pr on platformsh-examples

PR #84, opened 2015-03-11
Revamp UI settings

GuGuss:


GuGuss:
Should be OK to merge on DEV

PR #97, opened 2015-04-27
document undocumented expires option

inominate:


reference/configuration-files.rst
```diff
@@ -134,6 +134,7 @@ It has a few sub-keys which are:
 * **document_root**: The path relative to the root of the application that is exposed on the web. Typically ``/public`` or ``/web``.
 * **passthru**:  The URL that is used in case of a 404 (*which is the equivalent of the rewrite rules in Drupal*). Typically ``/index.php`` or ``/app.php``.
 * **whitelist**: A list of files (as regular expressions) that may be served.
+* **expires**: The number of seconds whitelisted (static) content should be cached by the browswer. This enables the cache-control and expires headers for static content.
```
Ralt: What's the default value is the key is not there?

inominate: There is no default, the expires directive and resulting headers are left out entirely if this isn't set.

Ralt: Fair enough then

PR #100, opened 2015-05-04
Redis improvements

pjcdawkins:
* Do not recommend `redis.path.inc` (see ticket 3001)
* Recommend setting cache_class_cache_form to persistent storage


pjcdawkins:
Just documenting here - `redis.path.inc` appears to have another problem (apart from our Redis not being persistent): https://www.drupal.org/node/2420137

PR #102, opened 2015-05-04
note whitelist gotcha, explain blacklisting

inominate:
I'm pretty sure this is not the behavior we want, but it is the behavior we have.


GuGuss:
Duplicate as: #110

PR #105, opened 2015-05-19
Fix DNS Make Easy link in going-live.rst

steveoliver:
Simple link fix.


pjcdawkins:
Thanks! Unfortunately reStructuredText doesn't like having the same text in more than one "named hyperlink". If they end with a double underscore `__` instead of the single underscore then it should work: see http://stackoverflow.com/a/14067756

PR #110, opened 2015-05-26
Document whitelist/blacklist key

GuGuss:


damz:
This doesn't sound right: blacklist is not limited to PHP files in any way.

GuGuss:
Would need a rebase also.

GuGuss:
Documented here: https://github.com/platformsh/platformsh-docs/commit/15a922219a8f369e8eee9a0b0d1503b125fd0f8b

Removed the PHP specific mention.

Thanks.

PR #111, opened 2015-05-26
Document "slug already built"

GuGuss:


GuGuss:
Would need a rebase also.

GuGuss:
Committed here: https://github.com/platformsh/platformsh-docs/commit/c0dff8b778289d74200926ce9a2daa50829af16a

Thanks.

PR #113, opened 2015-05-28
Fix php.ini comment sign and language

pjcdawkins:
A customer left in the `#` comment, and ended up with an error whenever php-cli ran.


pjcdawkins:
I'm not sure why every code example has to include a comment with the filename at the top (it seems to be conventional in the docs, but it's confusing to me)

GuGuss:
Would need a rebase also.

pjcdawkins:
Rebased. See #120 regarding the Travis failure

PR #118, opened 2015-06-08
Migrate to GitBook

GuGuss:


Kazanir:
This has been merged into master as well, and thus, deployed. Google search is soon to follow (to allow splitting into multiple books) but the default search is functional as is.

PR #119, opened 2015-06-11
[wip] Redirect /foo to /foo.html

pjcdawkins:


pjcdawkins:
ignore until I understand how this is set up, I'm sure we'd prefer no `.html` suffixes anyway

PR #134, opened 2015-07-02
Simplify the architecture page

damz:


GuGuss:
Let's rework on it and target master.

PR #136, opened 2015-07-06
[WIP] Split the documentation in guides.

GuGuss:
Let's create multiple books with very specific use cases and keep the general documentation an overview and technical reference.


GuGuss:
We'll have a v3 branch for this major revamp.

PR #137, opened 2015-07-08
Master

Cookiz:
How can I configure my routes to be able to have subdomains pointing to the right place ? 
I've got a project in french/english and I want to be able to have www.domain.org > french and en.domain.org > english


pjcdawkins:
Hi @Cookiz can you please open an ticket instead? via https://marketplace.commerceguys.com/platform/support

PR #163, opened 2015-08-10
Add tip to import-files.md on how to rsync directly from server to pl…

badrange:
…atform.sh

I experienced an issue with files having non-ascii-filenames while rsyncing from a server to platform.sh via my mac computer. Thanks a lot to fast & useful support by Damien - here is the solution that I ended up with.

Please feel free to improve on my formatting of the text!


OriPekelman:
@damz Wouldn't a better workaround consist of running `rsync -a --iconv=utf-8-mac,utf-8 ....`

badrange:
@OriPekelman You are right in my case - that solution would also give us mac users a working local copy of the site. 

Perhaps you can show both workarounds, the agent-forwarding solution might be very useful in some cases too - imagine a situation where you are syncing a site with 200gb of files and you have a laptop with a 100gb SSD disk.

sruggier:
Sorry, my mistake, I didn't mean to close this.

Crell:
Hm. I don't have a Mac so can't test this.  It needs to get rebased either way, though, as it doesn't merge cleanly at the moment.

badrange:
Hi Crell - there is nothing mac specific in these snippets of code, it could be useful for any developer on a system with ssh/rsync on it. 

Looking at my PR, I noticed that the commands seemed to be very mac-specific. I reorganised it now, emphasising the benefit of direct server-to-server transfers if the files folder is too big to fit on your own system. 

Hope this is better?

Crell:
Yep, that reads much better.  Thanks!

PR #170, opened 2015-08-19
[Do not merge] Solr configuration

GuGuss:


OriPekelman:
Don't really understand what I am supposed to put exactly in there
```
    configuration:
        core_config: !archive ""
```
Is this literally what the valued is ? if so this is ugly and should be changed. Or much much much better explained.

Also I would have liked a small "rational" section on top explaining why this is useful.

GuGuss:
#188.

PR #175, opened 2015-08-25
More detail for user administration

OriPekelman:
Needs review.

### DIFF
--- a/using/user-administration.md
+++ b/using/user-administration.md
@@ -49,11 +49,13 @@ environment to "Contributor".
 
 > **Important!**
 > An environment contributor can push code to the environment but that user 
-> role does not have SSH access to the environment.** only environment 
-> Administrators have SSH access**.
+> role does not have SSH access to the environment.** by default only 
+> environment Administrators have SSH access**. You can change the default
+> in `.platform.app.yaml` by example by specifying `ssh: contributor`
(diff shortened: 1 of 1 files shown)


using/user-administration.md
```diff
@@ -91,4 +116,62 @@ or delete the user's access to the environment entirely.
...
+    sprint1: viewer
+    hot-fix: viewer
+    pr-2: viewer
+    pr-3: viewer
+Adding users can result in additional charges.
+Are you sure you want to add this user? [Y/n]
+User alice@example.com created
+```
+
+Once this has been done the user will receive an email asking her to confirm
```
ralt: s/Once this has been done/Once this is done,/

?

ralt: "asking her"? Shouldn't we use the passive form to avoid gender issues?

pjcdawkins: Where a pronoun can't easily be avoided, IMO "they" / "them" is best: https://en.wikipedia.org/wiki/Singular_they

OriPekelman: Agreed, I generally prefer they/them as neutral. Almost never "he/him" (you won't find any in the docs afaik). In the example we have Alice. I believe here she is fine.
But feel free to amend

PR #185, opened 2015-09-11
Document sizing hints PHP parameters

GuGuss:


pjcdawkins:
Isn't this DevOps all over again, with a new / proprietary config format?

reference/platform-app-yaml.md
```diff
@@ -163,6 +163,23 @@ application in MB.
 > The minimal recommended disk size is 256MB. If you see the error **UserError: Error building the project: Disk size may not be smaller than 128MB**, increase the size to 256MB.
 
 
+## Runtime
+
+The `runtime` defines sizing hints (in MB) used to determine the maximum amount of requests to process concurrently (i.e. number of workers) for a PHP instance depending on its size. Those are not hard limits.
+
+Example:
```
robertDouglass: I'm missing how I know what the sum of the two numbers shoudld be? Why wouldn't a customer just write 512 and 1024? What happens if they do?

sruggier:
Sorry, my mistake, I didn't mean to close this.

reference/platform-app-yaml.md
damz: Yeah, we need to add a sentence here that says: "The more memory your application consumes, the less workers it can have at a particular service size."

Crell:
Is this still a WIP?  If those are real switches that work, they should be documented.  Although perhaps note that they should rarely be used?  (Drupal almost always uses more than 45 MB of RAM...)

reference/platform-app-yaml.md
```diff
@@ -163,6 +163,26 @@ application in MB.
 > The minimal recommended disk size is 256MB. If you see the error **UserError: Error building the project: Disk size may not be smaller than 128MB**, increase the size to 256MB.
 
 
+## Runtime
+
+The `runtime` defines sizing hints (in MB) used to determine the maximum amount of requests to process concurrently (i.e. number of workers) for a PHP instance depending on its size. Those are not hard limits.
```
pjcdawkins: Replace
> The `runtime` defines sizing hints (in MB) used to determine
with
> In the `runtime` section you can define sizing hints (in MB) to determine

Crell:
@GuGuss Can we include some recommendations on when/why one would want to set these?  As is, I have no idea when I'd want to touch them, or what it means that they're just "hints".  As someone potentially editing a config file I don't have enough information to touch them.

sruggier:
I've added an explanation of why one would want to change these: it has to do with how much memory your application tends to use in each worker process. I also included a small expanation of how to measure the memory usage of the worker processes in an environment.

mrkschan:
FYI https://gist.githubusercontent.com/damz/13947022592287d64a1e/raw/ab564f352e3f5a13471aecd19cdacd66cfbce360/HOG.md

Crell:
Woah, now that's more like it!  Thank you, @sruggier and @mrkschan!  Can one of you try to merge KS's changes into Simon's?  I like the more condensed approach in KS's proposed text, but Simon's goes deeper into how to measure your memory usage so you know what you can safely set it to.

mrkschan:
@Crell the gist I posted is written by @damz :)

Crell:
Rebased and reorganized. This PR now puts the sizing information in the PHP configuration page, since it's all PHP-specific.  The .platform.app.yaml page just points to language-specific configuration, which we can expand as we allow more language-specific configuration.

The main text is the snippet from @damz.  I also included the instructions from @sruggier on how to come up with a good guess for the memory size, but as discussed in Slack it's probably incomplete/less than useful.  We should fix up that bit before merging.  Simon, better suggestions welcome (including "don't, it's too complicated.")

Crell:
I removed the section on deriving a good size, as after talking with @sruggier I don't think it's going to be useful to people.  We're mostly documenting this section for completeness, not because people should be playing with it most of the time. :smile: 

I think this is ready to merge.  Assigning to @sruggier for validation.

sruggier:
I think whatever explanation we can give is much less confusing than for there to be no explanation. When I revisit this PR, I'll rewrite the explanation to mention both the `USED` column in top, and the memory usage in the PHP logs. Both of those are easy enough to interpret.

Crell:
Note: It looks like some earlier reorganization got lost in the rebase.  Simon is going to sort that out on Monday, then we'll re-review.

PR #189, opened 2015-09-17
Add a know issue for RabbitMQ.

GuGuss:


reference/services/rabbitmq.md
```diff
@@ -58,3 +58,17 @@ foreach ($relationships['rabbitmq'] as $endpoint) {
   $container->setParameter('rabbitmq_port', $endpoint['port']);
 }
 ```
+
+## Known issue
+
+The PHP container that is deployed doesn't ship with the [AMQP](https://pecl.php.net/package/amqp) PHP extension. This means you won't be able to use the RabbitMQ Management Command Line Tool as is.
```
damz: Isn't that confusing the "RabbitMQ Management Command Line Tool" with the "AMQP PHP extension", that sounds like they are two completely different things?

PR #192, opened 2015-09-29
Document using submodules

GuGuss:


using/submodules
```diff
@@ -0,0 +1,48 @@
+# Use Git submodules
```
ralt: According to other articles, it should be "Using Git submodules" (same as in the summary)

ralt: The file should be named "using/submodules.md" (add the extension)

using/submodules
```diff
@@ -0,0 +1,48 @@
...
+
+## Clone submodules during deployment
+
+Platform.sh allows you to use submodules in your Git repository. They are usually
+listed in a ``.gitmodules`` file at the root of your Git repository. When you push via Git,
```
ralt: There's no need for double ` in markdown, a single one as in:

    `.gitmodules`

is fine

ralt:
LGTM

PR #193, opened 2015-09-29
Add a known issue to PostgreSQL

GuGuss:


ralt:
"Known issue" feels like a wrong name when it's not a platform.sh issue, but rather an issue customers can get when using it.

sruggier:
Nice, I hit this when I was testing postgresql, but I didn't think of adding it to the documentation. I'm sure I won't be the last person to hit this.

PR #194, opened 2015-10-07
supplement redirect documentation

sharabash:
I examined the code platform/container/router/templates/nginx.conf.mako, there is nothing implemented that isn't already documented, so I supplemented the documentation with examples of 'prefix' and 'append_suffix' usage, and with example values for the 'expires' property (which I'm not sure about). I said "Examples of valid values include 3600s, 1d, 2w, 3m." -- I assume that's right, if somebody can confirm. I did some minor refactor in wording. The platform is up here: https://master-2vh7mcnp5eakq.eu.platform.sh/user_guide/reference/redirects.html

For some reason some of the `code` strings are highlighted red and some are grey -- I have no idea why.

### DIFF
--- a/reference/redirects.md
+++ b/reference/redirects.md
@@ -1,15 +1,18 @@
 # Redirects
 
-Managing redirection rules is a common requirement for web applications, especially in cases where you do not want to lose incoming links that have changed or been removed over time.
-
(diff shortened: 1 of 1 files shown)


GuGuss:
LGTM.

PR #198, opened 2015-10-16
Use correct database user for tunnel instructions

oxyc:
Or maybe add an example how to figure it out if it changes
```bash
php -r 'print_r(json_decode(base64_decode($_ENV['PLATFORM_RELATIONSHIPS'])));'
```


pjcdawkins:
Thanks!

PR #203, opened 2015-10-30
Document the Drush 7.x Workaround

joshmiller83:
It works, but there's two weird bugs. I will attach a screenshot in the PR.

image


pjcdawkins:
I think a simpler workaround is to define `drush-backups` as a mount:
https://github.com/pjcdawkins/platformsh-examples/blob/drush-backups-dir/.platform.app.yaml#L34

or this works without saving any files (just piping the dump via SSH, straight into the import):
```
drush @site.remote sql-dump | drush @site._local sqlc
```

pjcdawkins:
Actually since your PR helps, I'm going to merge it and edit afterwards to add the direct streaming method (I don't think we should be recommending Drush 6)

PR #204, opened 2015-11-03
Gitbook 2 5 1

OriPekelman:
This pull request updated gitbook to 2.5.1 and replaces the built-in search that is now optional with google search.

Google Search still needs to be themed/integrated.


OriPekelman:
Ok something is very wrong with my PR.. any page I visit re-displays the home..
And I broke GA.

OriPekelman:
And then I have this commit that insinuated itself in https://github.com/platformsh/platformsh-docs/commit/b044607ced05a8dc7d5ee9fdde4c32bc3ff2659e

PR #208, opened 2015-11-19
Php7

OriPekelman:


OriPekelman:
I am referring to @Kazanir work (we should probably run some newer ones)

Kazanir:
My stuff is all about requests per second and Ori's verbiage is fine in that sense. We could *also* comment that memory usage is much lower than PHP5; this is a separate metric than "performance" as we typically understand the term.

PR #210, opened 2015-11-20
Complete environment variables list, and sort it

pjcdawkins:


pjcdawkins:
I guess I could add PLATFORM_MODE, or do we want to keep Enterprise stuff out of here?

PR #213, opened 2015-12-03
PHP 7 has been released

GuGuss:


reference/toolstacks/php/php7.md
```diff
@@ -10,22 +10,6 @@ compared with PHP 5.6!
 To your joy you may also discover it has a much lower memory footprint than
 both PHP 5.6 and HHVM so in some cases it may outperform both by a margin.
 
-But this PHP 7.0 is not yet released. So this is beta level support. You may
-get into trouble... and we may not be able to help. Using this in production 
-is a risk, we can not recommend, until we announce official stable support.
-
->  ** BETA MEANS BETA **
```
ralt: Even though we're not in beta anymore, the list of available extensions is not going to be the same as on php-5.6 yet either, so this paragraph should stay.

PR #221, opened 2016-01-26
Update README.md

McGo:
removed duplice "which you can find here" without link.


sruggier:
Thank you very much for spotting this!

PR #222, opened 2016-01-26
Update redis-drupal-7.md

McGo:
As users might use the latest stable branch of the redis module (currently 3.x) I would recommend to point to a predis tar ball that is supported by the module. As the modules README suggests, it is 1.x, so I changed the documentation for that.


pjcdawkins:
Thanks!

PR #226, opened 2016-02-03
PF-1443: Fixing sidebar and anchor tags

dudenhofer:


sruggier:
From my perspective, the anchors work now, and the URL isn't mangled.

PR #227, opened 2016-02-03
Use English 'the'

mrkschan:


ralt:
Ah, thanks for spotting that!

PR #231, opened 2016-02-03
Drop full stop right before colon

mrkschan:


ralt:
Thanks, we really appreciate these contributions.

PR #241, opened 2016-03-10
Document the new application configuration

GuGuss:


reference/platform-app-yaml.md
```diff
@@ -2,47 +2,50 @@
...
+        redis: "redis:redis"
+    web:
+        locations:
+            "/":
+                root: "public"
+                expires: -1
+                passthru: "/index.php"
+                index: 
+                    - index.php
+                allow: true
```
pjcdawkins: What does this `allow: true` do? If this allows all static (non-`.php`) files to be served (?) I think we shouldn't have that in our example.

reference/platform-app-yaml.md
```diff
@@ -2,47 +2,50 @@
...
-    # We run deploy hook after your application has been deployed and started.
-    deploy: |
-        cd public
-        drush -y updatedb
-crons:
-    drupal:
-        spec: "*/20 * * * *"
-        cmd: "cd public ; drush core-cron"
-```
+    name: front
```
pjcdawkins: Can we have a brief comment on every single item in here, perhaps?

reference/platform-app-yaml.md
GuGuss: @pjcdawkins I think it's fine as this for an example. We probably don't want to have too many options in an example.

PR #249, opened 2016-04-18
Rephrase the rules of location

mrkschan:
Precisely state the list of regex are examples. This prevents confusion that the regexs on list are the only supported file extensions.


sruggier:
Thanks for fixing this!

PR #250, opened 2016-04-18
Prevent reusing "live" to describe the site with non-live environment

mrkschan:
It's confusing to describe the environment is non-live but the site is live. Let's prevent that.


sruggier:
I wasn't quite happy with how elsewhere in the paragraph, we call a non-production environment "non-live", so I ended up just reviewing the entire document.

PR #257, opened 2016-05-09
Update customizing-settings-php.md

joshbotnet:
Based on Patrick Dawkins' response to support ticket https://support.commerceguys.com/requests/8545

### DIFF
--- a/reference/toolstacks/php/drupal/customizing-settings-php.md
+++ b/reference/toolstacks/php/drupal/customizing-settings-php.md
@@ -69,3 +69,19 @@ if (!empty($_ENV['PLATFORM_RELATIONSHIPS'])) {
   $conf['cache_class_cache_form'] = 'DrupalDatabaseCache';
 }
 ```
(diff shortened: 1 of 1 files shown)


Crell:
Thanks, @joshbotnet. We're actually working on an update to the Drupal 8 example template that uses Composer and pulls in the needed patches automatically.  That should be deployed next week.  Once that's deployed I will revisit this issue and we can update it accordingly.

Crell:
Hi @joshbotnet. With the D8 composer branch now deployed, do you want to adjust this branch?  Mainly, the patches are already included, and we now use 8.1.2.  It's also not necessary to install elsewhere as a result, so we can remove that line.

The hash salt is now created from the settings.platformsh.php file, too, so that's not something the user needs to worry about (except on their local system).

pjcdawkins:
Link: https://github.com/platformsh/platformsh-example-drupal/tree/8.x

PR #261, opened 2016-05-12
Clarify Drupal settings in non-drupal build flavor

mrkschan:


Crell:
Mm, this is wrong.  Especially with the new D8 template that doesn't use the Drupal build flavor anymore.  Checking settings.local.php into the repository is against Drupal recommended practices.

mrkschan:
So, do we need a section for Drupal 8 and another one for Drupal 7? This is an issue we always see on Zendesk.

pjcdawkins:
The `settings.php` page is a bit of a mess - I suggest #288 instead

Crell:
Closing in favor of #288

PR #262, opened 2016-05-12
Update public-ip-addresses.md

rvanlaak:


inominate:
The outgoing IPs listed currently are correct, it's only three specific addresses, not the entire subnet.

PR #266, opened 2016-05-18
Add FAQ for data export

mrkschan:


reference/faq/known-issues.md
```diff
@@ -14,6 +14,32 @@ If you want to delete your project and cancel your subscription, simply go to yo
...
 
+## How can I export my data?
+
+If you have file mounts, you can keep a copy of the files using `rsync` like below.
+
+```
+rsync -r <project-id>-<environment>@ssh.<region>.platform.sh:/app/private/ .
+```
+
+Data stored in database can be dumped onto the filesystem and then `rsync` to local as well.
```
pjcdawkins: You can also dump it directly to local, streaming over SSH, with just:

```
platform sql-dump
```

That should work for MySQL or PostgreSQL

mrkschan: Would that transmission have compression turned on?

pjcdawkins: Yes, SSH compression is enabled. The resulting file is not compressed (unless you do something like  `platform sql-dump --stdout | gzip > dump.sql.gz`)

PR #267, opened 2016-05-19
Add section for WebSocket setup 

mrkschan:


Crell:
Can we also include a larger note, here or as its own page, on how one would setup a websocket route? Why does it need to be different? That's unclear from just this text.

mrkschan:
@Crell updated

reference/routes-yaml.md
```diff
@@ -178,6 +177,19 @@ You will see a route such as `https://*---add-theme-vmwklxcpbi6zq.eu.platform.sh
...
 
+##WebSocket routes
+
+To use WebSocket on a route, `cache` must be disabled. WebSocket is incompatible with buffering which is required for cache.
+Here is an example to define a route for serving WebSocket:
+
+```yaml
+"http://{default}/ws":
+    type: upstream
+    upstream: "websocket:http"
```
Crell: Where does this upstream come from?  Is websocket the name of a a project, or a forwarding type?  (This is a general problem of the documentation that it makes the foo:foo syntax very unclear, but in this particular case I have no idea what these magic values are or where they come from.)

ralt: @Crell "upstream" is part of our YAML schema. When you have "type: upstream", we look for the "upstream" key to know where the route should go.

As for the values, "websocket" is the name of the application, while "http" is the name of the endpoint.

mrkschan:
@Crell revised :)

PR #270, opened 2016-05-19
Update upstream property description

pawpy:
Update upstream property description in route configuration.


pawpy:
Closing since ":php" or ":http" doesn't signify either app runtime or communication protocol.

PR #279, opened 2016-05-27
fixed going live link

r0fls:
Using the full url will avoid breaking these links if they get moved around.


r0fls:
Oh I guess I never explicitly mentioned that the link doesn't work in the docs.

PR #282, opened 2016-06-01
Mention gzip static on router

mrkschan:


mrkschan:
gzip_static is on Nginx on app container, not on router, should use another section for docs.

PR #283, opened 2016-06-01
Mention gzip on the web server for the application

mrkschan:


reference/platform-app-yaml.md
```diff
@@ -152,6 +152,10 @@ variable. The right-hand side is in the form
 
 The `web` key defines how the application is exposed to the web (in HTTP). Here we tell the web application how to serve content, including static files, front-controller scripts, index files, index scripts, and so on. We support any directory structure, so the static files can be in a subdirectory, and the `index.php` file can be further down.
 
+> **Note**
+> Gzip compression is enabled to send precompressed static files with the ".gz" filename extension.
+> Though, dynamic content is not compressed due to [well known security issue](https://en.wikipedia.org/wiki/BREACH_%28security_exploit%29).
```
Crell: Reword: "However, dynamic content..."

Crell:
If I understand correctly, that means applications can still gzip stuff on their own and it won't get double-encoded?  That should be clarified.  (Drupal does nutty things with gzipping at times.)

mrkschan:
@crell do you really mean some Drupal modules would gzip compress using PHP?

mrkschan:
@crell updated

Crell:
Historically, Drupal has tried to figure out if it needs to gzip stuff itself and then does so, in case the web server doesn't.  Basically, Drupal has never trusted the server.  There was a period when the result was rather buggy.  I'm pretty sure it's all stable now, but I just wanted to confirm that if Drupal tries to gzip stuff it won't break everything.  (Honestly... off hand I'm not sure if it still tries to. I've not had to think about it in a while.)

mrkschan:
@Crell Since we only provide gzip compression for pre-compressed static files, I guess the Drupal case does not really matter here. What do you think?

Crell:
I think just adding the word "automatically" would be sufficient.  So it becomes "However, dynamic content is not automatically compressed...".  That helps clarify that an app that wants to send gzipped content can still do so, and nothing will get in its way or try to undo it.  (That's more what I was getting at.)

mrkschan:
updated

reference/platform-app-yaml.md
```diff
@@ -152,6 +152,10 @@ variable. The right-hand side is in the form
 
 The `web` key defines how the application is exposed to the web (in HTTP). Here we tell the web application how to serve content, including static files, front-controller scripts, index files, index scripts, and so on. We support any directory structure, so the static files can be in a subdirectory, and the `index.php` file can be further down.
 
+> **Note**
+> Gzip compression is enabled only for serving precompressed static files with the ".gz" filename extension.
+> However, dynamic content is not automatically compressed due to [well known security issue](https://en.wikipedia.org/wiki/BREACH_%28security_exploit%29).
```
Crell: Hm. This should be "due to a [well known security issue]".  Note the "a".

My bad for not catching that before, sorry. :frowning:

mrkschan: Good catch :) fixed.

Crell:
Thanks!

PR #285, opened 2016-06-03
fixes broken links by removing error reporting

r0fls:
Removing error reporting fixes a group of broken URLs and will redirect to the main page if the URL is invalid.

### DIFF
--- a/index.php
+++ b/index.php
@@ -8,6 +8,11 @@
 * Then Fallback to just the last part of the url
 * Then fallback to the root.
 **/
+// Don't show any errors since we're live. 
+// Comment out the below two lines when debugging.
+error_reporting(0);
+@ini_set('display_errors', 0);
+
 function matchPath($path, $current_urls){
     $location = "/"; // Fallback to root
     foreach ($current_urls as $url) {


ralt:
Thanks! But this should be done through the settings.php.

sruggier:
Sorry I'm just starting to pay attention now, but in my opinion, if the URL is invalid, we should return a 404 instead of a redirect, for the reasons I'll describe below.

As a user, it's very confusing sometimes when a redirect like this occurs. For example, if I open a link in a new tab and then get to it a couple minutes later, I'll probably be wondering why I have the front page of the docs site loaded, and it may take me a few seconds to realize that it was because of a redirect.

Also, some users may have an old link to a page in the docs that has been moved, and may want to look it up in the Wayback Machine or Google cache. There are bookmarklets that make this a one-click process, but if the redirect happens, then a bookmarklet can't be used, and the user has to manually copy and paste the URL.

Crell:
I agree with @sruggier.  We should only redirect in case we know what the new URL is and can send a 301 redirect.  Redirecting to the home page is just needlessly confusing for the user.

r0fls:
Well the current message for invalid URLs isn't very helpful and looks very unprofessional in my opinion:
https://docs.platform.sh/blah

sruggier:
Yeah, it wasn't my intent to shut down this whole pull request: the current behaviour still sucks, and we can fix it. We just need to return a 404 and clean error page instead the PHP errors that pop out now.

sruggier:
I'm not sure what changed, but the live docs site seems to redirect to the home page now. I guess we should revisit this in a separate PR.

r0fls:
Strange, I started this because https://docs.platform.sh/integrations/blackfire.html was giving the weird error. Now it works fine 👍

sruggier:
Yeah, I see some fixes in the commit history related to static file serving, that might explain the difference.

r0fls:
Looks like it was this https://github.com/platformsh/platformsh-docs/commit/349137513ca9c5c192f89579c9167de19eb7ce4e. I tested locally and that fixes it.

Note that it still doesn't give a 404:
https://docs.platform.sh/blah

But I think that's a next step and this is better than the poor error we were getting.

PR #287, opened 2016-06-06
Add bz2, igbinary, mongodb to PHP extensions list

mrkschan:


ralt:
Maybe it should be noted that the `mongo` extension is not available anymore in php7? (i.e. `mongodb` replaces `mongo` in php7)

sruggier:
Yeah, I think it would make sense to add something like "5.6 and older" in brackets on the `mongo` extension.

mrkschan:
updated.

PR #288, opened 2016-06-06
Update customizing-settings-php.md

pjcdawkins:
* Emphasises that this only applies to apps with the `drupal` build flavor.
* Emphasises that you should not have settings.local.php in your repository.
* Removes reference to "You should include a settings.local.php" because that was confusingly written (it means "include via PHP" not "include in your repository").
* Removes part about `settings.dev.php` because I think it was more confusing than helpful.


reference/toolstacks/php/drupal/customizing-settings-php.md
```diff
@@ -1,12 +1,13 @@
 # Customizing settings.php
 
-Platform.sh allows you to use your own `settings.php` that you commit to
-your Git repository.
+For applications using the `drupal` build flavor (those based on our [Drupal 7
+example](https://github.com/platformsh/platformsh-example-drupal/tree/7.x)),
+Platform.sh automatically generates a `settings.php` and `settings.local.php`
```
Crell: It generates both? Or only the former if it's missing? That's different than the text being replaced.

Crell:
Looks fine otherwise.

pjcdawkins:
Good point - updated

Issue #220, opened 2016-01-06
document_root implications


hackzilla:
I was recently talking to support about why an application gets deployed into /app or sometimes /app/public

```
This is actually controlled by the document_root setting in the .platform.app.yaml. If the document_root is set to /, it will move the application to /app/public and use that as the root to allow for the private mounts to be kept private. In the case of your stats-tracking project it's doing this because there is no .platform.app.yaml so it is using the defaults.

Best,
Brad
```

This information should be added to: https://docs.platform.sh/user_guide/reference/platform-app-yaml.html

Crell:
I agree.  PR opened (see above). Please comment there if the additional text is sufficient.  (I based it on the support comment above.)  Thanks!

PR #289, opened 2016-06-06
Rewrite path documentation for platform.app.yaml

Crell:
As noted in https://github.com/platformsh/platformsh-docs/issues/220


pjcdawkins:
* It should be `public` or `/public` rather than `/app/public`. The document root is relative to the application root, and it's best to avoid hardcoding that the application root is `/app`, if we can avoid it.
* We have changed this kind of web configuration from a single `web.document_root` to many `web.locations`, each of which is keyed by a URL path (e.g. `/`) and has an optional `root` property... so you would be defining `web.locations./.root`. But that could be addressed in a follow-up.

mrkschan:
I think we should upgrade all examples on the docs in flavor of "web.locations"... ref https://docs.platform.sh/user_guide/reference/upgrade/index.html#changes-in-version-2016-3

Crell:
A quick Find nets over a dozen cases of document_root.  We should definitely convert to the new style for all examples, but let's do that in another issue.

Switched to new style here for now, and changed /app/public to just "public".

This is another area where we really ought to be pulling the default straight out of our internal configuration to make sure it's always up to date.  Anyone know how easily we could do that?

reference/platform-app-yaml.md
```diff
@@ -396,4 +398,10 @@ crons:
         cmd: "cd public ; drush core-cron"
 ```
 
+## Top level document_roots
+
+Platform.sh requires that the document root not be at the root of the project.  It is important for security that
+private file mounts not be web-accessible. If the document_root is set to `/`, it will move the application to
```
mrkschan: i guess this is `document root` instead of `document_root`...

reference/platform-app-yaml.md
```diff
@@ -396,4 +398,10 @@ crons:
...
+`public` and use that as the root to allow for the private mounts to be kept private.
```
mrkschan: @ralt can you confirm this after your latest changes?

Crell:
@ralt See @mrkschan's comment to you; the text changed but the question is still relevant for confirmation.

reference/platform-app-yaml.md
ralt: No, that's wrong. With `web.locations./.root: ""`, the document root will be the root of the application, hence having mounts below it. This is obviously not recommended, but it's possible.

Crell:
@ralt Can you suggest better text for that paragraph?  Just go ahead and push a new commit to this branch if that's easier.

mrkschan:
@ralt don't we have set restriction on that? That is probably an insecure
setup in the container.

damz:
This is in a state of flux. At this point `root: /` (or `root: ""` which is actually normalized to the same value) is accepted as a valid value, but will or will not work depending on the build flavor. Until we figure out whether we want to actually support this or not moving forward, I would just recommend to shy away from `root: /`.

Crell:
OK, based on discussion in Slack, I went through and rewrote this diff and fixed up a lot of other commentary around paths elsewhere in the file.  Hopefully this should be clearer/more accurate.  There were some conflicts so I had to rebase.

reference/platform-app-yaml.md
```diff
@@ -187,10 +187,17 @@ The `locations` key allows you to provide specific parameters for different URL
 
 It has a few subkeys, which are:
 
+<<<<<<< f548b3b9943b7c351adf5e3d45852cf5fa1489de
```
pjcdawkins: Git conflict marker

reference/platform-app-yaml.md
```diff
@@ -187,10 +187,17 @@ The `locations` key allows you to provide specific parameters for different URL
...
 * `passthru`: Whether to forward disallowed and missing resources from this location to the application. Can be `true` or `false` (for an application capable of handling HTTP traffic directly) or a URI path string (for a CGI based application, e.g. PHP). Typically your application's front controller, such as `/index.php` or `/app.php`.
 * `index`: The file or files to consider when serving a request for a directory. Can be a file name, an array of file names or *null*. Typically `index.html`. Note that in order for this to work, the static file(s) should be allowed by your rules. For example, to use a file named `index.html` as an index file, your rules must allow elements that match the filename, like `- \.html$`.
 * `expires`: How long to allow static assets from this location to be cached (this affects the Cache-Control and Expires headers). Can be a time or *-1* for no caching. Times can be suffixed with "ms" (milliseconds), "s" (seconds), "m" (minutes), "h" (hours), "d" (days), "w" (weeks), "M" (months, 30d) or "y" (years, 365d). The `expires` directive and resulting headers are left out entirely if this isn't set.
+=======
```
pjcdawkins: Git conflict marker

reference/platform-app-yaml.md
```diff
@@ -187,10 +187,17 @@ The `locations` key allows you to provide specific parameters for different URL
...
+* `index`: The file or files to consider when serving a request for a directory. Can be file name, and array of file names or *null*. Typically `index.html`. Note that in order for this to work, the static file(s) should be allowed by your rules. For example, to use a file named `index.html` as an index file, your rules must allow elements that matches the filename, like `- \.html$`.
+* `expires`: How long to allow static assets from this location to be cached (this enables the cache-control and expires headers). Can be a time or *-1* for no caching. Times can be suffixed with "ms" (milliseconds), "s" (seconds), "m" (minutes), "h" (hours), "d" (days), "w" (weeks), "M" (months, 30d) or "y" (years, 365d). The `expires` directive and resulting headers are left out entirely if this isn't set.
+>>>>>>> Rework documentation and explanation around path handling.
```
pjcdawkins: Git conflict marker

reference/platform-app-yaml.md
```diff
@@ -390,8 +398,10 @@ relationships:
     solr: "solr:solr"
     redis: "redis:redis"
 web:
-    document_root: "/"
-    passthru: "/index.php"
+    locations:
+        "/":
+            root: "public"
+            passthru: "/index.php"
```
pjcdawkins: There are two relative paths here, one starts with `/`, the other doesn't. Would it work to make them consistent?

damz: One of them is a URL, these start with `/`. The other is a path on disk relative to the application root, these do not.

pjcdawkins: OK. In that case it makes sense (I had assumed the passthru was also a path on disk)

Crell: The confusion around that is exactly what I'm trying to sort out and properly document at the moment. :smile:

Crell:
Rebased again to fix the bad rebase before, and tweaked the passthru language a bit further. In my own experimentation, passthru simply doesn't work unless it's on the / entry. Anything else results in nothing but 404s or 500s.  For the time being at least I think it's best to treat it as only supported on the / entry.

reference/platform-app-yaml.md
```diff
@@ -174,7 +174,7 @@ It has a few subkeys which are:
 
 #### Locations
 
-The `locations` key allows you to provide specific parameters for different URL prefixes.
+The `locations` key allows you to provide specific parameters for different URL prefixes. Each entry's key is a domain-root-relative path, and its value is configuration directives for that path.  That is, if your domain is `example.com` then `"/"` means "requests for `example.com/`", while `"/admin"` means "requests for `example.com/admin`".
```
sruggier: I think this would be better worded as:
> The `locations` key allows you to customize how your application handles the path component of an incoming request URL. It contains a key-value mapping, where each key is a path prefix that will be compared with the request URL, and the value is another mapping of configuration keys that lets you customize how to handle request that match that path prefix. If more than one location matches the same request URL, the longest matching location is used.
> 
> For example, if your domain is `example.com`, and you have a web configuration with locations set as follows:
> ```
> web:
>     locations:
>         "/":
>             [...]
>         "/a":
>             [...]
>         "/a/b":
>             [...]
> ```
>
> Then request URLs will match as follows:
> - an incoming request with a URL path of `/a` will match the `/a` location
> - `/b` will match the `/` location
> - `/a/b/c` will match the `/a/b` location
> - `/a/c/b` will match the `/a` location
>
> This allows us to set up general URL handling within the `/` location, and
> then use more specific locations to override that behaviour and handle
> certain URL paths differently.

reference/platform-app-yaml.md
```diff
@@ -174,7 +174,7 @@ It has a few subkeys which are:
 
 #### Locations
 
-The `locations` key allows you to provide specific parameters for different URL prefixes.
+The `locations` key allows you to provide specific parameters for different URL prefixes. Each entry's key is an absolute URL path, and its value is configuration directives for that path.  That is, if your domain is `example.com` then `"/"` means "requests for `example.com/`", while `"/admin"` means "requests for `example.com/admin`".
```
mrkschan: According to https://tools.ietf.org/html/rfc3986#section-4.2

> A relative reference that begins with a single slash character is
   termed an absolute-path reference.  A relative reference that does
   not begin with a slash character is termed a relative-path reference.

Shall we replace "an absolute URL path" with either "an absolute-path reference" or "a relative URL using absolute-path reference"?

reference/platform-app-yaml.md
```diff
@@ -187,10 +187,10 @@ The `locations` key allows you to provide specific parameters for different URL
...
+* `index`: The file or files to consider when serving a request for a directory. Can be file name, and array of file names or *null*. Typically `index.html`. Note that in order for this to work, the static file(s) should be allowed by your rules. For example, to use a file named `index.html` as an index file, your rules must allow elements that matches the filename, like `- \.html$`.
+* `expires`: How long to allow static assets from this location to be cached (this enables the cache-control and expires headers). Can be a time or *-1* for no caching. Times can be suffixed with "ms" (milliseconds), "s" (seconds), "m" (minutes), "h" (hours), "d" (days), "w" (weeks), "M" (months, 30d) or "y" (years, 365d). The `expires` directive and resulting headers are left out entirely if this isn't set.
```
mrkschan: Why don't wrap Cache-Control and Expires with "`" since they are "code" of the HTTP protocol?

reference/platform-app-yaml.md
```diff
@@ -187,10 +187,10 @@ The `locations` key allows you to provide specific parameters for different URL
...
+* `root`: The folder to serve static assets for this location from relative to the application root. The application root is the directory in which the `.platform.app.yaml` file is located.  Typical values for this property include `public` or `web`.  Setting it to `"/"` is not recommended, and its behavior may vary depending on the type of application.
+* `passthru`: Whether to forward disallowed and missing resources from this location to the application. Can be true, false or a URI path string relative to the web root. For non-PHP applications it will generally be just true or false.  In a PHP application, this will typically be the front controller such as `/index.php` or `/app.php`.  If your application relies on a `.htaccess` file with `mod_rewrite` under Apache, this entry is the equivalent. Note: The value of `passthru` must be an absolute URL path, which means if it does not begin with the same value as the location key it is under the passthru may evaluate to another entry. That may be useful when you want different cache settings for different paths, for instance, but want missing files in all of them to map back to the same front controller.
```
mrkschan: `passthru` is not "equivalent" to mod_rewrite since we do not support Nginx rules like `rewrite /(.*).html$ /index.php?action=$1;` yet (which is `RewriteRule /(.*).html$ /index.php\?action=$1` in mod_rewrite). As you may already know, we simply do `rewrite . /index.php;` now.

mrkschan: The Note here is confusing me, can we simply say the following?

> Note: The value of `passthru` must be an absolute URL path, which can be used to forward the HTTP request to another location entry.

**note** Please see my comment on using the term "absolute URL path" above.

reference/platform-app-yaml.md
```diff
@@ -187,10 +187,10 @@ The `locations` key allows you to provide specific parameters for different URL
...
+* `index`: The file or files to consider when serving a request for a directory. Can be file name, and array of file names or *null*. Typically `index.html`. Note that in order for this to work, the static file(s) should be allowed by your rules. For example, to use a file named `index.html` as an index file, your rules must allow elements that matches the filename, like `- \.html$`.
```
mrkschan: Simply setting `allow: true` in this location entry should works as well.

reference/platform-app-yaml.md
```diff
@@ -228,7 +228,7 @@ application in MB.
 ### Mounts
 
 The `mounts` is an object whose keys are paths relative to the root of
-the application. It's in the form `volume_id[/subpath]`.
+the application (That is, where the `.platform.app.yaml` file lives). It's in the form `volume_id[/subpath]`.
```
mrkschan: For consistency, shall we use the following?

> (That is, where the `.platform.app.yaml` file is located).

See description of `root` above.

mrkschan: Aside, shall we also mention that `shared:files` used below is the only `volume_id` that we support right now?

reference/platform-app-yaml.md
```diff
@@ -289,13 +289,12 @@ Possible hooks are:
...
-    (MySQL, Solr, Redis...).
+    (MySQL, Solr, Redis...). However, the disk where the application lives is read-only at this point.
 
 > **Note**
-> The "home" directory is /app while your application will be
-mounted in /app/public (by default: you can define this yourself in your
-.platform.app.yaml file) - so you might want to `cd public` before
-running hooks.
+> Your application will always be mounted under `/app`, and whatever subdirectory
+you specified in the `web/locations/root` key above.  For example, `/app/web`. You may need
```
mrkschan: I think the original writing means `$HOME` is always set to `/app` and so we have to `cd web/` in order to run `php index.php` or whatever scripts under `/app/web`.

We can actually run any files in deploy hook without `cd` (e.g. `php /app/whatever.php`).

reference/platform-app-yaml.md
```diff
@@ -188,7 +188,7 @@ The `locations` key allows you to provide specific parameters for different URL
...
+* `passthru`: Whether to forward disallowed and missing resources from this location to the application. Can be true, false or a URI path string relative to the web root. For non-PHP applications it will generally be just true or false.  In a PHP application, this will typically be the front controller such as `/index.php` or `/app.php`.  If your application relies on a `.htaccess` file with `mod_rewrite` under Apache, this entry is the equivalent. Note: The value of `passthru` must be an absolute URL path, which means if it does not begin with the same value as the location key it is under the passthru may evaluate to another entry. That may be useful when you want different cache settings for different paths, for instance, but want missing files in all of them to map back to the same front controller.
```
pjcdawkins: Change
> a URI path string relative to the web root

to
> an absolute URL path

pjcdawkins: As it's an absolute URL path to the passthru, it can only evaluate to one thing - so we can remove the note about "the passthru may evaluate to another entry" (I find it confusing).

reference/platform-app-yaml.md
Crell: Should work for what?

reference/platform-app-yaml.md
Crell: Yes, probably, because that's quite unclear at present.

reference/platform-app-yaml.md
Crell: The reason that second entry is there (about the passthru evaluating to another entry) is because I just spent 2 days struggling with trying to figure out how things mapped and getting that evaluation happening from another entry, because it's not at all intuitive. I want to spare others those 2 days of "Wait, what??"

pjcdawkins: Maybe something like this then
> Note: if setting the `passthru` as a URL path, make sure it is absolute, i.e. begins with a `/`

because if it looks like an absolute path it should be fairly clear what it means, right?

reference/platform-app-yaml.md
mrkschan: `allow: true` means serving all static files that match the URL. We don't really need the "rules" in that case.

pjcdawkins: I wouldn't recommend `allow: true` for that very reason, but you're right, the text could perhaps be:

> Note that in order for this to work, the static files() should be allowed (by the `allow` or `rules` keys for this location).

reference/platform-app-yaml.md
Crell: hm. I think it's important to keep the "URL" in there, to distinguish between URL-based paths and filesystem-based paths.

reference/platform-app-yaml.md
mrkschan: fyi, `allow` is true by default :)

reference/platform-app-yaml.md
mrkschan: Perhaps ... "relative URL with a leading slash"? I'm not sure what is absolute URL path ...

reference/platform-app-yaml.md
pjcdawkins: I know, and that sucks :(

reference/platform-app-yaml.md
mrkschan: Oh ... what about "relative URL with absolute path"?

pjcdawkins: > absolute URL path (with a leading `/`)

Adding the word "relative" would make it even more confusing IMO

reference/platform-app-yaml.md
Crell: What do you recommend instead of "equivalent"?  I want to try and give context here for the many PHP apps that use a mod_rewrite-based front controller, for which this is the way they'd do the same thing. It's not as robust as mod_rewrite, certainly, but it's the equivalent technique.  How can this be phrased better?

reference/platform-app-yaml.md
mrkschan: Google doesn't tell me what is "absolute URL path" ...

pjcdawkins: "absolute URI path" may get better results

The "with a leading `/`" makes it pretty obvious IMO

reference/platform-app-yaml.md
mrkschan: We really have support ticket asking for "rewrite" support that we do not offer yet and so I think we better not to encourage our customers to look for the "equivalent" now.

I guess we can rephrase that as simple as the following?

> In a PHP application, this will typically be the front controller such as `/index.php` or `/app.php`.  This entry works like `mod_rewrite` under Apache.

reference/platform-app-yaml.md
Crell: I disagree that it's as obvious as you say. As I said, it took me a day and a half to wrap my brain around it. :smile:

reference/platform-app-yaml.md
sruggier: I agree with @mrkschan's feedback. Given that mod_rewrite is a module and not a specific configuration directive, it doesn't make sense to make a direct comparison with `passthru`. When we mention it in general terms like this, it implies that `passthru` somehow supports all of the use cases that `mod_rewrite` supports, which can only lead to confusion for readers that interpret it this way. I think it's enough to mention front controllers, and this part should be omitted:
>  This entry works like `mod_rewrite` under Apache

reference/platform-app-yaml.md
```diff
@@ -187,10 +187,10 @@ The `locations` key allows you to provide specific parameters for different URL
...
-* `index`: The file or files to consider when serving a request for a directory. Can be a file name, an array of file names or *null*. Typically `index.html`. Note that in order for this to work, the static file(s) should be allowed by your rules. For example, to use a file named `index.html` as an index file, your rules must allow elements that match the filename, like `- \.html$`.
-* `expires`: How long to allow static assets from this location to be cached (this affects the Cache-Control and Expires headers). Can be a time or *-1* for no caching. Times can be suffixed with "ms" (milliseconds), "s" (seconds), "m" (minutes), "h" (hours), "d" (days), "w" (weeks), "M" (months, 30d) or "y" (years, 365d). The `expires` directive and resulting headers are left out entirely if this isn't set.
+* `root`: The folder to serve static assets for this location from relative to the application root. The application root is the directory in which the `.platform.app.yaml` file is located.  Typical values for this property include `public` or `web`.  Setting it to `"/"` is not recommended, and its behavior may vary depending on the type of application.
```
sruggier: Instead of
>Setting it to `"/"` is not recommended, and its behavior may vary depending on the type of application.

We should broaden our advice to this, which I think is simpler, while still correct:
> Absolute paths are not supported.

pjcdawkins: I'd agree with that (the leading slash is ignored anyway, I believe, so absolute paths are impossible). Then the documentation doesn't deal with `root: ""`, but that would look like a hack anyway I'd hope. I also hope we ignore `../` ...

reference/platform-app-yaml.md
Crell: Updated that section.

reference/platform-app-yaml.md
Crell: Well, it's the application that is mounted there. ie, the directory tree starting from .platform.app.yaml.  That it's also $HOME when sshing in is not relevant here, I'd think.

reference/platform-app-yaml.md
```diff
@@ -228,14 +231,15 @@ application in MB.
 ### Mounts
 
 The `mounts` is an object whose keys are paths relative to the root of
-the application. It's in the form `volume_id[/subpath]`.
+the application (That is, where the `.platform.app.yaml` file lives). It's in the form `volume_id[/subpath]`.
+At this time, the only legal `volume_id` is `shared`.
```
mrkschan: The only supported `volume_id` is `shared:files`.

reference/platform-app-yaml.md
```diff
@@ -187,10 +187,10 @@ The `locations` key allows you to provide specific parameters for different URL
...
-* `index`: The file or files to consider when serving a request for a directory. Can be a file name, an array of file names or *null*. Typically `index.html`. Note that in order for this to work, the static file(s) should be allowed by your rules. For example, to use a file named `index.html` as an index file, your rules must allow elements that match the filename, like `- \.html$`.
-* `expires`: How long to allow static assets from this location to be cached (this affects the Cache-Control and Expires headers). Can be a time or *-1* for no caching. Times can be suffixed with "ms" (milliseconds), "s" (seconds), "m" (minutes), "h" (hours), "d" (days), "w" (weeks), "M" (months, 30d) or "y" (years, 365d). The `expires` directive and resulting headers are left out entirely if this isn't set.
+* `root`: The folder from which to serve static assets for this location, relative to the application root. The application root is the directory in which the `.platform.app.yaml` file is located.  Typical values for this property include `public` or `web`.  Setting it to `""` is not recommended, and its behavior may vary depending on the type of application.  Absolute paths are not supported.
```
pjcdawkins: Is this required? What's the default? Maybe we need that for all keys...

PR #291, opened 2016-06-07
Simplify logic to find basename of request URI

mrkschan:
This should suppress the notice which break the search function when visiting https://docs.platform.sh/user_guide/search_index.json.

```
Notice:  Only variables should be passed by reference in /app/_books/index.php on line 21
```

### DIFF
--- a/index.php
+++ b/index.php
@@ -18,7 +18,7 @@ function matchPath($path, $current_urls){
     }
     if ($location=="/"){ 
         // nothing was matched let's try a shorter match
-        $path = str_replace(["/index.html",".html"],"",array_pop(explode('/', $path)));
+        $path = basename($path);
         foreach ($current_urls as $url) {
             if (stripos($url, $path) !== FALSE ) {
                 $location = $url;


pjcdawkins:
That could all be replaced with:

```php
$path = basename($path);
```

mrkschan:
Not sure if this is the expected behavior...

```
curl -I https://pr-291-t2llqeifuhpzg.eu.platform.sh/user_guide/search_index.json
HTTP/1.1 301 Moved Permanently
Date: Tue, 07 Jun 2016 08:12:29 GMT
Content-Type: text/html; charset=UTF-8
Connection: keep-alive
Location: /
X-Platform-Processor: t2llqeifuhpzg-pr-291--php
X-Platform-Router: t2llqeifuhpzg-pr-291--router
X-Platform-Cache: MISS
```

mrkschan:
I guess it should not go through index.php according to the rule...

```
            rules:
                \.(css|js|gif|jpe?g|png|ttf|eot|woff|otf|search_index\.json|robots\.txt|html?
                    allow: true

```

mrkschan:
The real fix for search issue is https://github.com/platformsh/platformsh-docs/pull/292 instead :)

mrkschan:
@pjcdawkins updated.

pjcdawkins:
LGTM. There is a chance that `basename()` will return an empty string, but that should be OK, the `stripos()` will then always return `FALSE`.

PR #294, opened 2016-06-08
Add more details for cache control

mrkschan:
Reference - https://www.nginx.com/blog/nginx-caching-guide/

### DIFF
--- a/reference/cache.md
+++ b/reference/cache.md
@@ -14,10 +14,9 @@ won't access Drupal.
 
 ## Reference
 
-Cache is enabled by default in your `.platform/routes.yaml` file. This
-is an example:
+Cache can be enabled in your `.platform/routes.yaml` file like below:
 
-```bash
+```yaml
 http://{default}/:
     type: upstream
     upstream: php:php
@@ -69,18 +68,12 @@ And the following routes are **not** cached:
 > **note**
 > Regular expressions in routes are **not** supported.
 
-## Cache duration
-
-The cache duration is decided based on the `Cache-Control` response
-header value. If no `Cache-Control` header is in the response, then the
-`default_ttl` key is used.
-
 ## Cache key
 
 To decide how to cache a response, Platform.sh will build a cache key
 depending on several factors and store the response associated with this
-key. When a request comes with the same cache key, the response will be
-reused. It is similar to the `Vary` header in purpose.
+key. When a request comes with the same cache key, the cached response will
+be reused.
 
 Some parameters let you change this cache key: the `headers` key and the
 `cookies` key.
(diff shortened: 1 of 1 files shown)


mrkschan:
Does anyone has idea on when Nginx revalidate `ETag` in a response?

mrkschan:
Found this :) https://trac.nginx.org/nginx/timeline?from=2014-07-01T11%3A44%3A24Z&precision=second

reference/cache.md
```diff
@@ -69,6 +68,16 @@ And the following routes are **not** cached:
...
 
+## Cache behavior
+
+Cache is only applied to `GET` and `HEAD` requests.
+Responses with `Cache-Control` header set to `Private`, `No-Cache`,
+or `No-Store` are not cached. Also, this applies to responses with
+`Set-Cookie` header being set.
+
+Cache would also respect the `Vary` header in the response.
+Multiple versions of the cache will be created if `Vary` header is set.
```
ralt: Between the `Vary` header and the cache key defined in the `routes.yaml`, how does it play together?

mrkschan: `Vary` is similar to cache key. See https://www.nginx.com/blog/nginx-plus-r5-released/.

ralt: Maybe this should be properly documented? I won't be the only one asking that.

mrkschan: From http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache_valid

> If the header includes the “Vary” field with the special value “*”, such a response will not be cached (1.7.7). If the header includes the “Vary” field with another value, such a response will be cached taking into account the corresponding request header fields (1.7.7). 

First we need a understanding :)

mrkschan: Relevant code block? https://github.com/nginx/nginx/blob/bf838dc9dfca4e80b883c2e659c65f90fc1526a3/src/http/ngx_http_file_cache.c#L525

mrkschan: The cache data structure has `vary` and `varient` :) https://github.com/nginx/nginx/blob/c3aed0a23392a509f64b740064f5f6633e8c89d8/src/http/ngx_http_cache.h#L64

mrkschan: Example with cache key and vary header playing together - http://marc.info/?l=nginx&m=144432816821289&w=2

reference/cache.md
```diff
@@ -90,10 +83,41 @@ The default value for these keys are the following:
 ```yaml
 cache:
   enabled: true
-  headers: ["Accept-Language", "Accept"]
+  headers: []
   cookies: ["*"]
 ```
 
+`Vary` header in the response is also respected. Multiple copies under the
```
Crell: Missing "The" at the start of the sentence.

reference/cache.md
```diff
@@ -90,10 +83,41 @@ The default value for these keys are the following:
...
+same cache key would be created according to the value of this header.
+
+## Cache behavior
+
+Cache is only applied to `GET` and `HEAD` requests.
+Responses with `Cache-Control` header set to `Private`, `No-Cache`,
+or `No-Store` are not cached. Also, this applies to responses with
+`Set-Cookie` header being set.
```
Crell: Rephrase: "Responses with the `Set-Cookie` header set are also not cached."

reference/cache.md
ralt: Either way, we need to make that clear in the documentation.

mrkschan:
@ralt @Crell updated.

PR #296, opened 2016-06-09
Convert PHP extension list to actual lists

Crell:
For readability.  I also added some further clarifying notes around which extensions are available on which versions.  I tried to make it clearer, but the list is not consistent enough to really break down any nicer.

### DIFF
--- a/reference/toolstacks/php/configure-php.md
+++ b/reference/toolstacks/php/configure-php.md
@@ -34,25 +34,80 @@ runtime:
         - sqlite3
 ```
 
-The following extensions are enabled by default: *pdo, mysql, mysqli, mysqlnd,
-pdo_mysql, pdo_sqlite, sqlite3, gd, curl, intl, mcrypt and
-zendopcache/opcache*. You can disable those by adding them to the
-`disabled_extensions` list.
+The following extensions are enabled by default:
 
-This is the complete list of extensions that can be enabled: *apc (5.4 only),
-apcu, blackfire, bz2 (added in 7.0), curl, enchant, gd, gearman, geoip, gmp,
-http (5.4 and 5.5 only), igbinary (added in 7.0), imagick, imap, intl,
-json (added in 5.6), ldap, mcrypt, memcache, memcached, mongo (5.6 and older),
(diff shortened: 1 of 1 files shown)


reference/toolstacks/php/configure-php.md
```diff
@@ -34,25 +34,80 @@ runtime:
...
-This is the complete list of extensions that can be enabled: *apc (5.4 only),
-apcu, blackfire, bz2 (added in 7.0), curl, enchant, gd, gearman, geoip, gmp,
-http (5.4 and 5.5 only), igbinary (added in 7.0), imagick, imap, intl,
-json (added in 5.6), ldap, mcrypt, memcache, memcached, mongo (5.6 and older),
-mongodb (added in 7.0), msgpack (added in 5.6), mysql, mysqli, mysqlnd,
-opcache (5.5 and up), pdo, pdo_mysql, pdo_pgsql, pdo_sqlite,
-pecl-http (added in 5.6), pgsql, pinba, propro (added in 5.6), pspell,
-raphf (added in 5.6), readline, recode, redis, snmp, sqlite3, ssh2, tidy,
-xcache, xdebug, xhprof, xmlrpc, xsl, and zendopcache (5.4 only)*.
+The following extensions are enabled by default:
```
pjcdawkins: You have copied and reformatted the list perfectly, but I think it's incomplete

Crell: Is there some way to automate this list?  I considered scripting it, but my Python-fu wasn't quite strong enough yet. :smile:  (And doing it in PHP seemed like a bad one-off.)  It would likely be easier if we're OK with maintaining a separate list for each container version, since it's the weird overlaps that make it difficult to do.

Crell:
I added json, and converted the master list to a table derived off of running containers.  Unfortunately extracting what extensions are enabled by default but are not part of base PHP itself is surprisingly challenging, so we'll go with this list for now as it covers the biggies.

pjcdawkins:
The table loses a cell at the end 😞 

screen shot 2016-06-22 at 19 57 57

Crell:
Yeesh. That looks like an HTMLBook bug. The source is clean.  Since it would be an empty cell anyway, do we just go with it, or try to find some work around?  (I'm tempted to just go with it.)

Issue #214, opened 2015-12-03
Redis setup missing some details


sirtet:
A few issues in one:

1.
If i get it right, the redis server must also be enabled in .platform/services.yml

2.
on https://docs.platform.sh/user_guide/reference/toolstacks/php/drupal/redis-drupal-7.html
in the settings.php section it's not totally clear how this works:
Currently there's one line before the settings.local include setting which connector to use,
And then some more lines after the include. Amongst it one about the connector too, but it does not say to adjust there to the used connector, which i think i should.

3.
Can't i just as well put that code in settings.local.php to use redis eg. in production only? 
The article only promotes using environment variables for that.

4.
Also, the drupal-6 version tells how to check if redis is running, the D7 version is missing that.

sirtet:
5.
I could not get redis running following the D6/D7 instructions.
I also needed to setup the redis relationship in .platform.app.yaml

```
relationships:
   redis: "myredis:redis"
```
Seems like another missing bit in the D6/D7 doc.

PS:
The server is answering now on the shell, but i still did not get it to work:
> Fatal error: Call to undefined function db_merge() in /app/public/includes/bootstrap.inc on line 991

sirtet:
Ok, works now, i had a 2nd relationships- statement, which overwrote the first...

Crell:
@sirtet So is this issue resolved then, or do you think there are specific improvements to the docs still needed?

sirtet:
Well,my redis problems ars sloved, but i think the docs are still the same?
Sure, i would eg. add the testing- bit from D6 to the D7 page, and i think the bit about my 5. Point is also not in there...

Crell:
Let's continue this over in the PR: https://github.com/platformsh/platformsh-docs/pull/298.  Thanks @sirtet.

PR #298, opened 2016-06-10
Clean up Drupal 7 Redis documentation

Crell:
Per https://github.com/platformsh/platformsh-docs/issues/214, the Redis docs could use work.

I tested this by running through an install myself, and correcting everything I found that didn't work as advertised.  I also decided to officially recommend PhpRedis, as the C extension should be faster according to Benchmarks Found on the Internet(tm).

### DIFF
--- a/reference/toolstacks/php/drupal/redis-drupal-7.md
+++ b/reference/toolstacks/php/drupal/redis-drupal-7.md
@@ -2,9 +2,57 @@
 
 There are two options for using Redis with Drupal on Platform.sh, you
 can either use the [PhpRedis](https://github.com/nicolasff/phpredis)
-extension or the [Predis](http://github.com/nrk/predis) library.
+extension or the [Predis](http://github.com/nrk/predis) library.  PhpRedis
+requires a PHP extension (which we provide) and should therefore be faster in
+most situations. Predis is written entirely in PHP and so would require no PHP
+extension to install locally, but at the cost of some performance.
+
+If you are unsure which to use, we recommend using PhpRedis.
 
 ## Requirements
+
+### Add a redis service
+
+First you need to create a redis service.  In your `.platform/services.yaml` file,
+add or uncomment the following:
+
+```yaml
+rediscache:
+    type: redis:2.8
+```
+
+That will create a service named `rediscache`, of type `redis`, specifically version `2.8`.
+
+### Expose the Redis service to your application
+
+In your `.platform.app.yaml` file, we now need to open a connection to the new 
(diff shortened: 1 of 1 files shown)


reference/toolstacks/php/drupal/redis-drupal-7.md
```diff
@@ -57,46 +91,50 @@ variables to match.
...
+   "sites/all/modules/contrib/redis/redis.autoload.inc"
 ]
 ```
 
 > **note**
 > Remember to tick the JSON Value box.
 
 > **note**
-> Use the actual path to your Redis module in case it is in a different location. For example: `sites/all/modules/contrib/redis`.
+> Use the actual path to your Redis module in case it is in a different location. For example: `sites/all/modules/redis`. The
```
pjcdawkins: Actually Drush Make defaults to `contrib` only if you ask it to. But we do ask it to in our example (https://github.com/platformsh/platformsh-example-drupal/blob/7.x/project.make#L10).

pjcdawkins:
To me committing this kind of complex config to `settings.php` is really preferable to using the environment variables... do we _need_ to provide instructions for both?

sruggier:
I looked at the underlying docs themselves to compare the two approaches, and it looks like they already address this question:
```
 The advantage of using environment variables is that these won't be used
 in your local build where you might not have Redis installed.
```

So I'd suggest switching the order around so that the easy way is first, but explain that if you use local builds, it may be necessary to use variables instead. Alternatively, if the PHP configuration could automatically handle both cases, then the variables wouldn't be needed, even when local builds are build used. That would let us get away with one suggested configuration method.

Crell:
The settings.php example code, however, does nothing if the corresponding environment variables aren't defined.  It already self-checks against that.  So I think that doc is a bit misleading to being with.  Cf:

~~~php
if (!empty($_ENV['PLATFORM_RELATIONSHIPS'])) {
~~~

Arguably we could add some more conditional checks there to be sure (and we likely should anyway, looking at it), but making the PHP code run only-on-platform is quite easy.

sruggier:
Yeah, in that case, there's no reason why someone would have to use the other way, right?

Crell:
I suppose only if they want to; I expect most Drupalers would favor the settings.php approach, but that's just a gut feeling.  It would be different for other application stacks than Drupal, of course, but that would then be its own documentation page anyway.

Crell:
We discussed this a bit in the public Slack channel. The consensus was to favor the settings.php variant, but keep the environment version around for completeness.  I will update the PR later today.

PR #299, opened 2016-06-13
Add example to sanitize database in non-master env

mrkschan:
Ref: https://platform.sh/2016/05/sanitize-your-database/


Crell:
Good thinking!

PR #301, opened 2016-06-13
Clarify locations passthru

mrkschan:


ralt:
`passthru` can be confusing, I'd try to explain it in more detail separately.

Crell:
@ralt I'm not sure I follow. You want to update this PR, replace it with another, you'll write something elsewhere...?

ralt:
@Crell the PR can be merged as is, it's already an improvement. I was just thinking that the property could be explained in more detail, such as explaining where the file exists, etc.

PR #302, opened 2016-06-22
Add followup steps to serve robots.txt

mrkschan:


Crell:
How would one do that? Wouldn't that be just a matter of having a robots.txt file in your web root, or an application route that responds to that path?

mrkschan:
Yes, having `robots.txt` in the web root and `.platform.app.yaml` is configured to serve it. Re-deployment is also required after running the CLI.

Crell:
Then there's no real configuration to be done, is there?  You just... don't block it (via the cli command) and everything just works.  The text as is implies there's something else that I need to do, but there isn't.

mrkschan:
@Crell not really ... I saw some `.platform.app.yaml` does not use `allow: true` in "/"... For instance https://github.com/platformsh/platformsh-docs/blob/master/.platform.app.yaml#L14-L29

Crell:
Oh geez.  Well, then we need to either flesh out the description here of what it means to configure your app.yaml, or have some other page on the subject we can link to with a brief description here.  As is, the reader is left hanging.

mrkschan:
@Crell how's this now?

Crell:
Hm. It largely just pushes the question to an already not-entirely-clear section. But that section needs to be cleaned up anyway, and single-sourcing the text makes sense, so let's do this. Thanks. :smile:

PR #303, opened 2016-06-22
Mention Redis eviction policy

mrkschan:


reference/services/redis.md
```diff
@@ -8,6 +8,10 @@ cache area for your application.
 * 2.8 (default)
 * 3.0
 
+> **notes**
+> Redis is configured to serve as a LRU cache with the eviction policy `allkeys-lru` - its storage is not persistent. You should not use it as a database.
+
+> **notes**
```
Crell: These should each be "note", singular.  Looks good otherwise.

mrkschan: agree .. was sticking to the original one :)

mrkschan:
@Crell revised

PR #304, opened 2016-06-22
Add notice for using self hosted SMTP server

mrkschan:


reference/email.md
```diff
@@ -16,11 +16,15 @@ Emails from Platform.sh are sent via a SendGrid-based SMTP proxy. Each
...
 
->**note:** 
+> **note:** 
 > When you send email you should use the `mail` facility. There is no SMTP 
 > server running on localhost.
 
->**note:**
+> **note:**
+> If you choose to use your own SMTP server, do not connect to it
+> via port 25 since we block it for security reason.
```
Crell: Security reasonS, plural.

Crell:
I'm a bit concerned about the number of notes next to each other. They seem to be formatting badly. Also, having so many of them like that cheapens their value.  Can any of those be folded together into prose?

mrkschan:
@Crell revised, how's it now?

Crell:
Much nicer, thanks!

PR #306, opened 2016-06-22
Add Drupal 8 Redis instructions

Crell:
It's still alpha, but let's get out in front of it.

### DIFF
--- a/reference/toolstacks/php/drupal/redis-drupal-8.md
+++ b/reference/toolstacks/php/drupal/redis-drupal-8.md
@@ -14,9 +14,9 @@ Platform.sh
 
 ## Requirements
 
-### Add a redis service
+### Add a Redis service
 
-First you need to create a redis service.  In your `.platform/services.yaml` file,
+First you need to create a Redis service.  In your `.platform/services.yaml` file,
 add or uncomment the following:
 
 ```yaml
@@ -141,7 +141,7 @@ that on Platform.sh. Our Redis instance is not persistent, which means every dep
 would result in an emptied queue and possible lost items when the Redis instance
 gets full. It is, however, safe for transient information like caching and locking.
 
-### Verifying redis is running
+### Verifying Redis is running
 Run this command in a SSH session in your environment `redis-cli -h redis.internal info`. You should run it before you push all this new code to your repository.
 
 This should give you a baseline of activity on your Redis installation. There should be very little memory allocated to the Redis cache.
--- a/reference/toolstacks/php/drupal/redis-drupal-8.md
+++ b/reference/toolstacks/php/drupal/redis-drupal-8.md
@@ -128,6 +128,9 @@ if (!empty($relationships['redis'][0]) && extension_loaded('redis')) {
   // ensure this is also the case for a specific site before enabling this.
   // $settings['cache']['bins']['default'] = 'cache.backend.chainedfast';
 
+  // The static cache should always use a memory cache, not Redis
+  $settings['cache']['bins']['static'] = 'cache.backend.memory';
+
   // Set a fixed prefix so that all requests share that, the default is currently not reliable.
   $settings['cache_prefix'] = 'prefix_';
 }
--- a/reference/toolstacks/php/drupal/redis-drupal-8.md
+++ b/reference/toolstacks/php/drupal/redis-drupal-8.md
@@ -88,6 +88,7 @@ if (!empty($_ENV['PLATFORM_RELATIONSHIPS']) && extension_loaded('redis')) {
   if (!empty($relationships['redis'][0])) {
     $redis = $relationships['redis'][0];
 
+    // Set Redis as the default backend for any cache bin not otherwise specified.
     $settings['cache']['default'] = 'cache.backend.redis';
     $settings['redis.connection']['host'] = $redis['host'];
     $settings['redis.connection']['port'] = $redis['port'];
@@ -95,6 +96,12 @@ if (!empty($_ENV['PLATFORM_RELATIONSHIPS']) && extension_loaded('redis')) {
     // Enable the redis server overrides, e.g. cache tags checksum.
     // See that file, you might want to copy and customize it, it is meant as an example
     // and will change as more backends are offered by the redis module.
+
+    // Apply changes to the container configuration to better leverage Redis.
+    // This includes using Redis for the lock and flood control systems, as well
+    // as the cache tag checksum. Alternatively, copy the contents of that file
+    // to your project-specific services.yml file, modify as appropriate, and
+    // remove this line.
     $settings['container_yamls'][] = 'modules/contrib/redis/example.services.yml';
 
     // Manually add the classloader path, this is required for the container cache bin definition below
@@ -102,6 +109,10 @@ if (!empty($_ENV['PLATFORM_RELATIONSHIPS']) && extension_loaded('redis')) {
     $class_loader->addPsr4('Drupal\\redis\\', 'modules/contrib/redis/src');
 
     // Use redis for container cache.
+    // The container cache is used to load the container definition itself, and
+    // thus any configuration stored in the container itself is not available
+    // yet. These lines force the container cache to use Redis rather than the
+    // default SQL cache.
     $settings['bootstrap_container_definition'] = [
       'parameters' => [],
       'services' => [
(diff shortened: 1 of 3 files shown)


reference/toolstacks/php/drupal/redis-drupal-8.md
```diff
@@ -0,0 +1,154 @@
...
+```
+
+The key (left side) is the name that will be exposed to the application in the PLATFORM_RELATIONSHIPS
+variable.  The right hand side is the name of the service we specified above (`rediscache`) and
+the type of endpoint (`redis`).  If you named the service something different above, change `rediscache`
+to that.
+
+### Add the Redis PHP extension
+
+If you're using PhpRedis you will need to enable the PHP Redis extension.  In your `.platform.app.yaml` file,
```
pjcdawkins: The first sentence is a bit weird: PhpRedis _is_ the PHP Redis extension

reference/toolstacks/php/drupal/redis-drupal-8.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Configuration
+
+To make use of the Redis cache you will need to set some Drupal variables. The
+configuration is a bit more complex than can be easily represented in
+Platform.sh's environment variables configuration, so using `settings.php` directly
+is the recommended approach.
+
+Place the following at the end of `settings.php`, after the include directive
+for `settings.local.php`.  Note the inline comments, as you may wish to customize
```
pjcdawkins: Probably not, on Drupal 8 we provide `settings.platformsh.php`. I'd suggest decoding PLATFORM_RELATIONSHIPS a second time so you don't depend on the include order.

```php
if (!empty($_ENV['PLATFORM_RELATIONSHIPS'])) {
  $relationships = json_decode(base64_decode($_ENV['PLATFORM_RELATIONSHIPS']), TRUE);
  // etc.
```

reference/toolstacks/php/drupal/redis-drupal-8.md
```diff
@@ -0,0 +1,154 @@
...
+}
+```
+
+The `example.services.yml` file noted above will also use Redis for the lock and flood
+control systems.
+
+Although the redis module is able to use Redis for a queue backend, we do not recommend
+that on Platform.sh. Our Redis instance is not persistent, which means every deploy
+would result in an emptied queue and possible lost items when the Redis instance
+gets full. It is, however, safe for transient information like caching and locking.
```
pjcdawkins: Is locking transient information? (do you mind if your front door randomly unlocks itself?)

Crell: Hm. That's a very good point. Drupal's lock system is intended as a semaphore for actions like route rebuilding and such.  My thinking is that the site is going to hiccup slightly when the container is swapped out anyway, so a the potential for a race condition there is very slight.  However, I am not certain of that.  @Berdir, your thoughts?

reference/toolstacks/php/drupal/redis-drupal-8.md
Crell: It looks like that's what we recommend in the Drupal 7 instructions, so I agree it makes sense to be consistent.

reference/toolstacks/php/drupal/redis-drupal-8.md
Berdir: Hm. At least the non-persistent locks unlock themself at the end of the request anyway, so I don't see a problem there. The persistent lock could be a problem, that's e.g. used for config import, but I doubt you could do something like that successfully during a re-deploy anyway? It is however possible that sites have custom uses that do need to persist for a longer time. I already mention in my comment that sites should review and possibly customize that file (by copying it and changing it), you could mention the persistent lock as an example for that.

pjcdawkins: Maybe this page should link to our main description of Redis, which also includes a warning (see https://github.com/platformsh/platformsh-docs/blob/master/reference/services/redis.md)

reference/toolstacks/php/drupal/redis-drupal-8.md
```diff
@@ -0,0 +1,163 @@
...
+          'arguments' => ['container'],
+        ],
+        'cache_tags_provider.container' => [
+          'class' => 'Drupal\redis\Cache\RedisCacheTagsChecksum',
+          'arguments' => ['@redis.factory'],
+        ],
+      ],
+    ];
+
+    // Explicitly set the fast backend for bootstrap, default, discover and config,
```
pjcdawkins: This should be `bootstrap, discovery and config`, no?

reference/toolstacks/php/drupal/redis-drupal-8.md
Crell: I've added the persistent lock as an example that may belong in MySQL, and added a link to the Redis main service page.

reference/toolstacks/php/drupal/redis-drupal-8.md
```diff
@@ -0,0 +1,163 @@
...
+    // otherwise this gets lost when redis is enabled.
```
pjcdawkins: `redis` -> `the Redis module`

reference/toolstacks/php/drupal/redis-drupal-8.md
```diff
@@ -0,0 +1,163 @@
...
+    $settings['cache']['bins']['bootstrap'] = 'cache.backend.chainedfast';
+    $settings['cache']['bins']['discovery'] = 'cache.backend.chainedfast';
+    $settings['cache']['bins']['config'] = 'cache.backend.chainedfast';
+    // Default does not use the fast chained backend by default, but core only uses it for small caches
```
pjcdawkins: This comment gives me a bit of a headache.

* What does "Default does not use" mean?
* What do you mean by "core only uses it"? Isn't Redis a contrib module?
* What are the "small caches"?
* Should this be 2 sentences, split at "Ensure"?

reference/toolstacks/php/drupal/redis-drupal-8.md
Crell: Technically you don't have to enable the redis module; that's what all of this hazerai is about, wiring up Redis without enabling the module. (This is what happens when you need more than vanilla composer for autoloading, too.)

reference/toolstacks/php/drupal/redis-drupal-8.md
```diff
@@ -0,0 +1,163 @@
...
+          'arguments' => ['container'],
+        ],
+        'cache_tags_provider.container' => [
+          'class' => 'Drupal\redis\Cache\RedisCacheTagsChecksum',
+          'arguments' => ['@redis.factory'],
+        ],
+      ],
+    ];
+
+    // Explicitly set the fast backend for bootstrap, default, discovery and config,
```
pjcdawkins: This comment says `default` but then default isn't used below

reference/toolstacks/php/drupal/redis-drupal-8.md
pjcdawkins: What does "redis is enabled" mean, then?

reference/toolstacks/php/drupal/redis-drupal-8.md
Crell: @Berdir I defer to you here.

reference/toolstacks/php/drupal/redis-drupal-8.md
Crell: @Berdir I defer to you here.

Berdir: I see that I failed to describe this here ;)

Simply put, this is a workaround for this core bug: https://www.drupal.org/node/2753989, which I finally managed to create. Maybe you simply want to write that.

The part about default goes a step further, it is an optimization that we use on our sites because we noticed that it only contains small amounts of data (= small caches) that are often requested.

Redis is the cache backend provider, it doesn't *use* the cache. Everyone else does. The reason it is commented out here and has some confusing descriptions is because most of this configuration is based on what we are using and I quickly extended it a bit with a few additional comments. Feel free to remove the part about default completely, it possibly doesn't belong into a minimal configuration guide.

I will instead open a core issue to move more of the bootstrap-ish cache entries into the bootstrap bin.

Crell: Well that got committed fast. :smile: 

@Berdir So if I follow you correctly, as of 8.2 this entire block from line 127 to 137 can go away in 8.2?  It's all just to work around that bug?  If so, I'll definitely switch up the description and mark it to remove later.

Berdir: Correct. The only thing that's not removable due to the that issue is the bit about the default bin. But that's clearly too complicated to explain and I'm already working on moving the most often requested parts of the default cache bin to bootstrap in https://www.drupal.org/node/2754003.

So what I'd recommend is you drop the lines about default bin completely already and just say that the others are a workaround for that issue and can be removed in 8.2.0

Crell:
I've reorganized and updated the documentation for the code snippet.  It should be clearer now what parts do what, and the part at the end is marked for removal in 8.2.

@Berdir Go/no-go on if this is sufficiently safe/accurate to merge?

reference/toolstacks/php/drupal/redis-drupal-8.md
```diff
@@ -0,0 +1,172 @@
...
+    // Enable the redis server overrides, e.g. cache tags checksum.
+    // See that file, you might want to copy and customize it, it is meant as an example
+    // and will change as more backends are offered by the redis module.
+
+    // Apply changes to the container configuration to better leverage Redis.
+    // This includes using Redis for the lock and flood control systems, as well
+    // as the cache tag checksum. Alternatively, copy the contents of that file
+    // to your project-specific services.yml file, modify as appropriate, and
+    // remove this line.
+    $settings['container_yamls'][] = 'modules/contrib/redis/example.services.yml';
```
Berdir: seems like there are two separate blocks above that now describe this?

Berdir:
Everything looks fine to me apart from the comment above.

The cache_prefix line is technically also only a workaround that will no longer be required when fixed (there's an issue on github about it but no d.o issue yet). However, the prefix will then be a (partial) hash, and I prefer having a short, clear over a hash even if it would no longer be required to set it. (without it, different domains and drush use different prefixes right now).

reference/toolstacks/php/drupal/redis-drupal-8.md
Crell: Oops. I forgot to delete the old comment block.

Crell:
I tweaked the comment on the prefix a bit, and removed the outdated comment block.

@pjcdawkins I think this is ready to merge at your earliest convenience.

reference/toolstacks/php/drupal/redis-drupal-8.md
```diff
@@ -0,0 +1,169 @@
...
+        ],
+      ],
+    ];
+
+    // Set a fixed prefix so that all requests share the same prefix, even if
+    // on different domains.
+    $settings['cache_prefix'] = 'prefix_';
+
+    // Drupal 8.1 has a bug where certain special caches that should use the
+    // APUc cache if available will not do so if a non-SQL default is specified.
```
pjcdawkins: Tiny typo here, `APUc` should be `APCu`

PR #309, opened 2016-06-24
I believe it should be password

OriPekelman:


Crell:
The mysql cli docs agree: EIther -p or --password.

PR #311, opened 2016-06-29
cache_field needs to be transactional

pjcdawkins:
@damz pointed this out - see https://www.drupal.org/node/1679344


Crell:
Sigh, Drupal.

PR #313, opened 2016-06-30
Remove all remaining references to document_root

Crell:
Excluding the upgrading guide.  This should make us internally consistent at least.

### DIFF
--- a/guides/configuration/drupal_application.md
+++ b/guides/configuration/drupal_application.md
@@ -14,8 +14,10 @@ access:
 relationships:
     database: "mysql:mysql"
 web:
-    document_root: "/"
-    passthru: "/index.php"
+    locations:
+        "/":
+            root: "web"
+            passthru: "/index.php"
 disk: 2048
 mounts:
     "/public/sites/default/files": "shared:files/files"
@@ -29,4 +31,4 @@ crons:
 
 In this example, we are only using the ``mysql`` service. You can use more services (redis, solr...) if you need.
 
-We are also supposing that your Drupal codebase is at the root of your Git repository. If it's not the case, you can change the ``document_root`` to the proper folder where your Drupal files are located.
+This example presumes that your Drupal codebase is in the "web" directory of your repository. If it's not the case, you can change the ``root`` to the proper folder where your Drupal files are located. Note that putting your Drupal webroot at the root of the repository is not supported.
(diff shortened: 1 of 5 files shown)


pjcdawkins:
AFAIK a document_root of `/` translates to a root of `public`, possibly in our BC code and at least in our other Drupal examples

guides/configuration/drupal_application.md
```diff
@@ -29,4 +31,4 @@ crons:
 
 In this example, we are only using the ``mysql`` service. You can use more services (redis, solr...) if you need.
 
-We are also supposing that your Drupal codebase is at the root of your Git repository. If it's not the case, you can change the ``document_root`` to the proper folder where your Drupal files are located.
\ No newline at end of file
+This example presumes that your Drupal codebase is in the "web" directory of your repository. If it's not the case, you can change the ``root`` to the proper folder where your Drupal files are located. Note that putting your Drupal webroot at the root of the repository is not supported.
```
mrkschan: Having Drupal `index.php` at `/app` is supported but there will be complicated `web.locations` (see below). I think we should be saying that "is not recommended" instead.

```
locations:
  "/":
    root: ""
    passthru: "index.php"
    allow: false
  "sites/default":
    root: "public/sites/default"
    allow: true
  "sites/all":
    root: "sites/all"
    allow: true
```

pjcdawkins: "Not supported" doesn't mean "it doesn't work", it can just mean "we won't help you if you have problems"

mrkschan: I think that's the question of not supported by us or not supported by the system :)

sruggier: Usually, I assume it's the former if something that happens to work is described in the documentation as not supported.

Crell:
@pjcdawkins The Drupal and Symfony example repos all use locations now, not document_root.  After this PR, we can basically view document_root as unsupported for anything new. It's just for pre-existing projects.

pjcdawkins:
@Crell sure. But our Drupal 7 build flavor still builds into `public` which is why we have
https://github.com/platformsh/platformsh-example-drupal/blob/7.x/.platform.app.yaml#L35

I just wondered what's more normal between `public` and `web`. I think `web` is probably better: the name `public` is unnecessarily alarming in our model, when things are only public if you have configured them to be so.

So, ignore me.

guides/configuration/drupal_application.md
Crell: I'm inclined to go with the former definition here, unless someone can suggest better wording for this sentence.  (Right now it kinda works, mostly, but the results may not always work as intended.)

Crell:
So what are we waiting on for this PR?

PR #314, opened 2016-07-01
State that persistent disk cannot be downsized

mrkschan:
Also drop tips on reacting to a already helpful error message.

### DIFF
--- a/reference/platform-app-yaml.md
+++ b/reference/platform-app-yaml.md
@@ -225,7 +225,7 @@ The `disk` defines the size of the persistent disk of the
 application in MB.
 
 > **Note**
-> The minimum recommended disk size is 256MB. If you see the error `UserError: Error building the project: Disk size may not be smaller than 128MB`, increase the size to 256MB.
+> The minimum recommended disk size is 256MB. Currently, we do not support downsizing a persistent disk.
 
 
 ### Mounts
--- a/reference/platform-app-yaml.md
+++ b/reference/platform-app-yaml.md
@@ -225,7 +225,7 @@ The `disk` defines the size of the persistent disk of the
 application in MB.
 
 > **Note**
-> The minimum recommended disk size is 256MB. Currently, we do not support downsizing a persistent disk.
+> The minimum recommended disk size is 256MB. Currently, we do not support downsizing a persistent disk. The actual disk size will be the maximum that has been deployed to master at any point.
 
 
 ### Mounts


Crell:
As written, this asks more questions than it answers.  So what happens if I deploy with a 2048 disk, then realize that was dumb and resize it to 256?  Does it silently stay at 2048 forever, and I can do nothing about it, ever? Can I ask for it to be downsized via support?  How do I reclaim that space to use for something else?

(This is a real case; the php-versions repo originally had a huge default disk, even though it uses NO disk. I reduced the value there, but does that mean it's still eating up 2 GB each for 4 app containers that consist of nothing but a `phpinfo()` script?)

Let's add some more details here.

mrkschan:
@crell try downsizing! You will see a rare Exception in the build logs and Platform does not support downsizing storage currently (manual intervention does not work as well).

Crell:
Don't answer me, update the PR! :smile: (I actually did not get an error when downsizing those containers, perhaps because the shared files mounts were empty?)

mrkschan:
We just don't support downsizing storage volume currently, what else do you expect to put on the docs? @Crell

mrkschan:
@Crell i'm not able to reproduce the exception but the situation is like this...

1. I create a container with `size: 2048`.
2. I `dd if=/dev/zero of=/app/mount count=1024 bs=1M`.
3. I change .platform.app.yaml to `size: 1500`.
4. Build succeeded but `df -h` shows the following.

```
/dev/mapper/platform-w6ikvtghgyuty--symfony3----php       2.0G  1.1G  909M  54%
```

Would you like "Currently, downsizing a persistent disk is ignored." instead?

Crell:
What does "don't support" mean in this case?  Does it mean "will error out" or "will be ignored"?  Currently that's unclear, but is a very important distinction.

robertDouglass:
Well, you could argue that it means "don't do it", which is what the
message should be.

mrkschan:
Please come up with whatever phrase that ask them "don't do it" :)

Crell:
"Don't do it" without an "or else" is not meaningful.  What bad thing happens?  If we're going to tell people not to do something, we need to give some reason for it.

Crell:
Tweaked language per discussion with @damz. Someone commit if this is now accurate.

mrkschan:
@robertDouglass please review the changes.

Crell:
After some discussion in chat, we're going to skip this PR after all.

PR #315, opened 2016-07-01
Add note to downsizing disk of a service

mrkschan:


Crell:
As in #314, what happens if you try? Error? It silently ignores you? That should be included in the docs in both places.

Crell:
Tweaked language per discussion with @damz. Someone commit if this is now accurate.

mrkschan:
@robertDouglass please review the changes.

Crell:
@mrkschan Please merge if you're comfortable here.

sruggier:
I've merged the original commit.

Crell:
Er?  The original commit was incomplete, as noted, as it actively left the user wondering what would happen.  That's why we added more text.  Please don't merge something that's incomplete.

PR #317, opened 2016-07-01
The root is probably not required actually

pjcdawkins:
The default is definitely `null`, so I figure `null` (or rather, omitting the key) is probably valid. Use case: when your app has no static files at all in this location

- [ ] needs testing


sruggier:
This is true, but the current behaviour for sub-locations will change soon - we plan on making them inherit from the parent location if no root is specified, so it may be good to discourage people from omitting it until that behaviour is finalized.

Crell:
Let's sit on this until the behavior stabilizes.  @sruggier is there somewhere we can track that?

pjcdawkins:
Not the place to discuss it but just noting my $0.02: I think inheritance is confusing, especially when there could be a valid use case for "no root". It would need to have some benefits beyond saving a few characters of typing.

mrkschan:
I think we better just say that's required...

```
web:
  locations:
    "/":
      passthru: "/index.php"
```

I guess the above config is valid but the `passthru` would never work since no `root` is specified.

pjcdawkins:
KS, I was talking about apps with no specific file locations: no static files, and a passthru of `true`

mrkschan:
In that case... is it better to set `root: ""` and `allow: false` to keep the docs simple to understand where root is always required?

PR #318, opened 2016-07-12
Add a list of inbound IP addresses

mrkschan:


reference/public-ip-addresses.md
```diff
@@ -18,3 +20,23 @@ North American region (us.platform.sh):
 These addresses are stable, but not guaranteed to never change. Prior to
 any future change, all affected customers will receive ample warning.
 
+## Inbound IP addresses
+
+Platform.sh regions accept connections on a limited number of IP
+addresses. You may need to whitelist them on your corporate firewall
+to enable SSH connections.
```
Crell: This doesn't make sense. We accept inbound connections from any IP at all.  Even incoming SSH I can connect from an arbitrary IP, as long as I have keys setup.  What are we talking about here?

ralt: If you were behind a firewall, you would need to add platformsh's IPs in the outgoing rules.

Crell: Ahhh...  That's not what the text here implies. The intro text should therefore be rewritten.  (The issue isn't on our end, it's on your end if your firewall limits outgoing ssh connections.)

pjcdawkins: > Platform.sh regions have a limited number of IP addresses.
?

mrkschan:
@Crell revised, how's that now?

Crell:
Thanks, @mrkschan!

PR #319, opened 2016-07-15
Mention the default value of passthru

mrkschan:


pjcdawkins:
The "with leading `/`" just applies to "absolute URI path", so I don't think you should include the default in the same parentheses.

Just add a new sentence "The default value is `false`".

mrkschan:
right, fixed.

PR #320, opened 2016-07-15
Drop no-ops passthru settings in example

mrkschan:
Since the example is using PHP (running fastcgi), `passthru: true` will have no effect at all (reset to false internally).


pjcdawkins:
Shouldn't it be `passthru: "/index.php"`? Drupal expects to receive the request if the file is not found.

mrkschan:
If we really want to forward the request if not found, yes, `passthru: "/index.php"`. Let's close this and create a new one if this's the case.

PR #321, opened 2016-07-15
Use correct passthru value to fallback to "/" location block

mrkschan:


Crell:
Are you sure that's right? The comment above suggests that it should fall through to the `/` entry anyway as is.  If that's not the case and we need to explicitly specify `/index.php` again, the comment should be updated, too.

mrkschan:
@Crell see https://github.com/platformsh/platformsh-docs/pull/320, `passthru: true` is not supported for fastcgi.

pjcdawkins:
Yes, the comment needs updating, too - there is no inheritance.

mrkschan:
@Crell @pjcdawkins I don't really get why the comment is having issue...

> a missing URL will passthru to the "/" location above, and hit the application front-controller.

`passthru: /index.php` is using the "/" location block, what's the problem?

pjcdawkins:
Because it has nothing particular to do with the `"/"` location block - there is no inheritance.

mrkschan:
The comment here does not mention inheritance ... and `/index.php` is really handled by "/" location block .... no?

> a missing URL will passthru to the "/" location above, and hit the application front-controller.

pjcdawkins:
The `/index.php` is handled independently by the `"/"` and `"/sites/default/files"` blocks. The blocks do not have any relationship with each other.

mrkschan:
I have this .app.yaml.

```
web:
    locations:
        "/":
            root: "public"
            passthru: "/index.php"
        "/f":
            root: "public/sites/default/files"
            passthru: "/index.php"
            allow: true
```

The corresponding Nginx config is something like ...

```
server {
  location = "/f" {
    alias /app/public/sites/default/files;
    try_files       $uri @94fcba4f84335cd9108c542d573a95c1e4286bcf;
  }
  location "/" {
    alias /app/public/;
    location "/" {
      try_files       $uri @94fcba4f84335cd9108c542d573a95c1e4286bcf;
    }
    location ~ ".+?\.php(?=$|/)" {
      try_files       $fastcgi_script_name @94fcba4f84335cd9108c542d573a95c1e4286bcf;
      fastcgi_pass    unix:/run/app.sock;
      include         "fastcgi_params.normal";
      fastcgi_split_path_info ^(/.+?\.php(?=$|/))((?:/.*)?)$;
      set $path_info  $fastcgi_path_info;
    }
  }

  location "@94fcba4f84335cd9108c542d573a95c1e4286bcf" {
    rewrite . /index.php;
  }
}
```

So, `passthru: /index.php` is URL rewrite to "/index.php" which is `app/public/index.php` from location '/'.

Crell:
Closing in favor of #323

PR #323, opened 2016-07-20
Don't use allow: true in / for Drupal

pjcdawkins:
Would supersede #321 

https://github.com/platformsh/platformsh-example-drupal/pull/38 needs review first

### DIFF
--- a/reference/platform-app-yaml.md
+++ b/reference/platform-app-yaml.md
@@ -31,7 +31,7 @@ Here is an example of a `.platform.app.yaml` file for Drupal:
                     - index.php
                 allow: false
                 rules:
-                    "^/robots\.txt$":
+                    '^/robots\.txt$':
                         allow: true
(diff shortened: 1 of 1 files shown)


reference/platform-app-yaml.md
```diff
@@ -201,23 +204,20 @@ It has a few subkeys, which are:
...
-                        expires: -1
-            # Set a 5 min expiration time for static files here; a missing URL
-            # will passthru to the "/" location above, and hit the application
-            # front-controller.
+                    "^/robots\.txt$":
+                        allow: true
             "/sites/default/files":
-                expires: 300
-                passthru: true
+                expires: 5m
```
mrkschan: root?

Crell:
This is now inconsistent with what we're doing in the Drupal examples, and we want to replace that example block with something more streamlined anyway.  Closing this PR.

PR #324, opened 2016-07-22
D8 redis example: include redis.services.yml

pjcdawkins:


Crell:
Is this a new thing in the upstream module?  It was working without that lien before... :smile: 

One issue that I think I'm running into is that Composer isn't recognizing module alpha releases, and is therefore downloading dev snapshots. That's... not reliable.  Is this the change that caused that issue?  (Ie, dev version of Redis module now requires this extra line)

pjcdawkins:
Hm. There is no relevant change upstream - at least, the `example.services.yml` and `services.yml` files haven't changed since early January.

Crell:
Curious. This fixes the crash we were running into at GovCon, which begs the question of how it ever worked in earlier testing. :smile: Well, this fixes it so merging.  Thanks, @pjcdawkins!

PR #329, opened 2016-07-28
Recommend putting the redis configuration in the platformsh settings file

Crell:
A couple of people at GovCon suggested they'd rather organize code this way, and it doesn't matter functionally, so may as well be organized.


JGrubb:
Hooray for organizing!

sruggier:
I don't see `settings.platformsh.php` mentioned anywhere else in the documentation. The documentation should link back to an explanation somewhere else that describes how to include that file in settings.php.

PR #330, opened 2016-07-29
Add note for sending email via /usr/sbin/sendmail

mrkschan:


pjcdawkins:
How does this information help the user?

To me it seems like a mere implementation detail.

mrkschan:
Yes, for PHP container. You don't need to know that. What about non-PHP container?

Crell:
This change doesn't really tell Node or Python users how they should then send email, does it?  We should generalize beyond PHP, but this doesn't provide much help for non-PHP users.

pjcdawkins:
May I correct:

>`/usr/sbin/sendmail` -> `sendmail`

> `mail` facility -> `mail()` facility

mrkschan:
Updated, @Crell @pjcdawkins. Note, `/usr/sbin/` is not included in `$PATH` by default and thus we need `/usr/sbin/sendmail`.

mrkschan:
Regarding `mail` vs `mail()`, i have no idea in PHP :) Please let me know which is the official one. Also, shall we correct the `file spool` facility as well?

reference/email.md
```diff
@@ -1,25 +1,28 @@
...
-Emails from Platform.sh are sent via a SendGrid-based SMTP proxy. Each
-Platform.sh project is provisioned as a SendGrid sub-account. Note that SendGrid
-sub-accounts are capped at 12k emails per month. If you need to adopt SPF for
-improving email deliverability, please read
-[SPF setup guidelines on SendGrid](https://sendgrid.com/docs/Glossary/spf.html).
+## Sending email in PHP
 
-When you send email you should use the `mail` facility. There is no SMTP 
-server running on localhost. For example in Symfony, if you use the default `SwiftMailer`,
+When you send email, you should use the built-in `mail` facility in PHP.
```
Crell: Let's say "the build-in `mail()` function in PHP".  No room for confusion that way.

reference/email.md
```diff
@@ -1,25 +1,28 @@
...
+PHP runtime is configured to send email via the assigned SendGrid sub-account.
```
Crell: Needs a "The" at the beginning.  And this doesn't need to be a new paragraph now.

Crell:
The file spool is a SwiftMailer thing, I think?  Not something PHP-specific, so I think that's fine as is.

reference/email.md
mrkschan: New paragraph? Do you mean the mention of Symfony in the next paragraph?

PR #333, opened 2016-08-03
Split Drupal into separate sections for Drupal 7 and Drupal 8

Crell:
The two are different enough that it makes little sense to treat them as a single platform.

The W3C link checker has verified that there's no new broken internal links.  (Links from elsewhere are much harder to verify, so not going to bother right now.)

This is ready for review and merging!


reference/toolstacks/php/drupal8/solr.md
```diff
@@ -0,0 +1,27 @@
...
+Go to `/admin/config/search/search_api` and click on *Add Server*. Give
+the server a name and a description.
+
+-   Server class: *Solr service*
+-   Solr host: `solr.internal`
+-   Solr port: `8080`
+-   Solr path: `/solr`
+
+> **note**
+> The name of the host depends on the name you the relationship in `.platform.app.yaml`. If
```
sruggier: There's a missing word here: `s/name you the/name you give the/`.

reference/toolstacks/php/drupal8/README.md
```diff
@@ -0,0 +1,93 @@
+# Getting Started
+
+## Drupal 8 and Composer
+
+The recommended way to deploy Drupal 8 on Platform.sh is to use Composer. [Composer](https://getcomposer.org/) is the PHP package management suite, and is now supported by Drupal 8 (and Drupal 7 in a pinch). There is an unofficial but well-supported Composer flavor of Drupal 8 called [Drupal Composer](https://github.com/drupal-composer/drupal-project) that we recommend.  If you use the [Drupal 8 Example Repository](https://github.com/platformsh/platformsh-example-drupal8/) or select the Drupal 8 option when creating a new project from a template, that's what you will be using.  You can also create your own project directly from that repository and add the Platform.sh-specific configuration files.  Note that you will also need to add the Drupal.org Composer repositories to `composer.json` if you are not working from our template.
```
pjcdawkins: The last two sentences seem a bit too much information perhaps? (these are quite long paragraphs, and the note about Drupal.org repositories is perhaps a rarely needed detail)

reference/toolstacks/php/drupal8/developing-with-drupal.md
```diff
@@ -0,0 +1,113 @@
...
+Here, we'll see how to make code changes to an environment.
+
+> **note**
+> You should never be working on the Master branch since it's supposed to be your production environment.
+
+Make sure you're on a working environment. In this example we're on the
+*sprint1* branch:
+
+```bash
+$ git branch
```
pjcdawkins: I would have expected to read just:

```bash
git checkout sprint1
```

(and I don't like `$` signs at the beginning of shell examples - it makes copy/paste harder - but it exists in plenty of other places in our docs)

Crell: Yeah, I wasn't sure about the $. I saw it in the existing docs so went with it.  If we want to look into dropping that it would be a separate PR.

reference/toolstacks/php/drupal8/README.md
Crell: I think it's an important note, since there's 2 different and incompatible Drupal Packagist repositories right now and neither one is included in the Drupal-Composer skeleton by default.  I agree the paragraph is a bit long, though.  I'll split it up.

reference/toolstacks/php/drupal8/developing-with-drupal.md
Crell: Hm. The current example is just checking what branch we're on, rather than changing to it.  It's in the current D7 docs and I didn't change it.  You're right, we probably should, though.  I will do so.

JGrubb:
Question - under the "most common case" config section for [Redis/Drupal 8](https://pr-333-t2llqeifuhpzg.eu.platform.sh/user_guide/reference/toolstacks/php/drupal8/redis.html) - should the `$relationships['redis']` be `$relationships['rediscache']` to be consistent with the services setup above that section?

OriPekelman:
For incoming links what I do generally is to go to google Analytics, pick the top something pages and verify they are fine. If they are not, and they have a bunch of visits ,I create redirects for those pages.

Also note that we have the magic redirect script that does fuzzy matching, we need to regenerate a current url list for it (though it is disliked by some).

Crell:
@JGrubb The relationship name isn't the service name.  It's the relationship name used in the app.yaml file.  Yes, it's confusing. :-(

So if you have, in app.yaml:

```yaml
relationships:
  foo: "bar:baz"

Then foo is the name the application sees, bar is the name in services.yaml, and baz is the protocol that the application uses to speak to bar (aka, what port to open).  At the moment, therefore, it should still be "redis".

JGrubb:
Ok, so in that case, I read over the entire thing for D8 and it all makes quite a lot of sense.  I presume D7 is similarly good to go.  Are you comfortable with my merging this Herculean MR?

PR #325, opened 2016-07-22
Specify 7.x version for registry-rebuild

hanoii:
Running the documented command on a project with drush 8.x fails:

```
$ drush dl registry_rebuild --destination=/app/tmp
No release history available for registry_rebuild 8.x.                                                                                                         [error]
Could not download requested project(s).
```

PR #334, opened 2016-08-05
Fix "Could not locate registry_rebuild version 7.x-7.x-dev"

pjcdawkins:
The change in #325 did not cover all cases. When inside the root of a D7 site, it seems Drush appends `-7.x-dev` if the version is not fully specified.


Crell:
Seriously, drush?

PR #338, opened 2016-08-09
Add clarification about non-master environments for robots.txt

Crell:
Since this tripped up a couple of us internally for a while, it's probably even more confusing to people on the outside. :smile:


mrkschan:
iirc, @pjcdawkins said we don't add `-p` or `-e` in docs ... please correct me if i'm wrong ... anyway, there is a WIP JIRA that will bring this to the Web UI :)

pjcdawkins:
That's right, but in this case `-e` could be appropriate

PR #340, opened 2016-08-11
Add note about changing app.name in app.yaml

JGrubb:


Crell:
I think we have this warning in a few places around the code, but it's good to include here, too.  Thanks.

reference/platform-app-yaml.md
```diff
@@ -61,7 +61,12 @@ Here is an example of a `.platform.app.yaml` file for Drupal:
 The `name` is the unique identifier of the application. Platform.sh
 supports multiple applications within a project, so each application
 must have a **unique name** within a project. The name may only be
-composed of lower case alpha-numeric characters. (a-z0-9).
+composed of lower case alpha-numeric characters. (a-z0-9).  *Be advised
+that changing the `name` of your app after it's been deployed will
+destroy all storage volumes, and thus is typically a Very Bad Thing to do.*
+It could be useful under certain circumstances in the early stages
+of development but you almost certainly don't want to change it on 
+a live project.
```
mrkschan: Remove the notes below then?

PR #341, opened 2016-08-11
Unfuddled structure

OriPekelman:
@Crell Created (again) an unfuddled structure. Needs love. Integrated home page in the structure, but this of course would have to change.
To test locally just `gitbook install && gitbook build`


Crell:
Deprecated in favor of #354

PR #347, opened 2016-08-15
Add note for move_to_root

mrkschan:


reference/upgrade/README.md
```diff
@@ -72,6 +72,8 @@ As we are aiming to always provide you more control and flexibility on how to de
 
 The `web` key is now a set of `locations` where you can define very precisely the behavior of each URL prefix.
 
+Note, we no longer move your application from "/" to "public/" automatically if new format is adopted.
```
pjcdawkins: if _the_ new format

mrkschan:
Updated :)

reference/upgrade/README.md
```diff
@@ -72,6 +72,8 @@ As we are aiming to always provide you more control and flexibility on how to de
 
 The `web` key is now a set of `locations` where you can define very precisely the behavior of each URL prefix.
 
+Note, we no longer move your application from "/" to "public/" automatically if the new format is adopted.
```
Crell: Nit: "if the new format is used".  Adopted sounds awkward in this case.

mrkschan: "when the new format is used"?

Crell: I think "if" is slightly more grammatically correct in this case, although both work.

mrkschan: Shall we add note to "fix" this. Like moving the application to public/ in Git or using Build hook to do so...

Crell: Eh, unless someone has asked about that specifically let's not over-complicate it.  This is in the upgrade guide, so its applicability fades over time by design. :smile:

Berdir: I was actually the one to ask this, because we were surprised that this "magically" changed (technically, it magically changed before and now it no longer does).

Adding a suggestion for a fix might be helpful, because it actually doesn't seem completely obvious, sometehing as simple as mv * public as build step for example doesn't work as public/sites is auto-created. Still working out the best approach for moving it.

pjcdawkins: You could
* Have it in `public` in the first place (in the repository)
* Possibly a good move command would be: `rsync --archive --exclude=public . public`

pjcdawkins: ... well it would be a lot more complicated, actually:

```sh
rsync --archive --remove-source-files --exclude=public --exclude=.git . public
# and then remove empty directories somehow
```

Crell: Just reorganizing the repo itself seems like the better move long-term. At least that's what we should put in the documentation.

mrkschan:
updated

PR #348, opened 2016-08-16
Use a simpler, Symfony-inspired example file

Crell:
Because the one that's there now is only applicable to Drupal 7 users.

### DIFF
--- a/reference/platform-app-yaml.md
+++ b/reference/platform-app-yaml.md
@@ -6,7 +6,7 @@ You control your application and the way it will be built and deployed on Platfo
 via a single configuration file `.platform.app.yaml` located at the root of your application
 folder inside your Git repository.
 
-Here'sa n example of a minimalist `.platform.app.yaml` file for PHP:
+Here's an example of a minimalist `.platform.app.yaml` file for PHP:
 
 ```yaml
 # The name of this app. Must be unique within a project.
--- a/reference/platform-app-yaml.md
+++ b/reference/platform-app-yaml.md
@@ -36,16 +36,19 @@ web:
             # The front-controller script to send non-static requests to.
             passthru: '/app.php'
         # Allow uploaded files to be served, but do not run scripts.
-        '/files'
+        # Missing files get mapped to the front controller above.
+        '/files':
+            root: 'web/files'
             scripts: false
             allow: true
+            passthru: '/app.php'
 
 # The size of the persistent disk of the application (in MB).
 disk: 2048
 
 # The mounts that will be performed when the package is deployed.
 mounts:
-    '/app/web/files': 'shared:files/cache'
+    '/app/web/files': 'shared:files/files'
 
 
 # The hooks that will be performed when the package is deployed.
--- a/reference/platform-app-yaml.md
+++ b/reference/platform-app-yaml.md
@@ -46,7 +46,8 @@ web:
 # The size of the persistent disk of the application (in MB).
 disk: 2048
 
-# The mounts that will be performed when the package is deployed.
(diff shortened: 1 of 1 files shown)


reference/platform-app-yaml.md
```diff
@@ -6,48 +6,57 @@ You control your application and the way it will be built and deployed on Platfo
 via a single configuration file `.platform.app.yaml` located at the root of your application
 folder inside your Git repository.
 
-Here is an example of a `.platform.app.yaml` file for Drupal:
+Here'sa n example of a minimalist `.platform.app.yaml` file for PHP:
```
mrkschan: "Here's an"

Crell: Thanks, fixed.

pjcdawkins:
I'd prefer single quotes, otherwise LGTM

Crell:
Converted to single-quotes.

reference/platform-app-yaml.md
```diff
@@ -6,48 +6,57 @@ You control your application and the way it will be built and deployed on Platfo
...
+# The left-hand side is the name of the relationship as it will be exposed
+# to the application in the PLATFORM_RELATIONSHIPS variable. The right-hand
+# side is in the form `<service name>:<endpoint name>`.
+relationships:
+    database: 'mysqldb:mysql'
+
+# The configuration of app when it is exposed to the web.
+web:
+    locations:
+        '/':
```
pjcdawkins: Should the `/` location have `allow: false`?

reference/platform-app-yaml.md
```diff
@@ -6,48 +6,57 @@ You control your application and the way it will be built and deployed on Platfo
...
+            # The public directory of the app, relative to its root.
+            root: 'web'
+            # The front-controller script to send non-static requests to.
+            passthru: '/app.php'
+        # Allow uploaded files to be served, but do not run scripts.
+        '/files'
```
pjcdawkins: This location needs a `root` (presumably `'web/files'`). And maybe it needs the `passthru` too?

reference/platform-app-yaml.md
```diff
@@ -6,48 +6,57 @@ You control your application and the way it will be built and deployed on Platfo
...
+            scripts: false
+            allow: true
+
+# The size of the persistent disk of the application (in MB).
+disk: 2048
+
+# The mounts that will be performed when the package is deployed.
+mounts:
+    '/app/web/files': 'shared:files/cache'
```
pjcdawkins: `cache` is probably not accurate? `shared:files/files` or `shared:files/public-files` would be clearer to me

reference/platform-app-yaml.md
Crell: That's debatable, as there's use cases either way.  I'm going to say no for now, as this is just an example and we should keep it simple-ish.

reference/platform-app-yaml.md
```diff
@@ -6,48 +6,60 @@ You control your application and the way it will be built and deployed on Platfo
...
+            scripts: false
+            allow: true
+            passthru: '/app.php'
+
+# The size of the persistent disk of the application (in MB).
+disk: 2048
+
+# The mounts that will be performed when the package is deployed.
+mounts:
+    '/app/web/files': 'shared:files/files'
```
pjcdawkins: The mount path resolves of course to `/app/app/web/files`, so this seems like a source of confusion. Is the mount allowed to look like a relative path? (`app/web/files` without the leading slash?)

Crell: I totally goofed there, it should just be web/files.

PR #350, opened 2016-08-16
Rename remaining app container examples to app

Crell:
That way we're consistent everywhere.


pjcdawkins:
These should ideally just link to examples elsewhere on our GitHub so that there is less duplication. But this change is good.

Crell:
Yes they should, but GitBook is making that difficult. :smile: Separate patch.

PR #352, opened 2016-08-31
Clarify we are talking about the MRI

OriPekelman:


sruggier:
I think we can clarify this point in a simpler way if we add a header so the outline looks like this:
### Supported versions
#### Ruby MRI
* 2.3

Crell:
What's an MRI?

JGrubb:
@Crell "Matz' Ruby Interpreter" - the original implementation of Ruby by the creator, as opposed to JRuby or Rubinius or YARV or any of the alternative Ruby implementations out there.

OriPekelman:
Done in https://github.com/platformsh/platformsh-docs/commit/35ee0a46d8d0af9db6f5a62e301499b15dbb7a68

Issue #355, opened 2016-09-02
Improve and Update Drupal Migration Docs


drclaw:
Hi,

I'm just going through the process of importing an existing Drupal site into platform and I think the [the docs](https://docs.platform.sh/drupal_migrate/guides/type/php/drupal/migrate/init.html) are a bit out of date:

**Page Order**

[Since July 2016](https://docs.platform.sh/user_guide/reference/upgrade/index.html#changes-in-version-2016-4) you can't actually push your code to platform without having the configuration files already created and committed to the repo. However, [this step in the tutorial](https://docs.platform.sh/drupal_migrate/guides/getting-started/import-code.html) comes _before_ [the configuration step](https://docs.platform.sh/drupal_migrate/guides/configuration/index.html) which is going to be confusing for users following step-by-step (like me).

Additionally the image in the "Import your codebase" step does not quite match what I'm seeing. This isn't that big a deal but it did lead me to consider that maybe the documentation was out of date.

**The "Public" or "Web" subdirectory**

Initially reading through the docs (the drupal tutorial and the main docs), I was confused by various references to a "public" directory. As it turns out I think platform used to automatically move the application files into a public subdirectory, but [that changed sometime earlier this year](https://docs.platform.sh/user_guide/reference/upgrade/index.html#changes-in-version-2016-3). 

However, it's still required that your code _be_ in a subdirectory which isn't actually mentioned in the "import your codebase" step. I think it should be mentioned earlier in the docs (maybe even as a prerequisite?) that your Drupal code cannot be at the root of the repository as many people's existing codebases may not be setup this way.

Thanks!

Crell:
Hi @drclaw. I'm afraid when this issue was filed we were already in the process of eliminating the Drupal migration book, which rendered the page order changes obsolete.  Thank you for the PR, even if we weren't able to use it.

For the moment the migration guide has been replaced with https://docs.platform.sh/frameworks/drupal7/migrating.html. It doesn't cover as much, as the goal is to focus on data.  We're working on a better "getting started" section.

Meanwhile, if you spot any confusion about the "public" directory in the new docs can you file a PR for it?  We're not planning any more major overhauls in the foreseeable future, so they should be safe to file at this point.

Thanks for offering to help out!

PR #356, opened 2016-09-02
Update configuration docs

drclaw:
I posted an issue earlier today, but afterwards decided to see if I could get the ball rolling myself. For a detailed description of the issue, see #355 

Otherwise, I tried to give as much detail as possible in the commit messages so hopefully it explains itself.

Note that I don't actually know how to preview my changes in git book so I'm not sure if the changes I made will mess up the book compilation... mostly I was just editing existing pages so I assume those will be fine, but I do add a new page for setting up routes. Hopefully it works!

Thanks!


pjcdawkins:
Re. "Change the subdirectory from web to public for consistency", I think we should actually be consistent in:

* Drupal 8 docs: subdirectory should be `web`
* Drupal 7 docs: subdirectory should be `public`

Drupal 7 has historically used `public` and one of our legacy config formats defaults to `public`. Drupal 8, building via Composer, will use `web` because that's what the Drupal Composer Project chose (presumably because it follows the Symfony Framework example).

This is one reason why we probably shouldn't have _any_ docs in common between D7 and D8: they should be separated. cc @Crell

Crell:
With the massive reorganization of the docs we just posted I think most of these changes are no longer relevant.  If there are any that are still an issue in the new documentation structure please file a new PR for those.  Thanks.

PR #372, opened 2016-09-21
Don't hardcode the info for connecting the DB

mrkschan:


Crell:
I think we should at least link the new text to the appropriate page for the structure of the environment array, so people know how to get the domain out of it.

mrkschan:
What if we use the one like Redis everywhere? https://github.com/platformsh/platformsh-docs/blob/master/src/configuration/services/redis.md#using-redis-cli-to-access-your-redis-service

Crell:
Hm, that could work.  I'm all for better standardization across the services pages.

mrkschan:
Updated

Crell:
Thanks, KS!

PR #375, opened 2016-09-21
Add a copiable snippet for languages types & minor improvements

GuGuss:


src/languages/ruby.md
```diff
@@ -4,8 +4,8 @@ Platform.sh supports deploying any Ruby application. Your application can use
 any Ruby application server such as Unicorn or Puma and deploying a Rails or
 a Sinatra app is very straight forward.
 
-### Supported versions
-#### Ruby MRI
+## Supported versions
```
Crell: Ori specifically pushed to say MRI here.  I don't know Ruby well enough to know if that's sufficiently relevant to call out or not.

GuGuss: In any case, formatting is wrong so let's at least fix that. And since it's the only one we support and that's the one that everybody think of by default :), we should remove it.

Crell: @OriPekelman @sruggier Thoughts on ^^?

sruggier: I'm not totally convinced (yet?) that there's anything wrong with skipping header levels, but if that's a problem, then I think we should fix that without removing the mention of Ruby MRI: the one extra header is concise enough that it should be worth it to leave it in there.

Crell: We could say it's MRI without using an extra header for it.  Say, in the paragraph above.

JGrubb: Or a single item <ul>

- MRI

GuGuss: Fixed!

Crell:
Thanks, @GuGuss

PR #382, opened 2016-09-26
updated the routes.yaml

alam28:
I have added the routes.yaml file for 404s redirect

### DIFF
--- a/.platform/routes.yaml
+++ b/.platform/routes.yaml
@@ -34,6 +34,7 @@ https://{default}/:
       "/user_guide/overview/services.html": { "to": "/configuration/services.html", "code": "301", "prefix": false}
       "/drupal/guides/prerequisites/platform-cli.html": { "to": "/overview/cli.html", "code": "301", "prefix": false} 
       "/discover": { "to": "/", "code": "301", "prefix": false}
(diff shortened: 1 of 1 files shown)


Crell:
I just realized there's a typo in one of the current redirects: 

 "/user_guide/reference/paltform-app-yaml.html": { "to": "/configuration/app-containers.html", "code": "301", "prefix": false}

(paltform).  That's my bad.  Can we fix that in this PR as well?

Crell:
Here's the build error:

```
E: Error parsing configuration files:
    - routes.https://{default}/.redirects.key: path /discover/ conflicts with other path /discover

E: Error building the project: Invalid configuration files, aborting build.
```

Which is likely referring to this block:

```
+      "/discover": { "to": "/", "code": "301", "prefix": false}
 +      "/discover/": { "to": "/", "code": "301", "prefix": false}
```

Which suggests that double-entry is unnecessary.  @mrkschan, can you try it without that duplicate (and the other duplicates along the same lines) and see if deploy likes it?

mrkschan:
@crell we recently updated the validator.

We should drop `"/discover/": { "to": "/", "code": "301", "prefix": false}`.
There are other occurrence like `/user_guide/`.

@alam28 could you drop them all?

PR #383, opened 2016-09-26
removed the two fragments urls for sync and import files

alam28:
Hi, I removed fragment part from these two urls:
/drupal/guides/type/php/drupal/sql-sync.html
/drupal_migrate/guides/type/php/drupal/migrate/import-files.html


OriPekelman:
Also no reason to open a second pull-request for this.

PR #387, opened 2016-10-04
SSH URL now contains the [RANDOM-STRING]

GuGuss:


Crell:
I like this approach.  Much more sustainable going forward.

PR #389, opened 2016-10-12
Correct mcrypt extension inclusion status in PHP versions.

### LINKED ISSUE #388: PHP: mcrypt included by default or not?
Looking at https://docs.platform.sh/languages/php.html, it's indicated that mcrypt is "enabled by default".  However, when tinkering with Platform.sh I found that I had to specify it explicitly (was using PHP 7.0) or I got an error when Composer ran about mcrypt not being available.

I'd be happy to correct the docs via a PR, but I'm not sure whether this is a bug in the platform or the docs.

Relevant tidbits:
```
          - cakephp/cakephp 2.8.6 requires ext-mcrypt * -> the requested PHP extension mcrypt is missing from your system.

        To enable extensions, verify that they are enabled in those .ini files:
          - /etc/php/7.0/cli/php.ini
        You can also run `php --ini` inside terminal to see which files are used by PHP in CLI mode.
```

Fixed by adding the following to `.platform.app.yaml`
```
runtime:
    extensions:
        - mcrypt
```

themrwilliams:
Fixes #388

### DIFF
--- a/src/languages/php.md
+++ b/src/languages/php.md
@@ -55,7 +55,7 @@ The following extensions are enabled by default:
 * gd
 * curl
 * intl
-* mcrypt
+* mcrypt (5.6 and earlier)
 * zendopcache (5.4 only) / opcache (5.5 and later)
 
 You can disable those by adding them to the `disabled_extensions` list.


src/languages/php.md
```diff
@@ -83,7 +83,7 @@ This is the complete list of extensions that can be enabled:
 | intl         | *   | *   | *   | *   |
 | json         |     |     | *   | *   |
 | ldap         | *   | *   | *   | *   |
-| mcrypt       | *   | *   | *   | *   |
+| mcrypt       | *   | *   | *   |     |
```
Crell: mcrypt is still available in PHP 7.0, just not enabled by default, so it should still be in the table.  Just remove this hunk.

themrwilliams: My bad.  Better?

Crell: Yep, thanks!

PR #392, opened 2016-10-24
Document flavor: none

Crell:
And while we're at it, let's use it ourselves.

### DIFF
--- a/src/configuration/app-containers.md
+++ b/src/configuration/app-containers.md
@@ -103,11 +103,11 @@ The `type` defines what language will run your application.
 
 The `type` can be:
 
-* `php`
-* `hhvm`
-* `nodejs`
-* `python`
-* `ruby`
+* [`php`](/languages/php.md)
+* [`hhvm`](/languages/php.md)
+* [`nodejs`](/languages/nodejs.md)
+* [`python`](/languages/python.md)
+* [`ruby`](/languages/ruby.md)
 
 followed by a version.  See the appropriate language page for all available versions.
 
--- a/src/configuration/app-containers.md
+++ b/src/configuration/app-containers.md
@@ -119,7 +119,7 @@ The `build` defines what will happen by default when building the application.
 Its only property is `flavor`, which specifies a default set of build tasks to run. Flavors are language-specific.
 
 * PHP - default `composer`
-  * `composer` will run `composer install --optimize-imports --prefer-dist` if a `composer.json` file is detected.
+  * `composer` will run `composer --no-ansi --no-interaction install --no-progress --prefer-dist --optimize-autoloader` if a `composer.json` file is detected.
   * `drupal` will run `drush make` automatically if a `.make` file is detected.
 * NodeJs
   * `default` will run `npm prune --userconfig && npm install --userconfig` if a `package.json` file is detected.
--- a/src/configuration/app-containers.md
+++ b/src/configuration/app-containers.md
@@ -125,7 +125,7 @@ Its only property is `flavor`, which specifies a default set of build tasks to r
   * `composer` will run `composer --no-ansi --no-interaction install --no-progress --prefer-dist --optimize-autoloader` if a `composer.json` file is detected.
   * `drupal` will run `drush make` automatically if a `.make` file is detected.
 * Node.js
-  * `default` will run `npm prune --userconfig && npm install --userconfig` if a `package.json` file is detected.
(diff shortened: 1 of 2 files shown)


src/configuration/app-containers.md
```diff
@@ -116,11 +116,15 @@ The `type` can be:
...
-It has a sub property `flavor` for which the possible values are:
+Its only property is `flavor`, which specifies a default set of build tasks to run. Flavors are language-specific.
 
-* `drupal` means that `drush make` will automatically run if you provide `.make` files.
-* `composer` means that `composer install` will automatically run if you provide a `composer.json` or `composer.lock` file.
-* `symfony` is an alisas for `composer`.
+* PHP - default `composer`
+  * `composer` will run `composer --no-ansi --no-interaction install --no-progress --prefer-dist --optimize-autoloader` if a `composer.json` file is detected.
+  * `drupal` will run `drush make` automatically if a `.make` file is detected.
+* NodeJs
```
pjcdawkins: I'd prefer `Node.js` (https://www.google.co.uk/search?q=node.js)

Crell: You're right, we standardized that direction. I keep forgetting which one we use. Fixing.

Crell:
I also noticed we didn't have ruby listed under types, so I fixed that since it was right next to where we were.

src/configuration/app-containers.md
```diff
@@ -104,9 +104,12 @@ The `type` defines what language will run your application.
 The `type` can be:
 
 * `php`
+* `hhvm`
 * `nodejs`
 * `python`
-* `hhvm`
+* `ruby`
+
+followed by a version.  See the appropriate language page for all available versions.
```
pjcdawkins: "See page foo" should always have a hyperlink, IMO... (the language page should be linked for each `type`)

Crell: Can you even do links inside code quotes?  Or are you suggesting dropping the code quotes and making them links instead?

pjcdawkins: Well you can in HTML so I assumed you could in Markdown (code tag inside anchor tag) e.g. [`GitHub`](http://github.com)

Crell:
@pjcdawkins Hey, look at that, we can link the code snippets. Back at you.

src/configuration/app-containers.md
```diff
@@ -116,11 +119,15 @@ The `type` can be:
...
 The `build` defines what will happen by default when building the application.
 
-It has a sub property `flavor` for which the possible values are:
+Its only property is `flavor`, which specifies a default set of build tasks to run. Flavors are language-specific.
+
+* PHP - default `composer`
+  * `composer` will run `composer --no-ansi --no-interaction install --no-progress --prefer-dist --optimize-autoloader` if a `composer.json` file is detected.
+  * `drupal` will run `drush make` automatically if a `.make` file is detected.
+* Node.js
+  * `default` will run `npm prune --userconfig && npm install --userconfig` if a `package.json` file is detected.
```
pjcdawkins: Maybe just `npm prune && npm install`, because the --userconfig is actually `--userconfig <somefilename>` and it's probably a minor implementation detail.

Crell: @ralt Can you verify?  I saw the --userconfig in the source code, but I'm not entirely sure of the details as I'm not that familiar with npm.

src/configuration/app-containers.md
```diff
@@ -116,11 +119,15 @@ The `type` can be:
...
+* PHP - default `composer`
+  * `composer` will run `composer --no-ansi --no-interaction install --no-progress --prefer-dist --optimize-autoloader` if a `composer.json` file is detected.
+  * `drupal` will run `drush make` automatically if a `.make` file is detected.
+* Node.js
+  * `default` will run `npm prune --userconfig .npmrc && npm install --userconfig .npmrc` if a `package.json` file is detected. Note that this also allows you to provide a custom `.npmrc` file in the root of your application (as a sibling of the `.platform.app.yaml` file.)
 
-* `drupal` means that `drush make` will automatically run if you provide `.make` files.
-* `composer` means that `composer install` will automatically run if you provide a `composer.json` or `composer.lock` file.
-* `symfony` is an alisas for `composer`.
+In all languages you can also specify a flavor of `none` (which is the default for any language other than PHP and Node.js), which as the name suggests will take no action at all. That is useful when you want complete control over your build steps, such as to run a custom composer command or use an alternate Node.js package manager.
```
ralt: It isn't the default. The default is `default`, even for python and ruby. In practice, they do the same as the `none` flavor though.

PR #393, opened 2016-10-25
Document redirects with HTTPS instead of HTTP

GuGuss:


Crell:
I'm reluctant to give everyone copy-paste examples for SSL, when our master environments don't SSL by default and you need to go elsewhere to buy a cert.  That's setting a land mine for people.

Once we offer LetsEncrypt by default, this will be a good change.

Crell:
Closing for now, we'll reopen or do it again when we can assume free SSL everywhere.

PR #396, opened 2016-10-26
Add note for cache behavior change in static files

mrkschan:


pjcdawkins:
This has the old style web config in the YAML example- it should be updated with a new style "locations" block

mrkschan:
Provided example of both format

GuGuss:
I propose we also mention this change in the `expires` key definition with a link to this section.

src/configuration/app/upgrading.md
```diff
@@ -1,5 +1,45 @@
...
+```
+    web:
+        document_root: "/"
+        passthru: "/index.php"
+        index_files:
+            - "index.php"
+        expires: 300
+        whitelist:
+            - \.html$
+```
```
mrkschan: updated.

PR #402, opened 2016-11-07
This is PHP specific

OriPekelman:


pjcdawkins:
The relationship is not PHP-specific, just the extension. I've updated the PR.

PR #404, opened 2016-11-09
Provide more concrete instructions about upgrading PostgreSQL

sruggier:


Crell:
The new text looks fantastic, especially the sample code.  Nice work, @sruggier!

My only pushback would the the formatting.  The current contributing guide says to not hard-wrap text, as it makes updating it a much harder process.  Please remove the hard wraps and just let your editor wrap text on its own.  That makes maintenance easier.

sruggier:
Ok, I've updated the PR to remove the hard wrapping.

PR #406, opened 2016-11-11
postgresql: Add an alternative upgrade path using a dev environment

sruggier:
This could be useful to avoid exceeding plan limits.

### DIFF
--- a/src/configuration/services/postgresql.md
+++ b/src/configuration/services/postgresql.md
@@ -102,6 +102,14 @@ To minimize the downtime associated with such upgrades, we suggest the following
 8. Push the same commit from step 5 into your project's master branch. This will carry out the migration in your production environment.
 9. Finally, push out another commit that removes the old database service, which is no longer needed now that the upgrade process is finished.
 
+**Note**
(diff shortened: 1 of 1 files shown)


src/configuration/services/postgresql.md
```diff
@@ -95,13 +95,21 @@ To minimize the downtime associated with such upgrades, we suggest the following
...
 
+**Note**
+If your database is too large to duplicate within your plan size, we'd suggest temporarily raising your plan limits. Charges are prorated, so you'll only be billed for a single day of increased usage. However, you can also use a development environment to perform the migration:
+1. Put your master environment's application into maintenance mode, if possible, to prevent it from trying to read from or write to the database during the migration.
+2. Create an up-to-date copy of your database by cloning a development environment from master.
+3. Rename the master environment's postgresql service, and switch the version to 9.6. Renaming ensures that the service's storage is reset, which is necessary to avoid exposing the 9.6 server to incompatible data from version 9.3.
+4. Open an SSH session in your master environment, with agent forwarding enabled (so you can SSH from master to your development environment).
+5. Now, use SSH to run `pg_dump` on the development environment, and pipe the data from SSH into a `pg_restore` command running in the master environment.
```
Crell: Could we have an example command to do this?

sruggier: I've added more specific instructions about how to reach a command that will do it.

PR #408, opened 2016-11-12
Update mounts to clarify "share"

JGrubb:


src/configuration/app-containers.md
```diff
@@ -345,7 +345,7 @@ The format is:
 * `"/web/sites/default/files": "shared:files/files"`
 
 > **Note**
-> The `shared` means that the volume is shared between your applications inside an environment. The `disk` key defines the size available for that `shared` volume.
+> `shared` means that the volume is shared between builds of the same application, _not_ that the volume can be shared between different applications within a project. The `disk` key defines the size available for that `shared` volume.
```
kotnik: I agree with @pjcdawkins. We should also clarify that subfolders can be used in mounts too (ie. `/foo: "shared:files/foo/bar"`.

JGrubb: Wait sorry, subfolders _can_ be used in mounts?  Is that a new development, because that used to not be true, or so I was told.

pjcdawkins: It's new, since 8th November

JGrubb:
I'm closing this.  I suspect that the "correct" wording of this feature is too subjective, so I'm going to let someone else sort it out.

PR #409, opened 2016-11-14
Add documentation regarding php: prefixed environment variables

Crell:


xtfer:
I ended up reorganising this page, as when I tried to review it I made a number of basic errors trying to work with the variables. I believe the restructure resolves these. 

@Crell You can probably review yourself now.

cc @GuGuss

src/development/environment-variables.md
```diff
@@ -52,12 +73,7 @@ the runtime (*ie. PHP*) and prefixed with `PLATFORM_*`.
 * **PLATFORM_VARIABLES**: A base64-encoded JSON object which keys are variables names and values are variable values (*a string*).
 * **PLATFORM_PROJECT_ENTROPY**: A random value created when the project is first created, which is then stable throughout the project’s life. This can be used for Drupal hash salt, Symfony secret, or other similar values in other frameworks.
 
-Since values can change over time, the best thing is to just introspect the variable at runtime and use it to configure your application.
-
-For example with Drupal, we use the **PLATFORM_RELATIONSHIPS** variable
-to configure your `settings.local.php`.
-
-For example:
+Since values can change over time, the best thing is to just introspect the variable at runtime and use it to configure your application. For example:
```
pjcdawkins: "introspect" doesn't make sense here - let's fix that at the same time, perhaps just:

> the best thing is to use the variable at runtime to configure your application

src/development/environment-variables.md
```diff
@@ -3,40 +3,61 @@
...
+This allows you to define your development variables only once, and use them on all the child environments.
 
 > **note**
 > If a variable is added in the parent environment after the child environments are created, the child environments have to be re-deployed in order to load the inherited variables.
 
 ## Platform.sh variables
 
-Environment variables that are specific to Platform.sh are exposed in
-the runtime (*ie. PHP*) and prefixed with `PLATFORM_*`.
+Environment variables that are specific to Platform.sh are exposed in the runtime (*ie. PHP*) and prefixed with `PLATFORM_*`. These variables are used to provide a a range of information about your environment and services for use in application configuration. For example, for Drupal, we use the **PLATFORM_RELATIONSHIPS** variable to configure your `settings.local.php`.
```
Crell: The ie here should be an `e.g.`, since it's an example, not an "in other words". :smile:

src/development/environment-variables.md
```diff
@@ -3,40 +3,61 @@
...
+Custom environment variables which you have defined can be viewed [via the Web Interface](administration/web/configure-environment.html#settings) or using the CLI...
+
+```bash
+$ platform variables
+
++---------+-------+-----------+------+
+| ID      | Value | Inherited | JSON |
++---------+-------+-----------+------+
+| env:FOO | bar   | No        | No   |
++---------+-------+-----------+------+
```
Crell: This says the name is `env:FOO`, but the previous section just named it `foo`.  That's highly confusing. Is it what actually happens?  It seems not from the text below.

Crell:
If we're doing more cleanup here, we should also clean up the "Variables vs environment variables" question a bit; we have "variables" which get exposed as an array in environment variables, but we also have variables that get exposed directly as environment variables.  That's messy and confused both Fabien and I when doing prep for SensioCloud.

Since the original scope of this issue was just the `php:` variable support, we may want to push that to a follow-up.  At least the points above should be addressed, though.  @xtfer ?

pjcdawkins:
How about this terminology:

* Project variables (a Platform.sh concept, available at build-time and/or runtime)
* Environment variables (a Platform.sh concept, available at runtime only)
* System environment variables (a *nix concept)

src/development/environment-variables.md
xtfer: You are correct. They are actually different variables. I'll fix that.

Crell:
Environment Variables vs System Environment Variables sounds potentially confusing.  Previously we called them just "Variables"; it's my fault we started slipping "environment variables" into the UI. :smile: 

What about Project Variables, Platform Variables, and Environment Variables?  (Where Project and Platform variables end up in Environment variables in a variety of different ways that we'd need to document.)

pjcdawkins:
Using "Environment" to mean master/prod/dev/staging/local is very well established, we use it everywhere. Also using "Environment" to refer to the shell/application environment (`getenv`, `$_ENV` etc.) is also very well established. We can't get away from that.

src/development/environment-variables.md
```diff
@@ -96,17 +112,38 @@ echo $PLATFORM_RELATIONSHIPS | base64 --decode | json_pp
...
 
-For example with PHP, you can get that variable with `$_ENV['FOO']`.
+```ini
+$ platform variable:set env:foo bar
+```
+
+With PHP, you can get that variable with `$_ENV['FOO']` or `getenv('FOO')`.
 
-## Drupal specific variables
+### PHP specific variables
```
pjcdawkins: Ideally should be "PHP-specific variables"

Crell:
After discussion in Slack, let's postpone the definition question to a follow-up.

PR #410, opened 2016-11-14
Add a section on common log files

Crell:
There's probably more files that we should call out.  This file may also be a good place to put some extra search keywords.  Suggestions for those are welcome.  (We'll see after it builds what it's already picking up from the text.)


src/development/logs.md
```diff
@@ -0,0 +1,31 @@
+# Accessing logs
+
+Logs for various tasks on an application container are available in the `/var/log` directory.  They can be accessed on the normal shell after loging in with `platform ssh`.  Alternatively, they may also be accessed remotely using the `platform log` command.  The CLI lets you specify which log file to access (the name of the file below minus the `.log` extension), as well as view the while file in a pager, only the most recent lines, and so forth.  Run `platform log --help` for complete documentation.
```
ralt: "as well as view the while file" sounds pretty wrong :-)

src/development/logs.md
```diff
@@ -0,0 +1,31 @@
...
+
+nginx startup log messages will be recorded in this file.  It is rarely needed except when debugging possible nginx configuration errors.
+
+## `error.log`
+
+nginx-level errors that occur once nginx has fully started will be recorded here. This will include HTTP 500 errors for missing directories, file types that are excluded based on the `.platform.app.yaml` file, etc.
+
+## `deploy.log`
+
+The deploy log contains the output of the most recent run of the deploy hook for the container.  If there is no deploy hook then this file will be absent.
```
ralt: Not necessarily most recent?

Crell: What would it be then?  The last 5?  All ever? All since some date?

Crell:
I don't think I've seen a php.access.log yet on my projects. What would that hold that's different than access.log or php.log?

src/development/logs.md
Crell: Typo, will fix. :smile:

xtfer:
php.access.log seems to contain all requests to the php service, including those which are 404s etc.

I'll add it.

xtfer:
@Crell I think this is good. I've checked the php.access.log and added it, and made some other minor changes (ordered logs according to CLI order and fixed the menu item). I think this is ready to merge.

Crell:
Merged.  Thanks all!

PR #416, opened 2016-11-21
document postgres extensions

OriPekelman:


Crell:
Text looks good; I also filed a Jira ticket to automate this list in the future.

Should we also better format the list, either bolding the extension names or making them a table?

src/configuration/services/postgresql.md
```diff
@@ -78,6 +78,60 @@ The format exposed in the ``$PLATFORM_RELATIONSHIPS`` [environment variable](/de
 }
 ```
 
+## PostgreSQL Extensions
+We have full support for running PostgreSQL Extensions on Platform.sh. In your `services.yaml` 
+file you can simply add a configuration subkey with the following strucure:
```
mrkschan: structure instead of strucure?

OriPekelman: I am so sorry for having pushed a 't' to master; Couldn't bring myself to do a pull request for a single letter.

PR #417, opened 2016-11-22
disable sharing plugin instead of setting all to false

dudenhofer:
…  which causes a js error


Crell:
Hey cool that works!  Good thought.

PR #418, opened 2016-11-25
Document how to skip waiting for a build in Git push

mrkschan:


Crell:
That seems odd to repeat the env var multiple times. Also, the text sort of implies that it skips the building and deploy process entirely, when actually it's just allowing git push to complete quickly without blocking.  That should be clarified.

Crell:
I'd also prefer to see this information presented somewhere other than the FAQ.  In general, an FAQ is a sign that your documentation is buggy because people still have questions!  It's the "other/uncategorized"  option, which is generally bad IA.

pjcdawkins:
`GIT_SSH_COMMAND` is only available from Git version 2.3+, which still isn't the default on some platforms.

IMO it's easier to add in `$HOME/.ssh/config`:
```
Host *.platform.sh
  SendEnv PLATFORMSH_PUSH_*
```

then:
```sh
PLATFORMSH_PUSH_NO_WAIT=1 git push
```

pjcdawkins:
Actually this provides at least a little bit of a reason to implement a `platform push` command...

mrkschan:
So, let's have a "other/uncategorized" page ^^? Or, just leave this on FAQ?

Crell:
@mrkschan No, I mean an "Other/Uncategorized/FAQ" page is itself a code smell.  FAQ may work for now, but if we can find a real home for it instead we should do that.  I want to deprecate the FAQ eventually.

Crell:
We should include the information from @pjcdawkins above (that it's only available from Git 2.3 on, and there's a better way to set it as the default).

mrkschan:
Updated.

src/development/faq.md
```diff
@@ -14,6 +14,27 @@ If you want to delete your project and cancel your subscription, simply go to yo
...
 
+## Can I skip waiting for a build in Git push?
+
+Yes, with Git version 2.3+, use the following one-liner and your Git push will skip waiting for the build and deploy processes to finish.
+
+```
+GIT_SSH_COMMAND='ssh -o "SendEnv PLATFORMSH_PUSH_NO_WAIT"' PLATFORMSH_PUSH_NO_WAIT=1 git push
+```
+
+If you have to use old Git client, first modify your SSH client configuration by adding the followings to the `$HOME/.ssh/config` file:
```
pjcdawkins: an older Git client
adding the following

src/development/faq.md
```diff
@@ -14,6 +14,27 @@ If you want to delete your project and cancel your subscription, simply go to yo
...
+
+```
+Host *.platform.sh
+  SendEnv PLATFORMSH_PUSH_*
+```
+
+Then, perform Git push like below:
```
pjcdawkins: perform the Git push like this

(or "as follows")

mrkschan:
revised.

PR #424, opened 2016-12-03
Grammar corrections to users.md

uberhacker:


src/administration/users.md
```diff
@@ -137,14 +137,14 @@ Available commands:
 * user:role
   * View or change a user's role
 
-For example the following command would add the admin to alice@example.com role to the current
-project.
+For example the following command would add the admin to alice@example.com role
```
pjcdawkins: Should be "add the admin role to ..."

pjcdawkins:
Thanks! All looks good to me apart from the vestigial problem above

uberhacker:
@pjcdawkins: Good catch.  Added commit to address this issue.

PR #428, opened 2016-12-06
Grammar corrections to integrations.md

uberhacker:


src/administration/integrations.md
```diff
@@ -12,5 +12,6 @@ platform integrations
 ![Cli Integrations](/images/cli-integrations.png)
 
 > **note**
-> If you have created your account using the bitbucket or github oAuth Login in order to use the Platform CLI you will need to setup a
-> password which you can do by visiting this page [https://accounts.platform.sh/user/password](https://accounts.platform.sh/user/password)
+> If you have created your account using the Bitbucket or GitHub oAuth Login,
+in order to use the Platform CLI, you will need to set up a password by visiting
```
Crell: Let's put a "then" after the comma here.

Also, we favor soft-wrap for the documentation.  No need to hard wrap it.  (It makes it easier to edit when changing something so you don't need to manually reflow all of the text.)

src/administration/integrations.md
```diff
@@ -12,5 +12,4 @@ platform integrations
 ![Cli Integrations](/images/cli-integrations.png)
 
 > **note**
-> If you have created your account using the bitbucket or github oAuth Login in order to use the Platform CLI you will need to setup a
-> password which you can do by visiting this page [https://accounts.platform.sh/user/password](https://accounts.platform.sh/user/password)
+> If you have created your account using the Bitbucket or GitHub oAuth Login, then in order to use the Platform CLI, you will need to set up a password by visiting [https://accounts.platform.sh/user/password](https://accounts.platform.sh/user/password).
```
Crell: The comma after "CLI" is unnecessary.  The one after Login is technically optional in English, probably, but the latter is definitely unnecessary.  (It would be appropriate if there was also a comma after then, but that would be excessive.)

uberhacker: One of the things I hate most about the English language is how subjective it is in regards to "optional" grammar.  The way I was taught how to know whether to include a comma in a sentence or not was if it separates a topic or complete thought.  If the sentence still makes sense when you take out `then in order to use the Platform CLI`, then the comma is necessary.  In this case, the sentence still makes sense.  I can make the change you suggest but my English teacher would not approve. :)

PR #429, opened 2016-12-06
Add description of app.log

mrkschan:


Crell:
The rest of the files are listed alphabetically, so let's put app.log in alphabetically as well.

I believe I heard Simon say that the php.log will go away in favor of app.log soon-ish?  True?

Crell:
Actually, I'm going to close this in favor of #430 as those updates are just about ready to go anyway.  Thanks, @mrkschan.

PR #433, opened 2016-12-08
Clean up services and link to external docs

Crell:
The main change here is that all services pages now link to the service's documentation.  I also did some other cleanup while I was at it, and added a section on how to connect to services via ssh tunnel.

I'd like to also include an example or three for the latter, but I'm not sure how to do so as a clean command given that the name/port for services is not guaranteed and cannot be extracted in a single command.  Suggestions welcome for that.

### DIFF
--- a/src/configuration/services.md
+++ b/src/configuration/services.md
@@ -69,3 +69,36 @@ For example, the current default storage amount per project is 5GB (meaning 5120
 ## Using the services
 
 In order for a service to be available to an application in your project (Platform.sh supports not only multiple backends but also multiple applications in each project) you will need to refer to it in the [.platform.app.yaml](/configuration/app-containers.md) file which configures the *relationships* between applications and services.
+
+## Connecting to a service
+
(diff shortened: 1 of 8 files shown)


src/configuration/services.md
```diff
@@ -69,3 +69,36 @@ For example, the current default storage amount per project is 5GB (meaning 5120
...
+In this example, we can now securely connect to the database or redis server on our environment by connecting to the specified local ports.  The `platform tunnels` command will list all open tunnels:
+
+```text
++-------+---------------+-------------+-----------+--------------+
+| Port  | Project       | Environment | App       | Relationship |
++-------+---------------+-------------+-----------+--------------+
+| 30000 | a43m75zns6k4c | master      | [default] | redis        |
+| 30001 | a43m75zns6k4c | master      | [default] | database     |
++-------+---------------+-------------+-----------+--------------+
+```
```
mrkschan: OT to this MR.

I just find this is funny ... we have the SSH tunnel and user has to find the connection info from `$PLATFORM_RELATIONSHIPS` with another SSH session.

Is there CLI tool to prevent the 2nd SSH sessions just to obtain the database username / password? @pjcdawkins

mrkschan:
:shipit:

PR #439, opened 2016-12-15
Tutorials

Crell:
Add a section for tutorials.  The first tutorial is on exporting data.  More will be added in time.

Question: Should we mention exporting via ssh tunnel on the main exporting tutorial page?

Question: Should Tutorials be listed before or after Featured Frameworks?  I figure Framework-specific tutorials should go with the Framework; this section is for framework agnostic tutorials.

### DIFF
--- a/src/configuration/services/mongodb.md
+++ b/src/configuration/services/mongodb.md
@@ -71,7 +71,7 @@ foreach ($relationships['database'] as $endpoint) {
 
 ## Exporting data
 
-The most straightforward way to export data from a MongoDB database is to open an SSH tunnel to it and simply export the data directly using MongoDB's tools.  First, open an ssh tunnel with the Platform CLI:
+The most straightforward way to export data from a MongoDB database is to open an SSH tunnel to it and simply export the data directly using MongoDB's tools.  First, open an SSH tunnel with the Platform.sh CLI:
 
 ```bash
 platform tunnel:open
--- a/src/configuration/services/mongodb.md
+++ b/src/configuration/services/mongodb.md
@@ -68,3 +68,24 @@ foreach ($relationships['database'] as $endpoint) {
   $container->setParameter('database_password', $endpoint['password']);
 }
 ```
+
+## Exporting data
+
+The most straightforward way to export data from a MongoDB database is to open an SSH tunnel to it and simply export the data directly using MongoDB's tools.  First, open an ssh tunnel with the Platform CLI:
+
+```bash
+platform tunnel:open
+```
+
+That will open an SSH tunnel to all services on your current environment, and produce output something like the following:
+
+```text
+SSH tunnel opened on port 30000 to relationship: mongodb
+SSH tunnel opened on port 30001 to relationship: redis
+```
+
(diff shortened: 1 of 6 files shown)


src/configuration/services/mongodb.md
```diff
@@ -68,3 +68,24 @@ foreach ($relationships['database'] as $endpoint) {
   $container->setParameter('database_password', $endpoint['password']);
 }
 ```
+
+## Exporting data
+
+The most straightforward way to export data from a MongoDB database is to open an SSH tunnel to it and simply export the data directly using MongoDB's tools.  First, open an ssh tunnel with the Platform CLI:
```
pjcdawkins: ssh tunnel -> SSH tunnel
Platform CLI -> Platform.sh CLI

src/configuration/services/mysql.md
```diff
@@ -89,3 +89,23 @@ mysql -h database.internal -u user main
 ```
 
 Outside the application container, you can use Platform CLI `platform sql`.
+
+## Exporting data
+
+The easiest way to download all data in a MariaDB instance is with the Platform CLI.  If you have a single SQL database, the following command will export all data using the `mysqldump` command to a local file:
```
pjcdawkins: Platform CLI -> Platform.sh CLI

src/tutorials/exporting.md
```diff
@@ -0,0 +1,34 @@
...
+
+The only files to download are from any writeable file mounts you may have defined in your `.platform.app.yaml` file.  The easiest way to download those is using the `rsync` tool.  For instance, suppose you have a mounts section that defines one web-accessible directory and one non-web-accessible directory:
+
+```yaml
+mounts:
+    '/web/uploads': 'shared:files/uploads'
+    '/private': 'shared:files/private'
+```
+
+To use `rsync` to download each directory, we can use the following commands.  The `platform ssh --pipe` command will return the SSH URL for the current environment as an inline string that `rsync` can recognize. To use a non-default environment, use the `-e` switch after `--pipe`.  Note that the trailing slash on the remote path means `rsync` will copy just the files inside the specified directory, not the directory itself, .
```
pjcdawkins: Paragraph ends with `, .`

src/tutorials/exporting.md
```diff
@@ -0,0 +1,34 @@
...
+```bash
+rsync -az `platform ssh --pipe`:/app/private/ ./private/
+rsync -az `platform ssh --pipe`:/app/web/uploads ./uploads/
+```
+
+See the [`rsync` documentation](http://linuxcommand.org/man_pages/rsync1.html) for more details on how to adjust the download process.
+
+## Download data from services
+
+The mechanism for downloading from each service (such as your database) varies.  For services designed to hold non-persistent information (such as Redis or Solr) it's generally not necessary to download data as it can be rebuilt off of the primary data store.
```
pjcdawkins: off of -> from

pjcdawkins:
At some point, the text on the [Introduction](/platformsh/platformsh-docs/blob/master/src/README.md) page should be amended to mention the Tutorials section. But I don't think that should block this PR.

PR #440, opened 2016-12-15
Rewrite the variables page

Crell:
The variables page needs to cover project variables. It also needs to be clearer about what's what and how it gets used.  I've therefore rewritten it.

It's probably best to review just the pr environment, as the pr text itself is all over the place with additions and removals and moves.


src/development/variables.md
```diff
@@ -0,0 +1,213 @@
...
+
+As the above logic is defined in a file in your Git repository you are free to change it if desired.  The same behavior can also be easily implemented for any other application or framework.
+
+## Shell variables
+
+You can also provide a `.environment` file as part of your application, in your application root (as a sibling of your `.platform.app.yaml` file, or files in the case of a mult-app configuration).  That file will be sourced as a bash script when the container starts and on all SSH logins.  It can be used to set any environment variables directly, such as the PATH variable.  For example, the following `.environment` file will allow any executable installed using Composer as part of a project to be run regardless of the current directory:
```
pjcdawkins: mult-app -> multi-app

src/development/variables.md
```diff
@@ -0,0 +1,213 @@
...
+
+```bash
+platform project:variable:set foo bar --no-visible-build --no-visible-runtime
+```
+
+Naturally in practice you'll want to use only one or the other, or allow the variable to be visible in both cases.
+
+### Platform.sh Environment variables
+
+Environtment-level variables can also be set [through the web interface](administration/web/configure-environment.html#settings), or using the CLI. Environment variables are bound to a specific environment or branch.  An environment will also inherit variables from its parent environment, unless it has a variable defined with the same name.  That allows you to define your development variables only once, and use them on all the child environments.  For instance, to set the environment variable "foo" to the value "bar" on the currently checked out environment/branch, run:
```
pjcdawkins: Environtment-level -> Environment-level

src/development/variables.md
```diff
@@ -0,0 +1,213 @@
...
+
+```ini
+$ platform variable:set env:foo bar
+```
+
+With PHP, you can then access that variable with `getenv('FOO')`.
+
+### PHP-specific variables
+
+Any variable that is prepended with `php:` will also be added to the `php.ini` configuration of all PHP-based application containers.  For example, an environment variable named `php:display_errors` with value `On` is equivalent to placing the following in `php.ini`:
```
pjcdawkins: is prepended with -> begins with
(I'd prefer)

pjcdawkins:
LGTM apart from the very minor errors above 👍

PR #442, opened 2016-12-17
Grammar changes to configure-environment.md

uberhacker:


Crell:
We actually have another PR up (#440) that completely rewrites this page.  Grammar tweaks to the current text when it's about to be replaced entirely doesn't seem worthwhile.  If you want to review that PR for grammar, though, that would be greatly appreciated. :smile:

uberhacker:
Hey @Crell: Doesn't that PR reference variables.md?  My PR looks totally isolated from your mentioned PR.  I can certainly look over variables.md after it gets merged into master, however.

Crell:
It replaces environtment-variables.md with variables.md entirely; content and name, since it also is expanding the docs to include the new project-level variables we just

uberhacker:
Sorry, I guess I'm not understanding.  My PR references configure-environment.md.  I have not seen a PR which renames this doc or am I missing something?

Crell:
PR #440, if you look at the diff on it, deletes environment-variables.md entirely, and adds a new file variables.md.  (That's effectively "rename" as far as a git diff is concerned.)  It also updates the SUMMARY.md file to use the new variables.md instead.  So the file this PR touches is going to cease to exist in the very near future.

uberhacker:
I get that but my PR references configure-environment.md, not environment-variables.md.

Crell:
... Argh! You are completely correct. I misread the title.  Sorry about that. I'll review this PR now. :smile: 

Crell--

src/administration/web/configure-environment.md
```diff
@@ -15,7 +15,7 @@ environment.
 The *Delete* action will destroy all services running on this environment (PHP, MySQL, volumes...)
 so that only the Git branch remains. You can reactivate the environment later if needed.
 
-To also delete the branch, you need to use Git and run:
+To delete the branch also, you need to execute the following:
```
Crell: With also in that position, "as well" works better.  Or leave the also where it is.  Either one is legit I think.

src/administration/web/configure-environment.md
```diff
@@ -12,10 +12,10 @@ environment.
 
 ![Configure Platform.sh environment settings](/images/ui-conf-environment-settings.png)
 
-The *Delete* action will destroy all services running on this environment (PHP, MySQL, volumes...)
+The *Delete* action will destroy all services running on this environment (PHP, MySQL, volumes, ...)
```
Crell: I'm not sure if the comma is correct there.  However, it's probably better to instead use "volumes, etc." rather than the ellipsis anyway.

uberhacker:
Fair enough, change submitted.

Crell:
Thanks!  (And sorry again for the file confusion.)

PR #443, opened 2016-12-20
Provide more details on authentication for private composer

Crell:
Based on feedback from the EZ Platform team.

### DIFF
--- a/src/tutorials/composer-auth.md
+++ b/src/tutorials/composer-auth.md
@@ -46,10 +46,17 @@ Then in your build hook set the authentication information for the remote server
 hooks:
     build: |
         set -e
-        composer config my-private-repos.example.com $COMPOSER_KEY $COMPOSER_PASSWORD
+        composer config http-basic.my-private-repos.example.com $COMPOSER_KEY $COMPOSER_PASSWORD
         composer install --no-dev --prefer-dist --no-progress --no-interaction --optimize-autoloader
   ```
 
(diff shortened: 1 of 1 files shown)


src/tutorials/composer-auth.md
```diff
@@ -46,10 +46,17 @@ Then in your build hook set the authentication information for the remote server
...
 
 From here on, everything should proceed as normal.  Composer will download all listed packages from whatever repositories are appropriate, authenticating against them as needed, build the application, and then it will deploy as normal.  Because the variables are defined above to not be visible at runtime they will not appear at all in the running application.
+
+
+## Private repository hosting
+
+Typically, a private dependency will be hosted in a private Git repository.  While Platform.sh supports [synchronizing the primary repository](/development/private-repository.md) with a private third party, that does not apply for third party dependencies that may have their own authenticated repository.  For that reason, pulling a dependency directly from a third party Git repository that requires authentication is not supported.
```
pjcdawkins: > pulling a dependency directly from a third party Git repository that requires authentication is not supported

Huh? It is, if you can persuade the third party to add your project's deploy key.

Maybe it's the definition of "third party" that's not clear?

Crell: We ran into this with EZ Platform. Their private repos are on GitHub, but they can't add every single one of their customers to the GitHub repo via ssh key.  But their Satis caches the tagged release snapshots which dodges that issue.  This was my attempt at explaining that caveat.

PR #448, opened 2016-12-22
Grammar changes to app-containers.md

uberhacker:


src/configuration/app-containers.md
```diff
@@ -9,19 +9,19 @@ You control your application and the way it will be built and deployed on
 Platform.sh via a single configuration file, `.platform.app.yaml`, located at
 the root of your application folder inside your Git repository.
 
-Here's an example of a minimalist `.platform.app.yaml` file for PHP:
+An example of a minimalist `.platform.app.yaml` file for PHP is below:
```
Crell: Both old and new versions seem grammatically fine to me. What's the issue here?

src/configuration/app-containers.md
```diff
@@ -9,19 +9,19 @@ You control your application and the way it will be built and deployed on
...
 
 ```yaml
 # .platform.app.yaml
 
-# The name of this app. Must be unique within a project.
+# The name of this application which must be unique within a project.
```
Crell: Comma between `application` and `which`.

src/configuration/app-containers.md
```diff
@@ -9,19 +9,19 @@ You control your application and the way it will be built and deployed on
...
 name: 'app'
 
 # The type key specifies the language and version for your application.
 type: 'php:7.0'
 
-# On PHP, there are multiple build flavors available. Pretty much everyone except
-# Drupal 7 users will want the composer flavor.
+# On PHP, there are multiple build flavors available. Pretty much everyone
+# except Drupal 7 users will want the composer flavor.
```
Crell: Why the word wrap change here?

src/configuration/app-containers.md
```diff
@@ -51,59 +52,59 @@ web:
...
 ```
 
 > **Note**
 > This configuration file is specific to one application. If you have multiple
-> applications inside your Git repository (i.e. a RESTful
-> web service and a front-end, or a main web site and a blog), you need
-> one `.platform.app.yaml` at the root of each application. See the [Multi-app](/configuration/app/multi-app.md) documentation.
+> applications inside your Git repository (i.e. a RESTful web service and a
+> front-end or a main web site and a blog), you need one `.platform.app.yaml`
+> at the root of each application. See the [Multi-app](/configuration/app/multi-app.md) documentation.
```
Crell: I'm unclear what's changing here.  However, I'd argue the comma within the parenthetical is needed as otherwise we're left with a string of clauses linked by ands and ors, and left with order of operations to determine what goes where.  English's order of operations in such cases is highly inconsistent. :smile:

src/configuration/app-containers.md
```diff
@@ -51,59 +52,59 @@ web:
...
 
 The `name` is the unique identifier of the application. Platform.sh
 supports multiple applications within a project, so each application
 must have a **unique name** within a project. The name may only be
-composed of lower case alpha-numeric characters. (a-z0-9).  *Be advised
-that changing the `name` of your app after it's been deployed will
-destroy all storage volumes, and thus is typically a Very Bad Thing to do.*
+composed of lower case alpha-numeric characters (a-z0-9).  *Be advised
+that changing the `name` of your application after it has been deployed will
+destroy all storage volumes and thus, is typically a Very Bad Thing to do.*
```
Crell: The comma is in the correct place to begin with, I believe. The rest of the changes in this section are fine.

src/configuration/app-containers.md
```diff
@@ -51,59 +52,59 @@ web:
...
 It could be useful under certain circumstances in the early stages
-of development but you almost certainly don't want to change it on 
+of development but you almost certainly don't want to change it on
 a live project.
 
 This name is used in the `.platform/routes.yaml` file to define the HTTP upstream
-(by default `php:http` - if you called your application `app` you will
+(by default `php:http`; if you called your application `app` you will
 need to use `app:http` in the upstream field).
```
Crell: I love a good use of semicolons. :smile:  However, this line is no longer accurate.  There is no default application name, and all of our examples now use `app`, not `php`.  I'd say the parenthetical should be removed entirely, perhaps replaced with a "for instance, if your application is named...".

src/configuration/app-containers.md
```diff
@@ -51,59 +52,59 @@ web:
...
 
 You can also use this name in multi-application relationships.
 
 > **Note**
-> Changing the name of an app is the same as deleting it and replacing
-> it: your app data (static files) will be deleted.
+> Changing the name of an application is the same as deleting it and replacing
+> it. Your application data (static files) will be deleted.
```
Crell: application's, with a possessive.

src/configuration/app-containers.md
```diff
@@ -131,7 +132,7 @@ Its only property is `flavor`, which specifies a default set of build tasks to r
 * Node.js (`default` by default)
   * `default` will run `npm prune --userconfig .npmrc && npm install --userconfig .npmrc` if a `package.json` file is detected. Note that this also allows you to provide a custom `.npmrc` file in the root of your application (as a sibling of the `.platform.app.yaml` file.)
 
-In all languages you can also specify a flavor of `none` (which is the default for any language other than PHP and Node.js), which, as the name suggests, will take no action at all. That is useful when you want complete control over your build steps, such as to run a custom composer command or use an alternate Node.js package manager.
+In all languages, you can also specify a flavor of `none` (which is the default for any language other than PHP and Node.js); as the name suggests, will take no action at all. That is useful when you want complete control over your build steps, such as to run a custom composer command or use an alternate Node.js package manager.
```
Crell: The comma seems extraneous here.  The rest is fine.

src/configuration/app-containers.md
```diff
@@ -142,31 +143,31 @@ build:
 
 ### Access
 
-The `access` define the user roles who can log in via SSH to the
-environments they have access to.
+The `access` defines the user roles who can log in via SSH to the
+environments they have permission to access.
```
Crell: Let's remove the hard word wrap while we're at it.

src/configuration/app-containers.md
```diff
@@ -142,31 +143,31 @@ build:
...
 application.
 
 The left-hand side is the name of the relationship as it will be exposed
 to the application in the *PLATFORM_RELATIONSHIPS* environment
 variable. The right-hand side is in the form
-`<service name>:<endpoint name>`. where "service name" comes from the
+`<service name>:<endpoint name>` where "service name" comes from the
  `.platform/services.yaml` and  "endpoint name" should be the same as the
- value of "type"  declared in that same file.
+ value of "type" declared in that same file.
```
Crell: We're about to change this section as we introduce multi-endpoint services, but this is fine for now. :smile: (It's accurate right now only by convention and coincidence.)

src/configuration/app-containers.md
```diff
@@ -142,31 +143,31 @@ build:
...
 
 > **Note**
-> In the first  example above you could very well have something
+> In the first example above, you could very well have something
```
Crell: Comma unnecessary.

src/configuration/app-containers.md
```diff
@@ -211,37 +212,37 @@ web:
 #### Upstream
 
 `upstream` is an optional key that describes how your application listens to
-requests, and what protocol it speaks.
+requests and what protocol it speaks.
```
Crell: Remove hard line break while we're here.

src/configuration/app-containers.md
```diff
@@ -211,37 +212,37 @@ web:
...
 
 The following subkeys can be defined:
 * `socket_family`:
     Default: `tcp`. Describes whether your application will listen on a
-    Unix socket or a TCP socket. Can be either `tcp` or `unix`.
+    Unix socket or a TCP socket and can be either `tcp` or `unix`.
```
Crell: Hm. I agree we shouldn't have a sentence fragment here, but the alternative feels off to me, too.  I'm open to suggestions here.

src/configuration/app-containers.md
```diff
@@ -211,37 +212,37 @@ web:
...
 * `protocol`:
     Specifies whether your application is going to receive incoming requests
-    over HTTP, FastCGI, or UWSGI. Can be `http` or `fastcgi`, for now. The
+    over HTTP, FastCGI, or UWSGI and can be `http` or `fastcgi`, for now. The
```
Crell: Ibid.

src/configuration/app-containers.md
```diff
@@ -254,35 +255,35 @@ web:
             ...
 ```
 
-It has a few subkeys, which are:
+It has a few subkeys listed below:
```
Crell: Hm. I think this technically would be read as "there are subkeys listed below, some of which are used by web.locations", where the actual intent is "web.locations has subkeys. An exhaustive list of them is below."

src/configuration/app-containers.md
```diff
@@ -254,35 +255,35 @@ web:
...
 
 * `root`:
-    The folder from which to serve static assets for this location,
-    relative to the application root. The application root is the directory in
-    which the `.platform.app.yaml` file is located.  Typical values for this
+    The folder from which to serve static assets for this location
+    relative to the application root. The application root is the directory
+    where the `.platform.app.yaml` file is located.  Typical values for this
```
Crell: Subjective: I think "in which" works better in context here.  The comma could go either way.

src/configuration/app-containers.md
```diff
@@ -254,35 +255,35 @@ web:
...
     property include `public` or `web`.  Setting it to `""` is not recommended,
     and its behavior may vary depending on the type of application.  Absolute
     paths are not supported.
 * `passthru`:
     Whether to forward disallowed and missing resources from this location to
-    the application. Can be true, false or an absolute URI path (with leading
-    `/`), the default value is false. For non-PHP applications it will
+    the application and can be true, false or an absolute URI path (with leading
+    `/`). The default value is false. For non-PHP applications, it will
```
Crell: Comma after applications is not needed here.

src/configuration/app-containers.md
```diff
@@ -254,35 +255,35 @@ web:
...
+    relative to the application root. The application root is the directory
+    where the `.platform.app.yaml` file is located.  Typical values for this
     property include `public` or `web`.  Setting it to `""` is not recommended,
     and its behavior may vary depending on the type of application.  Absolute
     paths are not supported.
 * `passthru`:
     Whether to forward disallowed and missing resources from this location to
-    the application. Can be true, false or an absolute URI path (with leading
-    `/`), the default value is false. For non-PHP applications it will
+    the application and can be true, false or an absolute URI path (with leading
```
Crell: While we're here, true and false should get backticks around them as they're literal code values.

src/configuration/app-containers.md
```diff
@@ -254,35 +255,35 @@ web:
...
+    `/`). The default value is false. For non-PHP applications, it will
     generally be just true or false.  In a PHP application, this will typically
     be the front controller such as `/index.php` or `/app.php`.  This entry
-    works similar to `mod_rewrite` under Apache.  Note: if the value of
-    `passthru` does not begin with the same value as the location key it is
-    under the passthru may evaluate to another entry. That may be useful when
+    works similar to `mod_rewrite` under Apache.  Note: If the value of
+    `passthru` does not begin with the same value as the location key, it is
+    under the passthru and may evaluate to another entry. That may be useful when
```
Crell: The comma is in the wrong place here, and completely changes the meaning of the sentence. :smile: It should actually be after "under".  That is, in the situation where the passthru doesn't begin with the same value as the location under which it is located, then in that situation it may evaluate to another entry.

src/configuration/app-containers.md
```diff
@@ -254,35 +255,35 @@ web:
...
     you want different cache settings for different paths, for instance, but
     want missing files in all of them to map back to the same front controller.
     See the example block below.
 * `index`:
-    The file or files to consider when serving a request for a directory. Can
-    be file name, an array of file names, or *null*. Typically `index.html`.
+    The file or files to consider when serving a request for a directory and can
+    be a file name, an array of file names, or *null* (typically `index.html`).
```
Crell: Adding a parenthetical for the typical case is fine, but it IMO it should be a self-contained sentence with capital T then. Viz, same as the original text just with parentheses.

src/configuration/app-containers.md
```diff
@@ -362,10 +363,10 @@ languages:
 * Java (with integrated Maven and Ant support)
 
 Those dependencies are independent of the eventual dependencies of your
-application, and are available in the `PATH`, during the build process
+application and are available in the `PATH` during the build process
 and in the runtime environment of your application.
 
-You can specify those dependencies like this:
+You can specify those dependencies like below:
```
Crell: "like below" sounds awkward.  "like so"?

src/configuration/app-containers.md
```diff
@@ -386,21 +387,21 @@ The `hooks` let you define shell commands to run during the build and deployment
...
-Possible hooks are:
+Possible hooks are below:
 
 -   **build**: We run a build hook before your application has been
     packaged. No other services are accessible at this time since the
     application has not been deployed yet.
 -   **deploy**: We run a deploy hook after your application has been
     deployed and started. You can access other services at this stage
-    (MySQL, Solr, Redis...). However, the disk where the application lives is read-only at this point.
+    (MySQL, Solr, Redis, etc.), however, the disk where the application lives is read-only at this point.
```
Crell: Why join the sentences?  It seems fine as its own sentence.  (The "etc" is a good change, though.)

src/configuration/app-containers.md
```diff
@@ -467,17 +468,17 @@ hooks:
...
 
-It has a few subkeys which are:
+It has a few subkeys listed below:
 
 -   **spec**: The [cron specification](https://en.wikipedia.org/wiki/Cron#CRON_expression). For example: `*/20 * * * *` to run every 20 minutes.
 -   **cmd**: The command that is executed, for example
     `cd public ; drush core-cron`
 
-The minimum interval between cron runs is 5 minutes, even if specified as less.
+The minimum interval between cron runs is 5 minutes; even if specified as less.
```
Crell: Much as I love semi-colons, I don't think it's accurate in this case.

src/configuration/app-containers.md
```diff
@@ -467,17 +468,17 @@ hooks:
...
 
-Note that cron runs are executed using the dash shell, not the bash shell used by normal SSH logins. In most cases that makes no difference but may impact some more involved cron scripts.
+Note that cron runs are executed using the dash shell; not the bash shell used by normal SSH logins. In most cases, that makes no difference but may impact some more involved cron scripts.
```
Crell: Same.  The comma is better in the first case as it's a sentence fragment otherwise, and unnecessary in the latter case.  This line can be stet.

src/configuration/app-containers.md
```diff
@@ -509,15 +510,15 @@ web:
 
 ### Runtime
 
-The `.platform.app.yaml` file also supports a `runtime` key, which allows selected customizations to the language runtime. As those possibilities vary by language, please see the appropriate language documentation.
+The `.platform.app.yaml` file also supports a `runtime` key which allows selected customizations to the language runtime. As those possibilities vary by language, please see the appropriate language documentation.
```
Crell: Either "key, which" or "key that".  Either one is (I think) fine.

Crell:
Thanks, Ed, that's a big page to tackle. :smile: Merry Christmas!

src/configuration/app-containers.md
uberhacker: `Here's` doesn't clearly indicate "where" the reader should look.  Below is more understandable.

src/configuration/app-containers.md
uberhacker: Agree since there is a pause when you read the sentence.

src/configuration/app-containers.md
uberhacker: Since this line is a comment and most linters will balk when you go past 80 characters.

src/configuration/app-containers.md
uberhacker: The sentence was changed to improve clarity.

src/configuration/app-containers.md
uberhacker: I was thinking about where you would pause when reading the sentence.  It's totally subjective, however.

src/configuration/app-containers.md
uberhacker: I agree the paranthesis should be removed and replaced by a new sentence instead.

src/configuration/app-containers.md
uberhacker: I generally don't like using possessives if not necessary but I can add it.

src/configuration/app-containers.md
uberhacker: Just thinking about the pause when reading the sentence.

src/configuration/app-containers.md
uberhacker: Can do.

src/configuration/app-containers.md
uberhacker: Again, pause while reading the sentence.

src/configuration/app-containers.md
uberhacker: Can do.

src/configuration/app-containers.md
uberhacker: How about this?  ... over HTTP (`http`), FastCGI (`fastcgi`) or UWSGI (not available currently).

src/configuration/app-containers.md
uberhacker: `where` means a location (or place).  `in which` references a thing (or noun).

src/configuration/app-containers.md
uberhacker: Again, it's all based on the pause while reading the sentence.

src/configuration/app-containers.md
uberhacker: Can do.

src/configuration/app-containers.md
uberhacker: Agreed.  Now that I read the sentence again, it makes more sense since you explained it.

src/configuration/app-containers.md
uberhacker: How about remove the parenthesis and put a comma after *null* instead?

src/configuration/app-containers.md
uberhacker: The `below` keyword clearly indicates to the reader where they should look to find the dependencies.

src/configuration/app-containers.md
uberhacker: I was thinking the `however` was acting more like a connecting word but it could work as a separate sentence.

src/configuration/app-containers.md
uberhacker: If you substitute the word `and` for the semicolon, it does make sense but the comma works also.

src/configuration/app-containers.md
uberhacker: It's all about the pauses.

src/configuration/app-containers.md
uberhacker: Fair enough.

Crell:
@uberhacker Are you going to be returning to this PR?  Looks like it needs a rebase as well.

uberhacker:
Hey @Crell:  Happy New Year!  Hope you enjoyed the holidays.  I was kind of expecting some feedback on some of my questions before I proceeded, but definitely, I can resume pending feedback. :)
Thanks

src/configuration/app-containers.md
Crell: Verbal pauses and grammatical commas do not always correlate one to one.  English is weird. :smile:

src/configuration/app-containers.md
Crell: As above, verbal pause and written comma do not always correlate, because English.

src/configuration/app-containers.md
Crell: If UWSGI isn't an option, I guess it shouldn't be mentioned at all?

src/configuration/app-containers.md
Crell: "and" can join both independent (full sentence on its own) and dependent (incomplete sentence) clauses.  Semicolon only joins full sentences.  Since "even if specified as less" is not a complete sentence on is own a comma is more appropriate here.

src/configuration/app-containers.md
Crell: Only in verbal English, not written. :smile: (English is hard; it took me a while to properly grok commas, too, even as a native speaker.)

Crell:
Ah, I only skimmed before and didn't see any asking questions. I left a few more comments on the comments (yo dawg).  Looking forward to the update/rebase/merge/thing.

uberhacker:
OK, ready for another review.

src/configuration/app-containers.md
```diff
@@ -51,59 +52,59 @@ web:
...
     deploy: |
       app/console --env=prod cache:clear
 ```
 
 > **Note**
 > This configuration file is specific to one application. If you have multiple
-> applications inside your Git repository (i.e. a RESTful
-> web service and a front-end, or a main web site and a blog), you need
-> one `.platform.app.yaml` at the root of each application. See the [Multi-app](/configuration/app/multi-app.md) documentation.
+> applications inside your Git repository (i.e. a RESTful web service and a
```
Crell: Minor nit: This should be e.g., not i.e..  eg is equivalent to "for example", i.e. to "that is" or "put another way".

src/configuration/app-containers.md
```diff
@@ -131,7 +132,7 @@ Its only property is `flavor`, which specifies a default set of build tasks to r
 * Node.js (`default` by default)
   * `default` will run `npm prune --userconfig .npmrc && npm install --userconfig .npmrc` if a `package.json` file is detected. Note that this also allows you to provide a custom `.npmrc` file in the root of your application (as a sibling of the `.platform.app.yaml` file.)
 
-In all languages you can also specify a flavor of `none` (which is the default for any language other than PHP and Node.js), which, as the name suggests, will take no action at all. That is useful when you want complete control over your build steps, such as to run a custom composer command or use an alternate Node.js package manager.
+In all languages you can also specify a flavor of `none` (which is the default for any language other than PHP and Node.js); as the name suggests, will take no action at all. That is useful when you want complete control over your build steps, such as to run a custom composer command or use an alternate Node.js package manager.
```
Crell: With a semi-colon here what comes after must be a complete sentence. I suggest replacing the comma with "it".

src/configuration/app-containers.md
```diff
@@ -254,35 +249,35 @@ web:
             ...
 ```
 
-It has a few subkeys, which are:
+There are subkeys listed below, some of which are used by web.locations as follows:
```
Crell: This sentence feels far clumsier to me.  Honestly I'd leave this line as is.  I can't think of a less confusing way to word it.

Crell:
Yay!  Thanks, @uberhacker!

PR #449, opened 2016-12-23
Add generic migration guide

Crell:
This involves removing the Symfony-specific one as vestigial, and improving the documentation for individual services.  There's little that's specific to any one framework, so no reason to have more than one I'd argue.

If we find anything especially peculiar about one framework or another we can explore that case when we get to it.


src/tutorials/migrating.md
```diff
@@ -0,0 +1,54 @@
+---
+search:
+    keywords: ['migrating', 'upload']
+---
+
+# Migrating to Platform.sh
+
+Moving an already-built site to Platform.sh is generally straightforward.  For the most part, the only part that will vary from one framework to another is the details of the Platform.sh configuration files.  See the **Featured Framework** section or our [project templates](https://github.com/platformsh/) for more project-specific documentation.
```
pjcdawkins: Should be "Featured Frameworks"

pjcdawkins: Ideally "Platform.sh configuration files" should be linked to something, but I notice we probably don't have a single page that can be linked (there are 3, for app, routes, and services).

Crell:
Question. Should this page be called "Importing" to parallel the existing "Exporting"?  We can still make "migrate" a search keyword.

src/tutorials/migrating.md
Crell: Hm. Our options here are:

1) No link (stet)
2) Link just to the app-containers page
3) Link different words to different doc pages
4) Put links to all 3 pages in a parenthetical

I don't like option 3 as it has too little context for the user.  I could deal with option 2 but it ignores the other two configuration files.  I'm not a huge fan of option as it feels very forced.  And since the config section is now very easy to find I'm OK with option 1.

Other thoughts?

pjcdawkins:
I think Migrating makes sense, because it's not just about importing data, it's about configuration files, etc.

Crell:
Anything else here that needs to be updated or is it good to merge?

PR #452, opened 2017-01-04
Dropbox tutorial v0.1

JGrubb:


src/tutorials/dropbox.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Install the daemon
+
+Per the instructions on [installing the Linux client](https://www.dropbox.com/install-linux), you can install the client via a simple one-liner shell command, so place this into your build hook - `cd ~ && wget -O - "https://www.dropbox.com/download?plat=lnx.x86_64" | tar xzf -
+`, like so
+
+```yaml
+hooks:
+  build: |
+    echo "This is the build hook firing!"
```
Crell: We don't have lines like this is any other examples.  Let's remove.

src/tutorials/dropbox.md
```diff
@@ -0,0 +1,72 @@
...
+    cd ~ && wget -O - "https://www.dropbox.com/download?plat=lnx.x86_64" | tar
+```
+
+The build hook runs when your application's codebase is being assembled for deployment, so the filesystem is still writable.  This means that the above command will create a `~/.dropbox-dist` directory in the home directory of your project, aka "the project root", aka `/app`.  Starting the daemon is _almost_ as simple as running `~/.dropbox-dist/dropboxd` when you `ssh` into your application's container.
```
Crell: We don't mention that `/app` is the home dir anywhere else, since that's technically an implementation detail.  Let's elide that.

pjcdawkins: Nitpick: "home directory of your application" (not your project)

JGrubb: Excellent nitpick actually, thanks.

src/tutorials/dropbox.md
```diff
@@ -0,0 +1,72 @@
+# Setting up the Dropbox daemon
+
+- Time needed: ~10 minutes
+
+Dropbox has a very useful [Linux version](https://www.dropbox.com/install-linux) that can be installed into your Platform.sh application with a bit of intermediate [build hook](/configuration/app-containers.html#hooks) hacking.
+
```
Crell: The introduction should include when/why you'd want to do this. It's not something most people are likely to think of.

src/tutorials/dropbox.md
```diff
@@ -0,0 +1,72 @@
+# Setting up the Dropbox daemon
+
+- Time needed: ~10 minutes
```
Crell: Did you time yourself? :smile: None of the other tutorials have this.

src/tutorials/dropbox.md
```diff
@@ -0,0 +1,72 @@
...
+
+Dropbox has a very useful [Linux version](https://www.dropbox.com/install-linux) that can be installed into your Platform.sh application with a bit of intermediate [build hook](/configuration/app-containers.html#hooks) hacking.
+
+## Install the daemon
+
+Per the instructions on [installing the Linux client](https://www.dropbox.com/install-linux), you can install the client via a simple one-liner shell command, but I needed to modify it slightly so modify your build hook to add this to it -
```
Crell: Documentation shouldn't use "I". It should be addressed to the user ("you"), who is the one actually taking action.

src/tutorials/dropbox.md
```diff
@@ -0,0 +1,72 @@
...
+```yaml
+hooks:
+  build: |
+    wget "https://www.dropbox.com/download?plat=lnx.x86_64" -O archive.tar
+    tar xzvf archive.tar && rm archive.tar
+```
+
+
+
+The build hook runs when your application's codebase is being assembled for deployment, so the filesystem is still writable.  This means that the above command will create a `~/.dropbox-dist` directory in the home directory of your application, aka "the application root".  Starting the daemon is _almost_ as simple as running `~/.dropbox-dist/dropboxd` when you `ssh` into your application's container.
```
Crell: Too many newlines.

src/tutorials/dropbox.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Creating the necessary writable directories
+
+Before we can start the daemon, we'll need to add a couple directories that Dropbox expects to be there and to be writable.  The main one is `~/Dropbox`, which contains **your entire Dropbox**.  (There may be a way to only mount part of it, and please do open a Merge Request to these docs if you figure it out).  For obvious reasons, this directory needs to be at least as large as all the contents of your Dropbox or you're going to have a Bad Time.
```
Crell: As above, all verb tenses should be in the instructive: "You need to X, then do Y."

Crell: Let's skip the parenthetical.  Also, the "Bad Time" phrase is more appropriate for a blog than for the documentation section.

src/tutorials/dropbox.md
```diff
@@ -0,0 +1,72 @@
...
+
+The other is a metadata/etc directory at `~/.dropbox`.  This author's `.dropbox` directory is 43M, so it doesn't appear to take up too much room.
```
Crell: The author of Docs.Platform.sh is Platform.sh, Inc. :smile: Suggested: "This directory is usually very modestly sized so doesn't need to be factored into your needed-size calculations."

src/tutorials/dropbox.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Creating the necessary writable directories
+
+Before we can start the daemon, we'll need to add a couple directories that Dropbox expects to be there and to be writable.  The main one is `~/Dropbox`, which contains **your entire Dropbox**.  (There may be a way to only mount part of it, and please do open a Merge Request to these docs if you figure it out).  For obvious reasons, this directory needs to be at least as large as all the contents of your Dropbox or you're going to have a Bad Time.
+
+The other is a metadata/etc directory at `~/.dropbox`.  This author's `.dropbox` directory is 43M, so it doesn't appear to take up too much room.  
+
+The combined total of these two directories needs to be smaller than whatever you've specified in the [`disk` parameter](/configuration/app-containers.html#disk) in your `.platform.app.yaml`, or put another way - the `disk` parameter needs to be larger than your total Dropbox usage.  (Don't forget to add a nice buffer on top of that too, because if your writable disk space fills up completely then Bad Things can happen).
```
Crell: This can also be greatly condensed and formalized.  Suggested: 

"Because Dropbox files are mirrored on your writable disk mount, your application container's `disk` size needs to be large enough to accommodate all of your Dropbox files as well as any user-uploaded files and your own code.  You should also include sufficient extra space to allow for growth over time.

src/tutorials/dropbox.md
JGrubb: > Did you time yourself?

I did actually

> None of the other tutorials have this.

If it makes sense, maybe they should?

src/tutorials/dropbox.md
```diff
@@ -0,0 +1,69 @@
...
+```yaml
+hooks:
+  build: |
+    wget "https://www.dropbox.com/download?plat=lnx.x86_64" -O archive.tar
+    tar xzvf archive.tar && rm archive.tar
+  deploy: |
+    # starts the process as a daemon and sends all
+    # output to /dev/null.  Obviously, feel free to
+    # direct this output to a logfile(s) if you wish.
+    nohup ~/.dropbox-dist/dropboxd >/dev/null 2>&1 &
```
JGrubb: It _could_ be in `web.commands.start`, I don't understand why that would be a benefit to put it there though vs the deploy hook.

ralt: 1/ It semantically makes more sense
2/ When adding support for multiple app containers, `web.commands.start` will be run on every container, while the deploy hook will only be run on one.

JGrubb: Multiple app containers will be managed from one `.platform.app.yaml` file?

JGrubb: re: #1, no real argument there, but it does spread the scope of the tutorial out from being all in the hooks section.  You also then have to recreate the default start command for PHP, which again is not that big a deal but it kind of makes this whole tutorial a bit more confusing than really necessary IMO.

src/tutorials/dropbox.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Creating the necessary writable directories
+
+You'll need to add two writable/mounted directories that Dropbox expects to be there and to be writable.  The main one is `~/Dropbox`, which contains **your entire Dropbox**.  The other is a metadata directory at `~/.dropbox`.  This is typically modestly sized, on the order of less than 100MB.  
+
+Because Dropbox files are mirrored on your writable disk mount, your application container's disk size needs to be large enough to accommodate all of your Dropbox files as well as any user-uploaded files and your own code. You should also include sufficient extra space to allow for growth over time.
+
+```yaml
+# for a 1G total Dropbox size, 
+# leaving another 1G for the application
```
Crell: Nitpick: For should be capitalized, and a period after application.

src/tutorials/dropbox.md
```diff
@@ -0,0 +1,72 @@
...
+Before we can start the daemon, we'll need to add a couple directories that Dropbox expects to be there and to be writable.  The main one is `~/Dropbox`, which contains **your entire Dropbox**.  (There may be a way to only mount part of it, and please do open a Merge Request to these docs if you figure it out).  For obvious reasons, this directory needs to be at least as large as all the contents of your Dropbox or you're going to have a Bad Time.
+
+The other is a metadata/etc directory at `~/.dropbox`.  This author's `.dropbox` directory is 43M, so it doesn't appear to take up too much room.  
+
+The combined total of these two directories needs to be smaller than whatever you've specified in the [`disk` parameter](/configuration/app-containers.html#disk) in your `.platform.app.yaml`, or put another way - the `disk` parameter needs to be larger than your total Dropbox usage.  (Don't forget to add a nice buffer on top of that too, because if your writable disk space fills up completely then Bad Things can happen).
+
+```yaml
+# for a 1G total Dropbox size, 
+# leaving another 1G for the application
+disk: 2048
```
JGrubb: Should I just ditch the comment altogether then?

src/tutorials/dropbox.md
```diff
@@ -0,0 +1,72 @@
...
+  build: |
+    wget "https://www.dropbox.com/download?plat=lnx.x86_64" -O archive.tar
+    tar xzvf archive.tar && rm archive.tar
+```
+
+Commit this to your project and push it to Platform.sh.  This will download a fresh copy of the Dropbox daemon and run the startup command every time you deploy from now on.
+
+## Caveat
+
+The configuration that was stored in the `~/.dropbox` directory will be available in any child environments that are created out of whichever one you set all this up in.  They can also be sync'd to any child environments you've already created and will function as intended.
```
Crell: This caveat needs a follow-up paragraph that notes in big bold letters that the Dropbox share is therefore shared between environments, which is contrary to the way everything else we do works.  That's really the number one biggest drawback of using Dropbox or similar, and it needs to be called out.

src/tutorials/dropbox.md
Crell: Yeah, let's just skip it.  It's confusing.

PR #455, opened 2017-01-09
Document --gzip option for db:dump

pjcdawkins:
Also bumps it above the --relationship option (as I believe it'd be more commonly wanted)

### DIFF
--- a/src/configuration/services/mysql.md
+++ b/src/configuration/services/mysql.md
@@ -98,17 +98,17 @@ The easiest way to download all data in a MariaDB instance is with the Platform.
 platform db:dump
 ```
 
+By default the file will be uncompressed. If you want to compress it, use the `--gzip` (`-z`) option:
+
+```bash
+platform db:dump --gzip
+```
+
 If you have multiple SQL databases it will prompt you which one to export. You can also specify one by relationship name explicitly:
 
 ```bash
 platform db:dump --relationship database
 ```
-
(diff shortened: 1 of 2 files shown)


src/configuration/services/mysql.md
```diff
@@ -98,17 +98,17 @@ The easiest way to download all data in a MariaDB instance is with the Platform.
 platform db:dump
 ```
 
-If you have multiple SQL databases it will prompt you which one to export. You can also specify one by relationship name explicitly:
+By default the file will be uncompressed. If you want to compress it, use the `--gzip` (`-z`) option:
 
 ```bash
-platform db:dump --relationship database
+platform db:dump --gzip
```
Crell: Do we only support gzip?  bzip2 (as the previous example used) has much better compression for things like giant text files.

pjcdawkins: Good point... for right now, gzip, but bz2 is available in the containers so I could add support for that (very soon)

Crell: OK. Do you want to merge these docs now and adjust later or wait for adding bzip2?  I'm fine either way.

Crell:
I also wonder about removing the examples that include --stdout.  That's not obvious elsewhere in the docs that it exists, but it is of course useful.

pjcdawkins:
OK, maybe we could document `--gzip` and leave in `--stdout | bzip > dump.sql.bz2`, then it shows both.

PR #461, opened 2017-01-11
Update php.md

kimkennof:
Added BCMath package on defaults list and on full list.


src/languages/php.md
```diff
@@ -73,6 +74,7 @@ This is the complete list of extensions that can be enabled:
 | apcu_bc      |     |     |     | *   | *   |
 | blackfire    | *   | *   | *   | *   | *   |
 | bz2          |     |     |     | *   | *   |
+| bcmath       | *   | *   | *   | *   | *   |
```
pjcdawkins: It shouldn't be in this table, if it's in the list above

PR #467, opened 2017-01-18
Persistent Redis

Crell:
In preparation for when it goes live. Don't publish until it does, obviously.

### DIFF
--- a/src/configuration/services/redis.md
+++ b/src/configuration/services/redis.md
@@ -1,14 +1,14 @@
 # Redis (Object cache)
 
-Redis is a high-performance in-memory object cache, well-suited for application level caching.
+Redis is a high-performance in-memory object store, well-suited for application level caching.
 
 See the [Redis documentation](https://redis.io/documentation) for more information.
 
-Platform.sh supports two different Redis configurations: One persistent (useful for key-value application data) and one in-memory-only (useful for application caching).  Aside from that distinction they are identical.
+Platform.sh supports two different Redis configurations: One persistent (useful for key-value application data) and one ephemeral (in-memory only, useful for application caching).  Aside from that distinction they are identical.
 
 ## Supported versions
 
-* 2.8 (cache only)
+* 2.8 (ephemeral only)
 * 3.0
 
 > **note**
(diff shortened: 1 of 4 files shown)


src/configuration/services/redis.md
```diff
@@ -4,17 +4,43 @@ Redis is a high-performance in-memory object cache, well-suited for application
...
 
 > **note**
-> Redis is configured to serve as a LRU cache with the eviction policy `allkeys-lru` - its storage is not persistent. You should not use it as a database.
-
-> **note**
 > The 3.0 version supports up to 64 different databases per instance of the service, while the 2.8 only allows for a single database
 
+### Ephemeral Redis
+
+The `redis` service type, available for both Redis 2.8 and 3.0, is configured to serve as a LRU cache with the eviction policy `allkeys-lru` - its storage is not persistent.  Only store disposable and regeneratable content in it.
```
ralt: "regeneratable" is a real word?

src/configuration/services/redis.md
```diff
@@ -4,17 +4,43 @@ Redis is a high-performance in-memory object cache, well-suited for application
...
+    type: redis:3.0
+```
+
+Data in an Ephemeral Redis instance is stored only in memory, and thus requires no disk space.  When the service hits its memory limit it will automatically evict old cache items to make room for new ones.
+
+### Persistent Redis
+
+The `redis-persistent` service type, available only for Redis 3.0, is configured for persistent storage. That makes it a good choice for fast application-level key-value storage.
+
+To add an Persistent Redis service, specify it in your `.platform/services.yaml` file like so:
```
ralt: "a" persistent redis service.

src/configuration/services/redis.md
```diff
@@ -4,17 +4,43 @@ Redis is a high-performance in-memory object cache, well-suited for application
 
 See the [Redis documentation](https://redis.io/documentation) for more information.
 
+Platform.sh supports two different Redis configurations: One persistent (useful for key-value application data) and one in-memory-only (useful for application caching).  Aside from that distinction they are identical.
```
pjcdawkins: Currently there are three adjective terms for the same thing: "in-memory-only", "cache only", and "Ephemeral" - just use one term, and define it here at the top

src/configuration/services/redis.md
Crell: In English, anything is a word. :smile: I'll reword.

src/configuration/services/redis.md
```diff
@@ -31,6 +57,8 @@ The format exposed in the ``$PLATFORM_RELATIONSHIPS`` [environment variable](/de
 }
 ```
 
+The format is identical regardless of whether it's a Persistent or Ephemeral service.
```
sruggier: In this sentence, persistent and ephemeral aren't proper nouns, so I think they should be lowercase.

PR #469, opened 2017-01-19
Clean up Robots.txt documentation

Crell:
Our current documentation for robots.txt is wrong and misleading and assumes Drupal 7.  Let's fix all 3 of those.

### DIFF
--- a/src/administration/web/configure-environment.md
+++ b/src/administration/web/configure-environment.md
@@ -85,6 +85,6 @@ Or to disable it for a specific environment other than the one that is currently
 platform environment:info -e ENVNAME restrict_robots false
 ```
 
-(Where `ENVNAME` is the name of the environment, which is generally the branch associated with it.)  Note that you will need to trigger a new deployment after changing this setting.
+where `ENVNAME` is the name of the environment.
 
 On a production instance (the master branch, after a domain has been assigned) the search-blocker is disabled and your application can serve a `robots.txt` file as normal.  However, you must ensure that the file is in your project's web root (the directory where the `/` location maps to) and your application is configured to serve it.  See [the location section in `.platform.app.yaml`](/configuration/app-container.md#locations).
--- a/src/administration/web/configure-environment.md
+++ b/src/administration/web/configure-environment.md
@@ -66,22 +66,25 @@ project configuration screen.
 
 ## Robots.txt
 
-By default, Platform.sh returns a restrictive `robots.txt` on all environments. If you need to provide a custom `robots.txt`, first disable the default one using the [Platform.sh CLI](/overview/cli.md) command below:
+By default, Platform.sh serves the following default and restrictive `robots.txt` on all non-production environments:
+
+```
+User-agent: *
+Disallow: /
+```
+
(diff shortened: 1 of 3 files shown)


src/administration/web/configure-environment.md
```diff
@@ -66,22 +66,25 @@ project configuration screen.
...
+Or to disable it for a specific environment other than the one that is currently checked out, execute the following:
 
-Then you have to serve your `robots.txt` by configuring ["/" location in `.platform.app.yaml`](/configuration/app-container.md#locations).
+```
+platform environment:info -e ENVNAME restrict_robots false
+```
 
-> **note**
-> A deployment is required after running the CLI command to make the changes effective.
+(Where `ENVNAME` is the name of the environment, which is generally the branch associated with it.)  Note that you will need to trigger a new deployment after changing this setting.
```
pjcdawkins: The sentence in parentheses is a bit weird. The `name` or `id` of an environment is always the same as the Git branch name.

pjcdawkins: Also, a deployment is triggered automatically (and, if for some reason it isn't, the CLI would warn).

src/frameworks/drupal7/faq.md
```diff
@@ -19,22 +19,9 @@ deployment hooks to automatically run the updates.
 
 ## How can I override the default robots.txt?
 
-If your project is using a make file, you will end up with the default
-`robots.txt` provided by Drupal.
+Platform.sh allows the use of a custom `robots.txt` file on production. For pre-production environments we provide a [default that blocks all search engines](/administration/web/configure-environment.html#robotstxt) (to keep pre-production sites out of search indexes).
```
pjcdawkins: Not 100% true about "production" vs "pre-production" (the truth being: the custom `robots.txt` would work any environment where `restrict_robots` is false, which is the default for just `master`)

src/frameworks/drupal7/faq.md
```diff
@@ -19,22 +19,9 @@ deployment hooks to automatically run the updates.
...
-User-agent: *
-Disallow: /
-```
-
-You can customize the `robots.txt` by placing your own version at the
-root of your repository.
-
-To override the behavior of robots.txt for Development environments you would
-currently need to use the API, contact our support for details.
+If using the `drupal` build mode with a Drush make file, place your `robots.txt` file in your project root, as a sibling of `.platform.app.yaml`.  It will get moved to the appropriate location automatically by the build process.  For all other cases just include the file in your web root normally.
```
pjcdawkins: `application root` not `project root`

src/administration/web/configure-environment.md
Crell: The parenthetical is pre-existing in that file. I just moved it around.  Should I just remove the "which is generally..." part of the sentence?

The note is also pre-existing.  Has that changed at some point, or was it always wrong? :smile:

Crell: Actually the environment would not match the branch name a user is working on when it's a GitHub PR.  Those get a special name.

src/frameworks/drupal7/faq.md
Crell: True. I'll reword.

src/administration/web/configure-environment.md
Crell: Hm, tested and you're right, it does redeploy automatically.  I'll update.

PR #472, opened 2017-01-21
Revise Going Live section to be up to date and clearer

Crell:
The Going Live page is long, hard to follow, and contains way too much information about SSL for most mortals.  Let's fix all of those.

(This is still a WIP; I just wanted a place to store it.)


src/configuration/routes.md
```diff
@@ -113,6 +113,24 @@ section for a detailed view on how to define routes that work with
...
+"https://{default}/":
+    type: upstream
+    upstream: "app:http"
+
+"http://{default}/":
+    type: redirect
+    to: "https://{default}/"
+```
+
+Alternatively, if you define only routes that use HTTPS Platform.sh will automatically create redirects from the HTTP to HTTPS version of each route.
```
Crell: Someone please verify this is the case.  @damz @ralt ?

src/configuration/routes.md
```diff
@@ -139,17 +157,7 @@ For your master environment, this would function as a catch-all domain.
...
-put any subdomain before the triple dashes to reach your application.
-HTTP request to both `http://foo---add-theme-vmwklxcpbi6zq.eu.platform.sh/` and
-`http://bar---add-theme-vmwklxcpbi6zq.eu.platform.sh/` URLs will be routed
-to your application properly. However, request to
-`*---add-theme-vmwklxcpbi6zq.eu.platform.sh` will not be routed since it is not
-a legitimate domain name.
+Let's say we have a project on the EU cluster whose ID is "vmwklxcpbi6zq" and we created a branch called "add-theme". The generated apex domain of this environment will be `add-theme-vmwklxcpbi6zq.eu.platform.sh`. If we have a `http://*.{default}/` route defined, the generated route will be `http://*---add-theme-vmwklxcpbi6zq.eu.platform.sh/`. This means you could put any subdomain before the triple dashes to reach your application. HTTP request to both `http://foo---add-theme-vmwklxcpbi6zq.eu.platform.sh/` and `http://bar---add-theme-vmwklxcpbi6zq.eu.platform.sh/` URLs will be routed to your application properly. However, request to `http://*---add-theme-vmwklxcpbi6zq.eu.platform.sh/` will not be routed since it is not a legitimate domain name.
```
Crell: I am torn on whether this belongs in the routing section or the going live section.  There's a lot of overlap there, still, which makes me quite uncomfortable.

src/development/going-live/multi-app.md
```diff
@@ -0,0 +1,75 @@
+# Serving multiple applications
```
Crell: This page is still incomplete, pending it being way easier to add multiple domains in the first place.  Feedback still welcome but it's likely buggy.

src/development/going-live/ssl.md
```diff
@@ -0,0 +1,180 @@
+# SSL in Production
```
Crell: I am not sure if this page is needed at all, frankly.  Shouldn't most 3rd party SSL providers have their own instructions and generate keys for you?  Do we really need this?

src/development/going-live.md
```diff
@@ -1,103 +1,89 @@
...
+4) Your router responds with an HTTP 301 redirect to `www.mysite.com`.
+5) Your browser looks up `www.mysite.com` and, as above, gets an alias for `www---master-def456-abc123.us.platform.sh`, which is IP 1.2.3.4.
+6) Your browser sends a request to `1.2.3.4` for domain `www.mysite.com`.  Your router passes the request through to your application which in turn responds with whatever it's supposed to do.
 
-## 3 - DNS
+On subsequent requests, your browser will know to simply connect to `1.2.3.4` for domain `www.mysite.com` and skip the rest.  The entire process takes only a few milliseconds.
 
-Configure your DNS provider to point your domain to your
-[Platform.sh](https://platform.sh) Master environment.
+## Old Stuff, is it relevant?
```
Crell: Anything past here in this page I am not sure if we should keep or reword.  My registrar lets me CNAME to foo.com just fine.  Also, my read of this section is "you can't actually serve foo.com from us even though all of our documentation says you can, because reasons."  Which... seems both incorrect and not a good thing to say. :smile: If kept it would still need to be greatly rewritten.

src/development/going-live.md
```diff
@@ -1,103 +1,89 @@
...
-Web Interface or within your `.platform/routes.yaml` file.
+## Configure your DNS provider
 
-### Single hostname
+Configure your DNS provider to point your domain to your Platform.sh Master environment.
 
-If you want your site to only be available at `http://mydomain.com` and
-have `http://www.mydomain.com` redirect to `http://mydomain.com`, you
-need define your `routes.yaml` as follow:
+The way to do so will vary somewhat depending on your registrar, but nearly all registrars should allow you to set a CNAME.  Some will call it an Alias or similar alternate name, but either way the intent is to say "this domain should always resolve to... this other domain".  Add a CNAME record from your desired domain (`example.com`) to the master environment hostname you wrote down earlier.
```
xtfer: I see what you are trying to achieve here in terms of simplicity, however I think this will confuse anyone who actually understands CNAMEs, and their inability to be used on apex records. The instruction to add a CNAME from "example.com", for example, is both misleading and almost impossible. :)

I would suggest an alternative approach where we first provide advice on the apex domain, then on subdomains. Happy to take a shot at that.

src/development/going-live.md
```diff
@@ -1,103 +1,89 @@
...
 ```
 
-> **note**
-> You can test those routes on your development environments with:
-> * `http://[branch]-[project-id].[region].platform.sh`
-> * `http://www-[branch]-[project-id].[region].platform.sh`
+That will result in two domains being created on Platform.sh: `master-def456-abc123.us.platform.sh` and `www---master-def456-abc123.us.platform.sh`.  The former will automatically redirect to the latter.  In the `routes.yaml` file, `{default}` will automatically be replaced with `master-def456-abc123.us.platform.sh`.  In domain prefixes (like `www`), the `.` will be replaced with `---`.
 
-### HTTPS
+### Set your domain
```
xtfer: This appears to duplicate the Set your domain section, above

Crell: This is a 3-rd level section; the previous is 2nd level.  The intent is to mirror the section above with a concrete, non-abstract example.

src/development/going-live.md
Crell: Except I did it on my own web site by adding a CNAME for @.garfieldtech.com to www.garfieldtech.com, which in turn CNAMEs to Platform.sh.

Do I simply have a rare non-crappy registrar? :smile:

Crell:
I've updated the DNS section to reintegrate the discussion of Apex domains and then removed the rest of the redundant content.  I also dropped the multi-domain go-live page for now; it's in another branch to come back to once that's made approachable enough to document.

I consider this now ready for real review, so removing the WIP tag.

src/configuration/routes.md
pjcdawkins: It's wrong, anyway. The generated route will now include a random string after `add-theme`...

Crell: Updated the paragraph for that change. That still doesn't answer where it goes. :smile:

src/development/going-live/ssl.md
pjcdawkins: There are so many problems with this page. I agree it should be deleted.

Possibly there are tiny bits that could be re-used, like "Use the Platform.sh CLI to add the certificate".

Crell: I've removed that page and moved selected sections to the main document.  For now I stuck the SSL page into the `old` directory just in case we ever want that content back.

src/configuration/routes.md
mrkschan: I confirm this ... I tested that before I wrote that

src/development/going-live.md
```diff
@@ -1,371 +1,139 @@
...
+----------------- ------------------- ------ ----------------------------------------------
+you can do this   `www.domain.tld`    CNAME  `<environment>-<project>.<region>.platform.sh`
+you can't do this `domain.tld`        CNAME  `<environment>-<project>.<region>.platform.sh`
+----------------- ------------------- ------ ----------------------------------------------
 
-  The Common Name field must match the secure domain. You cannot purchase
-a certificate for the root domain, e.g. `example.com`, and expect to
-secure `www.example.com`. The inverse is also true. Each domain can have
-one certificate and it can be attached to the main domain or subdomain.
+However, using an A Record is not possible as the IP address of the server may change from time to time, especially with frequent redeployments as in Platform.sh's case.
```
mrkschan: I don't get this ... The IP address will be those on our gateways. They are considered stable. And, some DNS provider supports adding multiple 'A' records and load balance them. Customer may consider using this kind of 'A' records when 'CNAME' record is not applicable.

On the other hand, some DNS provider supports HTTP/HTTPS redirect on apex domain. That can be used as well when 'CNAME' is not applicable.

PR #476, opened 2017-01-24
Expand Solr docs to include expanded alternative

Crell:


src/configuration/services/solr.md
```diff
@@ -145,17 +147,49 @@ The relationships array would then look something like the following:
...
+                    configSet=mainconfig
+                    schema=arabic/schema.xml
+        endpoints:
+            na:
+                core: na_index
+            mideast:
+                core: mideast_index
+```
+
+In this example, the directory `.platform/configsets/solr6` contains the configuration definition for multiple cores.  There are then two cores created: `na_index` uses the defined configset, but specifically the `.platform/configsets/solr6/english/schema.xml` file, while `mideast_index` is identical except for using the `.platform/configsets/solr6/english/schema.xml` file.  Note that not all of Solrs core.properties file is available in the `core_properties` key.
```
sruggier: There's a copy-paste mistake in the second schema.xml path here.

Crell: Good catch, fixed.

src/configuration/services/solr.md
```diff
@@ -145,17 +147,49 @@ The relationships array would then look something like the following:
...
+                    configSet=mainconfig
+                    schema=arabic/schema.xml
+        endpoints:
+            na:
+                core: na_index
+            mideast:
+                core: mideast_index
+```
+
+In this example, the directory `.platform/configsets/solr6` contains the configuration definition for multiple cores.  There are then two cores created: `na_index` uses the defined configset, but specifically the `.platform/configsets/solr6/english/schema.xml` file, while `mideast_index` is identical except for using the `.platform/configsets/solr6/arabic/schema.xml` file.  Note that not all of Solr's core.properties file is available in the `core_properties` key.
```
sruggier: I think you mean to say something like:
> Note that not all core.properties features make sense to specify in the `core_properties`. Some keys, such as `name` and `dataDir`, are not supported, and may result in a solrconfig that fails to work as intended, or at all.

Crell: Updated, using your language directly.

sruggier:
Looks good to me :)

PR #478, opened 2017-01-24
Update userwidget to add talkus

dudenhofer:


acemtp:
Talkus founder here, don't hesitate if you need any help to setup that!

PR #481, opened 2017-01-29
URL Mangling has triply dashes

OriPekelman:


Crell:
This is going to conflict with #472, which is already gutting that page.  Please verify that PR has the right URLs and we'll do it all at once.  (Also review the whole PR and merge if appropriate.)

PR #485, opened 2017-02-02
Ini settings

Crell:
Better document PHP.ini settings and their defaults.

### DIFF
--- a/src/languages/php.md
+++ b/src/languages/php.md
@@ -231,21 +231,6 @@ Another example is to set the timezone of the PHP runtime (though, the timezone 
 date.timezone = "Europe/Paris"
 ```
 
-After pushing your file, you can check that the custom PHP configuration has been added to your environment by logging in via SSH and type:
-
-```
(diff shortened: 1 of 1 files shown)


src/languages/php.md
```diff
@@ -233,11 +231,10 @@ Another example is to set the timezone of the PHP runtime (though, the timezone
 date.timezone = "Europe/Paris"
 ```
 
-After pushing your file, you can check that the custom PHP configuration
-has been added to your environment by logging in via SSH and type:
+After pushing your file, you can check that the custom PHP configuration has been added to your environment by logging in via SSH and type:
```
pjcdawkins: This bit is weird - I suggest removing it. If you've pushed the right thing, it is _guaranteed_ to be deployed... why would you need to log in to confirm?

Crell: I don't know; that line has been in the docs forever. :smile: I think the intent is to make sure you didn't typo something.  The Redis docs have a similar stanza.  I'm open to removing it, though.

PR #487, opened 2017-02-03
Add description of endpoints and MySQL's use of them

Crell:
This covers the new support for multiple MySQL databases on a single service.

### DIFF
--- a/src/configuration/services.md
+++ b/src/configuration/services.md
@@ -75,7 +75,7 @@ In order for a service to be available to an application in your project (Platfo
 
 All services offer one or more `endpoints`.  An endpoint is simply a named set of credentials that can be used to access the service from other applications or services in your project.  Only some services support multiple user-defined endpoints.  If you do not specify one then one will be created with a standard defined name, generally the name of the service type (e.g., `mysql` or `solr`).  An application container, defined by a `.platform.app.yaml` file, always exposes and endpoint named `http` to allow the [router](/configuration/routes.md) to forward requests to it.
 
-When definign relationships in a configuration file you will always address a service as `<servicename>`.`<endpoint>`.  See the appropriate service page for details on how to configure multiple endpoints for each service that supports it.
+When defining relationships in a configuration file you will always address a service as `<servicename>`:`<endpoint>`.  See the appropriate service page for details on how to configure multiple endpoints for each service that supports it.
 
 
 ## Connecting to a service
--- a/src/configuration/services.md
+++ b/src/configuration/services.md
@@ -70,6 +70,14 @@ For example, the current default storage amount per project is 5GB (meaning 5120
 
 In order for a service to be available to an application in your project (Platform.sh supports not only multiple backends but also multiple applications in each project) you will need to refer to it in the [.platform.app.yaml](/configuration/app-containers.md) file which configures the *relationships* between applications and services.
 
+
+## Endpoints
+
+All services offer one or more `endpoints`.  An endpoint is simply a named set of credentials that can be used to access the service from other applications or services in your project.  Only some services support multiple user-defined endpoints.  If you do not specify one then one will be created with a standard defined name, generally the name of the service type (e.g., `mysql` or `solr`).  An application container, defined by a `.platform.app.yaml` file, always exposes and endpoint named `http` to allow the [router](/configuration/routes.md) to forward requests to it.
+
+When definign relationships in a configuration file you will always address a service as `<servicename>`.`<endpoint>`.  See the appropriate service page for details on how to configure multiple endpoints for each service that supports it.
+
+
 ## Connecting to a service
 
 Once a service is running and exposed as a relationship, its appropriate credentials (host name, username if appropriate, etc.) will be exposed through the `PLATFORM_RELATIONSHIPS` environment variable.  The structure of each is documented on the appropriate service's page, along with sample code for how to connect to it from your application. Note that different applications manage configuration differently so the exact code will vary from one application to another.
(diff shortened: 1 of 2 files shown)


src/configuration/services/mysql.md
```diff
@@ -79,6 +79,71 @@ foreach ($relationships['database'] as $endpoint) {
...
+
+On the 10.0 version and later of MariaDB it is possible to define multiple databases as well as multiple users with different permissions.  To do so requires defining multiple endpoints.  Under the `configuration` key of your service there are two additional keys:
+
+* `schemas`:  This is a YAML array listing the databases that should be created.  If not specified, a single database named `main` will be created.
+* `endpoints`: This is a nested YAML array defining different credentials.  Each endpoint may have access to one or more schemas (databases), and may have different levels of permission on each.  The valid permission levels are:
+** `ro` - Using this endpoint only SELECT queries are allowed.
+** `rw` - Using this endpoint SELECT queries as well INSERT/UPDATE/DELETE queries are allowed.
+** 'admin`' - Using this endpoint all queries are allowed, including DDL queries (CREATE TABLE, DROP TABLE, etc.).
+
+If no endpoints are defined, a single endpoint will be created named `mysql` that has `admin` access to all defined databases.
```
ralt: "a single endpoint named `mysql` will be created" would be better, no?

src/configuration/services.md
```diff
@@ -70,6 +70,14 @@ For example, the current default storage amount per project is 5GB (meaning 5120
 
 In order for a service to be available to an application in your project (Platform.sh supports not only multiple backends but also multiple applications in each project) you will need to refer to it in the [.platform.app.yaml](/configuration/app-containers.md) file which configures the *relationships* between applications and services.
 
+
+## Endpoints
+
+All services offer one or more `endpoints`.  An endpoint is simply a named set of credentials that can be used to access the service from other applications or services in your project.  Only some services support multiple user-defined endpoints.  If you do not specify one then one will be created with a standard defined name, generally the name of the service type (e.g., `mysql` or `solr`).  An application container, defined by a `.platform.app.yaml` file, always exposes and endpoint named `http` to allow the [router](/configuration/routes.md) to forward requests to it.
+
+When definign relationships in a configuration file you will always address a service as `<servicename>`.`<endpoint>`.  See the appropriate service page for details on how to configure multiple endpoints for each service that supports it.
```
danmilon: definign typo.

danmilon: Maybe use the colon notation we have in the yaml files, instead of dot for relationships? Like `<service-name>:<endpoint-name>`.

src/configuration/services/mysql.md
```diff
@@ -79,6 +79,71 @@ foreach ($relationships['database'] as $endpoint) {
 2. MySQL Errors such as "PDO Exception 'MySQL server has gone away'" are usually simply the result of exhausting your existing diskspace. Be sure you have sufficient space allocated to the service in [.platform/services.yaml](/configuration/services.md).
 
 
+## Multiple databases
+
+On the 10.0 version and later of MariaDB it is possible to define multiple databases as well as multiple users with different permissions.  To do so requires defining multiple endpoints.  Under the `configuration` key of your service there are two additional keys:
```
danmilon: maybw s/MariaDB/MySQL ? cause this is not engine-specific, and also the service name is mysql.

src/configuration/services/mysql.md
danmilon: same thing to me.

src/configuration/services.md
Crell: Argh, wrong punctuation. It should be a colon.

src/configuration/services/mysql.md
Crell: Calling it MySQL is a problem when talking about versions.  There is no such thing as MySQL 10.0, there's only MariaDB 10.0.  We need to clear up these names one of these days.  The naming is a bit tricky.

danmilon: You're right. What do you think about referring explicitly to the "MariaDB service"? To me the current wording implies that this is a feature that was added to the MariaDB software itself. If you dont see that as a native speaker, fine by me.

Crell: I reworded it to not say the name, since technically it's our service that only supports it as of 10.0.  MariaDB/MySQL itself has always supported multiple DBs, of course.

OriPekelman:
I am fine with this going out before the blog post / newsletter etc..

src/configuration/services.md
```diff
@@ -70,6 +70,14 @@ For example, the current default storage amount per project is 5GB (meaning 5120
 
 In order for a service to be available to an application in your project (Platform.sh supports not only multiple backends but also multiple applications in each project) you will need to refer to it in the [.platform.app.yaml](/configuration/app-containers.md) file which configures the *relationships* between applications and services.
 
+
+## Endpoints
+
+All services offer one or more `endpoints`.  An endpoint is simply a named set of credentials that can be used to access the service from other applications or services in your project.  Only some services support multiple user-defined endpoints.  If you do not specify one then one will be created with a standard defined name, generally the name of the service type (e.g., `mysql` or `solr`).  An application container, defined by a `.platform.app.yaml` file, always exposes and endpoint named `http` to allow the [router](/configuration/routes.md) to forward requests to it.
```
Crell: Well it doesn't correspond to a port, or a protocol, or really anything else other than a set of credentials.  As we've implemented it I don't know what else we can call it that isn't inaccurate.

@damz This is your design. :smile: Any suggestions?

pjcdawkins: If is _is_ a set of credentials, rather than just something that 1-1 maps to a set of credentials, I'd prefer we call it a "user", or... a "set of credentials"

mrkschan: I'm actually having confusion with [DSN](https://en.wikipedia.org/wiki/Data_source_name).

PR #494, opened 2017-02-13
Opt-out redis integration during installation

Crell:
Per changes here, recommended by Berdir: https://github.com/md-systems/platformsh-example-thunder/pull/1/files/6189b0cd5e88cf772c1aa5f20e7835d2f7f07c40..60d6d1f04771145457a44699d07872e057d61e09

This lets the Redis code work correctly during install, in case someone wants to start with Redis from the get-go.  (Which we probably do want anyway.)


Crell:
I've been considering that. There's even a Jira ticket for it. The main question is if doing so would be detrimental as it is more resources being used by default.  I haven't looked into whether that would push anyone toward implicit caps they don't want.

Definitely something we should consider, but let's get this into the docs for now.

PR #498, opened 2017-02-15
Use search-plus search plugin

Crell:
Let's see if we like it's search index better.

PR #499, opened 2017-02-15
Use Search Pro plugin

Crell:
Let's see if we like it's search index better.

### DIFF
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -53,5 +53,5 @@ web:
                     allow: true
                 ^/robots\.txt$:
                     allow: true
-                search_index\.json$:
+                search_plus_index\.json$:
                     allow: true
--- a/book.json
+++ b/book.json
@@ -10,7 +10,10 @@
     "gtm",
     "edit-link",
     "toc",
-    "-sharing"
+    "-sharing",
+    "-lunr",
+    "-search",
+    "search-pro"
   ],
   "pluginsConfig": {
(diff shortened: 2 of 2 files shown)


mrkschan:
👎 on this ... it does not show complete word on the search result ... 

https://pr-499-qx3e57a-t2llqeifuhpzg.eu.platform.sh/?q=socket

> lopment environments. It replaces the dot (.). WebSocket routes To use WebSocket on a route, cache must be disabled because WebSocket is incompatible with buffering, which is a requirement of caching on our router. Here is an example to define a route that serves WebSocket: "https://{default}/ws": type: upstream upstream: "app:http" cache: enabled: false...

> tarted by commands.start above). It has two keys: socket_family: Default: tcp. Describes whether your application will listen on a Unix socket (unix) or a TCP socket (tcp). protocol: Specifies whether your application is going to receive incoming requests over HTTP (http) or FastCGI (fastcgi). The default varies depending on which application runtime you're using. Other values will be supported in the future. On a PHP container with FPM there is almost never a reason to set the upstream explicit...

mrkschan:
sitedocs platform sh socket at duckduckgo 26-5-2017 11-34-47 pm

It's missing this indexed by [ddg](https://duckduckgo.com/?q=site%3Adocs.platform.sh+socket&t=ffsb&atb=v65-6&ia=web)

Fidelix:
Closed in favor of #498

PR #502, opened 2017-02-22
Document regex capture groups in rules blocks

Crell:
Someone from engineering please confirm that this is what we did. :smile:


src/configuration/app-containers.md
```diff
@@ -316,20 +316,45 @@ web:
             passthru: "/index.php"
             index:
                 - index.php
+            # No caching, unless the PHP script sends a cache header.
```
pjcdawkins: I thought `expires` only applies to static files, and doesn't affect scripts at all?

Crell: Isn't that implied in that comment? Could we clarify it further?

pjcdawkins: > No caching (for static files).

pjcdawkins:
I corrected the regex, and because we have examples of YAML regular expressions on that page, I changed all the YAML in the file to single quotes instead of double quotes. Someone else will need to review this now.

src/configuration/app-containers.md
```diff
@@ -462,7 +483,7 @@ Your `.platform.app.yaml` file should include:
 ```yaml
 hooks:
   deploy: |
-    if [ $PLATFORM_BRANCH = "master" ]; then
+    if [ "$PLATFORM_BRANCH" = master ]; then
```
Crell: The "master" doesn't need to be quoted?  Are you sure about that?

pjcdawkins: Yes (we know exactly what's in the string, so the quotes there are superfluous, whereas we know less about what would be in the PLATFORM_BRANCH variable)

src/configuration/app-containers.md
```diff
@@ -306,30 +301,56 @@ It has a few subkeys listed below:
...
             expires: 300
             passthru: true
+            allow: false
+            rules:
+                # Only allow static image files from the images directory.
+                '\.(jpe?g|png|gif|svgz?|ico|bmp)$':
+                    allow: true
+```
+
+### [Example] Advanced rewrite rules
```
mrkschan: @sruggier Is this block superseding https://docs.platform.sh/configuration/app-containers.html#url-rewriting?

PR #510, opened 2017-03-01
--room parameter doesn't exist

isholgueras:
The correct parameter is --hipchat-room-id


Crell:
Hm, are you sure?  It looks like it's "room" in our code, both on the server and in the CLI client.

pjcdawkins:
Ah. Looks like a regression in the CLI

pjcdawkins:
... fixed in v3.14.0 of the CLI - sorry about that (and thanks for the PR)

isholgueras:
That's why I made the PR:
issue

No problem @pjcdawkins.

Thanks for the fix.

PR #512, opened 2017-03-02
Mention PLATFORM_RELATIONSHIPS and port in sample usage

mrkschan:
User should always refer to PLATFORM_RELATIONSHIPS for hostname and port
number of a service. Our sample usage should not skip them.

### DIFF
--- a/src/configuration/services/mysql.md
+++ b/src/configuration/services/mysql.md
@@ -146,11 +146,10 @@ configuration:
 
 ## Access your MariaDB service
 
-Assuming your MariaDB relationship is named `database`, you can access it by connecting
-to a host named `database.internal` using the MySQL command line client.
+Assuming your MariaDB relationship is named `database`, the host name and port number obtained from `PLATFORM_RELATIONSHIPS` would be `database.internal` and `3306`. Open an [SSH session](/development/ssh.md) and run the MySQL command line client.
 
 ```bash
-mysql -h database.internal -u user main
+mysql -h database.internal -P 3306 -u user main
 ```
 
 Outside the application container, you can use Platform CLI `platform sql`.
--- a/src/configuration/services/redis.md
+++ b/src/configuration/services/redis.md
@@ -93,9 +93,8 @@ if (getenv('PLATFORM_RELATIONSHIPS')) {
 
 ## Using redis-cli to access your Redis service
 
-Assuming your Redis relationship is named `rediscache`, you can access it by
-connecting to a host named `rediscache.internal` using the redis-cli tool. Open an [SSH session](/development/ssh.md) and access the Redis server as follows:
+Assuming your Redis relationship is named `rediscache`, the host name and port number obtained from `PLATFORM_RELATIONSHIPS` would be `rediscache.internal` and `6379`. Open an [SSH session](/development/ssh.md) and access the Redis server using the `redis-cli` tool as follows:
 
 ```bash
-redis-cli -h rediscache.internal
+redis-cli -h rediscache.internal -p 6379
 ```


mrkschan:
Shall we consider preinstalling `jq` binary from https://stedolan.github.io/jq/ so that the sample always refers to `PLATFORM_RELATIONSHIPS` and it becomes copy-paste friendly? E.g.

```
export DB_HOST="$(echo $PLATFORM_RELATIONSHIPS | base64 -d | ./jq-linux64 .database[0].host)"
export DB_PORT="$(echo $PLATFORM_RELATIONSHIPS | base64 -d | ./jq-linux64 .database[0].port)"
export DB_USER="$(echo $PLATFORM_RELATIONSHIPS | base64 -d | ./jq-linux64 .database[0].username)"
export DB_NAME="$(echo $PLATFORM_RELATIONSHIPS | base64 -d | ./jq-linux64 .database[0].path)"
mysql -A -P $DB_PORT -h $DB_HOST -u $DB_USER $DB_NAME
```

src/configuration/services/mysql.md
```diff
@@ -146,11 +146,10 @@ configuration:
 
 ## Access your MariaDB service
 
-Assuming your MariaDB relationship is named `database`, you can access it by connecting
-to a host named `database.internal` using the MySQL command line client.
+Assuming your MariaDB relationship is named `database`, the host name and port number obtained from `PLATFORM_RELATIONSHIPS` would be `database.internal` and `3306`. Open an [SSH session](/development/ssh.md) and run the MySQL command line client.
 
 ```bash
-mysql -h database.internal -u user main
+mysql -A -h database.internal -P 3306 -u user main
```
Crell: Why add the -A here?  Rehashing on connect has never been particularly slow for me.

Crell:
I'm fine with including jq; but I worry that the copy-paste-able excerpt becomes so complex as to not be a benefit.  Plus, it's not truly copyable since the relationship name is only a recommendation; we don't know what someone may have named their relationships.  If they named it something different then both the current sample and the proposed one would be equally wrong.

src/configuration/services/mysql.md
mrkschan: I don't really know why it's a bit slow when the client is connecting to our HA MariaDB Galeria cluster. The question is do we want our customer to enjoy fast startup anywhere versus auto-complete.

mrkschan:
@Crell Nice catch on the copy-paste issue with the relationship name. Let's keep the way we're doing now.

src/configuration/services/mysql.md
Crell: I don't think this is really the place to mention that.  The mysql command has its own extensive documentation with a dozen extra switches that could be used if desired. Let's stick to the ones directly relevant to the task at hand.

mrkschan: Updated.

src/configuration/services/mysql.md
```diff
@@ -146,11 +146,10 @@ configuration:
 
 ## Access your MariaDB service
 
-Assuming your MariaDB relationship is named `database`, you can access it by connecting
-to a host named `database.internal` using the MySQL command line client.
+Assuming your MariaDB relationship is named `database`, the host name and port number obtained from `PLATFORM_RELATIONSHIPS` would be `database.internal` and `3306`. Open an [SSH session](/development/ssh.md) and run the MySQL command line client.
```
Crell: Just realized, this should be "open an SSH connection and run..."  The "connection" is missing.

src/configuration/services/redis.md
```diff
@@ -93,9 +93,8 @@ if (getenv('PLATFORM_RELATIONSHIPS')) {
 
 ## Using redis-cli to access your Redis service
 
-Assuming your Redis relationship is named `rediscache`, you can access it by
-connecting to a host named `rediscache.internal` using the redis-cli tool. Open an [SSH session](/development/ssh.md) and access the Redis server as follows:
+Assuming your Redis relationship is named `rediscache`, the host name and port number obtained from `PLATFORM_RELATIONSHIPS` would be `rediscache.internal` and `6379`. Open an [SSH session](/development/ssh.md) and access the Redis server using the `redis-cli` tool as follows:
```
Crell: Same as above.

src/configuration/services/mysql.md
mrkschan: @Crell i don't get this ... Do you want to replace `Open an SSH session` with `Open an SSH connection`?

mrkschan: FYI, it was using `SSH session` previously https://github.com/platformsh/platformsh-docs/pull/512/files/4b2e990617a3793dc5a192e9e4f057e0b1ba92a1#diff-fecddd292d23722cd2e396ab813d2a90L97

Crell: ... No, that's me reading too fast and not registering that the word "session" was already part of the link.  False alarm, please ignore me.

PR #513, opened 2017-03-02
Route cache

Crell:
Some mundane improvements to the caching page.


src/configuration/routes/cache.md
```diff
@@ -2,15 +2,11 @@
...
+Many hosting solutions are adding an additional layer (Varnish...) to handle caching. Platform.sh supports HTTP caching right at the web server level.
 
-If you disable caching, Platform.sh serves the files that are stored in
-the application directly. For example with Drupal, this means that all
-HTTP requests will bootstrap Drupal and query the database. When the
-cache is enabled, if the page has been stored in the Nginx cache, it
-won't access Drupal.
+By default, the router will respect whatever cache headers are sent by the application (either from your application code or from the `.platform.app.yaml` configuration), but the presence of any cookie headers will cause the cache to be bypassed.
+
+If you disable caching, Platform.sh serves the files that are stored in the application directly. For example with Drupal, this means that all HTTP requests will bootstrap Drupal and query the database. When the cache is enabled, if the page has been stored in the Nginx cache, it won't access Drupal.
```
JGrubb: re: "all HTTP requests" - This would only be the case for requests that hit Drupal.  Static files (which could also be cached at the router) would not cause a Drupal bootstrap but would be served the webserver in the app container.

src/configuration/routes/cache.md
```diff
@@ -178,3 +176,5 @@ default value.
 If the response does not have a `Cache-Control` header, the
 `default_ttl` key is used to define the cache duration, in seconds. The
 default value is `0`.
+
+If the application code returns a Cache-Control header, or if your `.platform.app.yaml` file is configured to set a cache lifetime for static assets, then this value is ignored in favor of the application container's headers.
```
JGrubb: "in favor of the application's headers" maybe?

src/configuration/routes/cache.md
Crell: That paragraph is in the current text verbatim.  It is over-reaching, though.  I think the intent is to try and scare people into using caching because OMG the application is expensive, but I think that's an over-reach.  how should we revise it?  Or just remove it entirely?

src/configuration/routes/cache.md
Crell: Well, there's the subtle question of whether "the application" includes the user's code, or the user's code plus the configured nginx instance. It technically means both depending on the sentence. :smile:

src/configuration/routes/cache.md
JGrubb: "If you disable caching, Platform.sh serves the static files or requests generated by the application directly." /STOP

?

src/configuration/routes/cache.md
JGrubb: Per Slack convo and the expires header which _can_ be set in web.locations, I can let this go now...

PR #516, opened 2017-03-06
Update going-live.md

robertDouglass:
You should actually configure your DNS provider first, as that will almost always take longer, and it requires information (the full URL of Master) that goes away in the Set Your Domain step.

This PR supersedes my previous PR and they shouldn't be merged together.

### DIFF
--- a/src/development/going-live.md
+++ b/src/development/going-live.md
@@ -52,7 +52,7 @@ Platform.sh recommends ensuring that your DNS Provider supports dynamic apex dom
 
 ## Set your domain
 
-First step is to [add your domain](/administration/web/configure-project.html#domains).
+The next step is to [add your domain](/administration/web/configure-project.html#domains).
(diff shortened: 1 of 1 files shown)


src/development/going-live.md
```diff
@@ -64,6 +50,19 @@ changes, the IP address for the mapped domain changes automatically as well.
 
 Platform.sh recommends ensuring that your DNS Provider supports dynamic apex domains before registering your domain name with them.  If you are using a DNS Provider that does not support dynamic apex domains then you will be unable to use `example.com` with Platform.sh, and will need to use only `www.example.com` (or similar) instead.
 
+## Set your domain
+
+First step is to [add your domain](/administration/web/configure-project.html#domains).
```
Crell: This should probably say "The next step" now, since it's no longer first.

PR #518, opened 2017-03-15
Influxdb documentation

Crell:
It's coming momentarily, so let's get the docs ready.


src/configuration/services/influxdb.md
```diff
@@ -0,0 +1,80 @@
...
+
+* 1.2
+
+## Relationship
+
+The format exposed in the ``$PLATFORM_RELATIONSHIPS`` [environment variable](/development/variables.md#platformsh-provided-variables):
+
+```json
+{
+   "servicename" : [
```
mrkschan: @Crell  Is this a new standard? The relationship in below section is `timedb`. If we want a generic one, it should be `relationshipname` instead of `servicename`.

PR #520, opened 2017-03-22
Consolidating Cron Notes

JGrubb:
Also adding a note about jitter.

### DIFF
--- a/src/configuration/app-containers.md
+++ b/src/configuration/app-containers.md
@@ -523,12 +523,12 @@ crons:
 > 
 > Cron runs are executed using the dash shell, not the bash shell used by normal SSH logins. 
 > In most cases that makes no difference but may impact some more involved cron scripts.
-> 
-> All server times are in UTC, and hence so are the cron schedules.
 >
 > A variable delay is added to each cron job in each project in order to prevent host overloading
 > at midnight every night when every project runs their nightlies.  Your crons will *not* run exactly
 > at the time that you specify, but will be delayed by 0-300 seconds.
+
+All server times are in UTC, and hence so are the cron schedules.
 
 
 ## URL rewriting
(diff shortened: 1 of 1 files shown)


src/configuration/app-containers.md
```diff
@@ -521,7 +519,17 @@ crons:
         cmd: 'cd web ; drush queue-run aggregator_feeds'
 ```
 
-Also note that all server times are in UTC, and hence so are the cron schedules.
+> ** Cron notes **
```
Crell: I don't think we even need to put this in a notes box.  It's just more info about cron, let it be prose.

JGrubb: Ha, ok.  My eye grepped for "note" and landed on that.

Crell: I meant any of it. :smile: I'm not generally a fan of callout boxes.

src/configuration/app-containers.md
```diff
@@ -521,7 +519,17 @@ crons:
...
+> 
+> Cron runs are executed using the dash shell, not the bash shell used by normal SSH logins. 
+> In most cases that makes no difference but may impact some more involved cron scripts.
+>
+> A variable delay is added to each cron job in each project in order to prevent host overloading
+> at midnight every night when every project runs their nightlies.  Your crons will *not* run exactly
+> at the time that you specify, but will be delayed by 0-300 seconds.
+
+All server times are in UTC, and hence so are the cron schedules.
```
ralt: You can specify the timezone for the cron, no?

Crell: Can you?  If you know how that's a good thing to include. :smile:

ralt: It will be whatever you put in the `.platform.app.yaml`'s `timezone` key.

Crell: The timezone key is not currently documented anywhere in the file. *sigh*  Let's get back to that later.

PR #521, opened 2017-03-27
HHVM 3.15 & 3.18

GuGuss:


src/changelog/2017.md
```diff
@@ -1,13 +1,23 @@
...
+# February 2017
+
+## HHVM 3.15 and 3.18
+
+Two new HHVM versions are now available.
+
+## InfluxDB 1.2
+
+A new service type is available for InfluxDB 1.2, a time-series database.  See the [InfluxDB documentation](/configuration/services/influxdb.md) for more information.
+
```
Crell: InfluxDB is already listed above for March, which is when it went live.  No need to list it in February, too. :smile:

PR #528, opened 2017-04-03
Add documentation for Go support

Crell:


src/languages/go.md
```diff
@@ -0,0 +1,70 @@
+# Ruby
```
pjcdawkins: Probably not right

Crell: Fact...

src/languages/go.md
```diff
@@ -0,0 +1,70 @@
...
+
+```yaml
+web:
+    upstream:
+        socket_family: tcp
+        protocol: http
+
+    commands:
+        start: |
+          ./bin/myapp
```
ralt: 4 spaces pretty please

PR #530, opened 2017-04-05
Drop reduntant URL rewrite section

mrkschan:
There's a new [example](https://docs.platform.sh/configuration/app-containers.html#example-advanced-rewrite-rules) added to explain the same thing.


mrkschan:
That example was introduced in https://github.com/platformsh/platformsh-docs/pull/502/files

PR #533, opened 2017-04-12
Adjust 'sendmail' to 'mail' in Symfony email example

stefandoorn:
I didn't get it working with 'sendmail', but 'mail' does the trick. I think it cannot find the path to sendmail (`which sendmail` returns nothing).


ralt:
The problem with `mail` is that it's vulnerable to the return-path vulnerability. We need to pin down exactly how to properly use `sendmail`.

PR #539, opened 2017-04-28
Refactor application docs section

Crell:
After some discussion with Fabien, Ryan Weaver, Damien, and Beau at DrupalCon we decided to refactor the Application page in Docs.  The plan is to break it up into logical chunks (sub-pages of Application), with each chunk being first a reference section followed by one or more "How do I?" type examples.  The main Application page is just an overview and a few simple examples of degenerate cases.

This is also a step toward the multi-container logic that will be added to the application containers later.  It doesn't cover that yet but, hopefully, will make it easier to adapt the docs at that point.  Also, as part of this we should shift around the order of blocks in the examples so that they match the docs better. (That will be much more relevant once the multi-container logic is in place.)  The current order is more a historical accident than anything deliberate.


Crell:
This seems like a reasonable breakup of the text of that page.  Feedback welcome.  Note that the "Web" page is just a blind cut-paste right now and still needs to be greatly cleaned up, hence why this is still a WIP branch.

MacWeber:
Really cool to see all this information more accessible.

src/configuration/app/build.md
```diff
@@ -0,0 +1,135 @@
...
+
+The `.platform.app.yaml` file provides a number of ways to control how an application gets turned from a directory in Git into a running application.  There are three blocks that control different parts of the process: the build flavor, dependencies, and hooks.  The build process will run the build flavor, then install dependencies, then run the user-provided build hook.  The deploy process will run the deploy hook.
+
+## Build
+
+The `build` defines what happens when building the application.  Its only property is `flavor`, which specifies a default set of build tasks to run. Flavors are language-specific.
+
+### PHP (`composer` by default)
+
+`composer` will run `composer --no-ansi --no-interaction install --no-progress --prefer-dist --optimize-autoloader` if a `composer.json` file is detected.
```
rachellawson: Might be worth copying this (or, better, a link to this) into the readme of the default Drupal 8 project repo. That way, devs can be reminded to run essentially the same thing on their own local environments. Give some consistency between builds.

JGrubb:
Hey @Crell - what about `(required)` or something to that effect next to the params that are mandatory?  It's specified in the description of the parameter and there are only a handful, but..

src/configuration/app-containers.md
```diff
@@ -76,466 +73,18 @@ hooks:
...
-
-* [PHP](/languages/php.md)
+The `.platform.app.yaml` file is extremely flexible.  Depending on your needs it could be less than 10 lines long or over 100.  The only required keys are `name`, `type`, `disk`, and a minimal `web` block.  All others are optional.
 
-## Top level document roots
+The basic properties are described on their own pages.
 
-Platform.sh requires that the document root not be at the root of the project.  It is important for security that
-private file mounts are not web-accessible.
+* [`name`](/configuration/app/name.php) (required) - Sets the unique name of the application container.
```
pjcdawkins: All the links are broken in this list. I guess they should end with `.md`, not `.php`

src/configuration/app/build.md
Crell: Hm.  That composer line isn't Drupal specific; it's anything with a composer.json file.  I'm not sure that linking to it from every README is necessary, is it?  How else can we make that clear without just pasting the full line into every config template?

pjcdawkins: Maybe run `platform build` (`--clone`) on the local environment, to stay consistent?

src/configuration/app-containers.md
Crell: Argh.  Yep, typo.  Fixed.

src/configuration/app/build.md
rachellawson: Good point that we shouldn’t copy it - too many things to then maintain. A link to here in example repos, regardless of code type, does make sense, though. Something along the lines of “On deploy, composer will be executed as per {link} so you would probably want to do something similar locally for consistency"

PR #544, opened 2017-05-12
Drupal SAML documentation

Crell:
Instructions for setting up SimpleSAML on Drupal 7.  Still not fully tested as I don't have a SAML server available.

If someone has a IdP available, care to help test this? :smile: 

Drupal 8 version to come.  Probably largely the same, but may use Composer instead.  TBD.

### DIFF
--- a/src/frameworks/drupal7/simplesaml.md
+++ b/src/frameworks/drupal7/simplesaml.md
@@ -1,0 +1,91 @@
+# SimpleSAML
+
+SimpleSAMLphp is a library for authenticating a PHP-based application against a SAML server, such as Shibboleth.  Although Drupal has modules available to authenticate using SimpleSAML some additional setup is required.
+
+The following setup assumes you're using the `drupal` build flavor and building your site with Drush Make.  If not, you may need to adjust some paths in the configuration but the basics are the same.
+
+## Download the library
+
+First, download the 3rd party [SimpleSAMLphp library](https://simplesamlphp.org/download).  When you unpack the `tar.gz` file it will contain a directory named `simplesamplephp-???`, where the `???` is the version number of the library.  Place that directory at the root of your application, as a sibling of your `.platform.app.yaml` file, named `simplesamplephp`.  (The directory name doesn't really matter but removing the version number means that it won't change in future updates.) 
+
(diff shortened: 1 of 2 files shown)


src/frameworks/drupal7/simplesaml.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+(Adjust the version to whatever is current.)
+
+Much of the module configuration will depend on your Identity Provider (IdP).  However, the module also need to know the location of your `simplesamlphp_auth` module.  The easiest way to set it is to include the following at the end of your `settings.platformsh.php` file:
+
+```php
+// Set the path for the SimpleSAMLphp library dynamically.
+if (isset($_ENV['PLATFORM_APP_DIR'])) {
+  $conf['simplesamlphp_auth_installdir'] = dirname(realpath(__FILE__)) . '/simplesamlphp';
```
pjcdawkins: I think you wanted to use the `PLATFORM_APP_DIR` here?

Crell: No, because post-build it's not going to be in `PLATFORM_APP_DIR`, it's going to be in `sites/default`.  I could potentially build down `APP_DIR`, but decided to try going relative instead.  I could be convinced otherwise if you think that's more stable/generic for different site configurations.

pjcdawkins: Why check for PLATFORM_APP_DIR at all?

And, use `__DIR__` directly, it is already a realpath.

Crell: I wanted to check for something to only run when on Platform.sh, not local, and APP_DIR seemed like the easiest thing to check. It could have been anything, really.

And yeah, __DIR__ would be better. That's my old-PHP brain at work. :smile: Will fix.

pjcdawkins: OK, PLATFORM_APP_DIR would work for that, but wouldn't this directory still be inside sites/default on non-Platform.sh environments too?

Crell: Hm, a valid point.  This line probably is safe to run even on local, precisely because it's relative.  I'll just remove the conditional entirely.  Good catch!

PR #549, opened 2017-05-18
Adding some diagrams to config docs

JGrubb:


OriPekelman:
This is a huge improvement. So let's merge this and let Igor redo those prettily later ?

JGrubb:
Thanks @OriPekelman .  I'd like to get a thumbs up from @Crell before moving the prototype into production tho.

pjcdawkins:
"Typical Platform Professional Architecture" should just be "Typical Platform.sh Architecture"

OriPekelman:
I am going to say this again: I strongly believe that on documentation obvious / trivial corrections should, most of the time, just be directly committed to the branch. It is a sane default. You know this is git. We can revert.

pjcdawkins:
Agreed. If that's in response to me... I would certainly commit that correction, if I could easily edit the JPEG.

JGrubb:
I mean, the whole point of these diagrams was as prototype, and to have them replaced by someone with some design skill.  Let’s just incorporate that into the next version.

Crell:
OK, let's do this!

PR #550, opened 2017-05-19
Note that log files are now trimmed

Crell:
Not much else to say here.


pjcdawkins:
I merged, but out of interest, "trimmed" how? Truncated from the beginning? Rotated?

PR #553, opened 2017-05-24
Document variables support in .platform.app.yaml

Crell:


damz:
It is probably worth explicitly documenting the precedence. Which is:

* Environment-level variables (at run time);
* Project-level variables (at run time and build time);
* `.platform.app.yaml` (at run time and build time)

Crell:
@damz I was wondering about that.  So .platform.app.yaml-defined values are the lowest priority in case of conflict (viz, will always be overridden)?  Where does .environment fall in that list?

damz:
Yes, `.platform.app.yaml` is the lowest priority. The `.environment` file is different, it's just a shell script sourced after everything else is done. So it is higher priority, but it is also not limited to variables and can do completely arbitrary things.

src/configuration/app/variables.md
```diff
@@ -0,0 +1,115 @@
+# Variables
+
+Platform.sh provides a number of ways to set [environment variables](/development/variables.md), either globally or specific to a single environment.  For values that should be consistent between different environments (because they're configuring the application or runtime itself, generally) the easiest way to control them is to set them in the `.platform.app.yaml` file.
```
damz: We should probably standardize on "variables" instead of "environment variables".

src/configuration/app/variables.md
```diff
@@ -0,0 +1,115 @@
...
+variables:
+    env:
+        AUTHOR: 'Larry'
+```
+
+That will have the exact same runtime effect as setting the variable via the CLI as follows, except it will be versioned along with the code:
+
+```bash
+$ platform variable:set env:AUTHOR Larry
+```
```
damz: Maybe we can find an example with less Larry in it?

Crell: I will try not to take that personally. :-P  Suggestions?

src/configuration/app/variables.md
Crell: I've deliberately kept those two terms slightly separate based on my conversations with Fabien.  Calling values that only end up in $PLATFORM_VARIABLES "environment variables" is subtly incorrect.  But there's also Unix environment variables vs. variables defined per-environment, which are also "environment variables" in English but a different meaning of "environment".

Or are you referring to just the first line in the link text?

damz: Yes, I was referring to the first line. It should read "Platform.sh provides a number of ways to set variables".

src/configuration/app/variables.md
pjcdawkins: Isn't this more similar to `project:variable:set`, in effect? As in, you can use it at build time as well as runtime

src/configuration/app/variables.md
```diff
@@ -0,0 +1,115 @@
...
+variables:
+    d8config:
+        "system.site:name": 'My site rocks'
+```
+
+This will create a Platform.sh variable, that is, an item in the `$PLATFORM_VARIABLES` environment variable, named `d8config:system.site:name` with value "My site rocks".
+
+## Complex values
+
+The value for a variable may be more than just a string; it may also be a nested structure.  If the variable is in the `env` namespace, it will be mapped to a Unix environment variable as a JSON string.  If not, it will be included in the `PLATFORM_VARIABLES` environment variable as the native data type appropriate for the equivalent JSON structure.
```
pjcdawkins: Huh? When I read this I thought you meant the serialization/encoding would change depending on the container's main language.

But it doesn't - JSON is what defines the data types: it can have objects, arrays, and scalars, nothing else. From PHP or from Python it is exactly the same JSON. I suggest removing the reference to "native data type appropriate...".

src/configuration/app/variables.md
Crell: Hm, yes. Will update.

src/configuration/app/variables.md
Crell: Hm. I was intending to express that YAML arrays end up as JSON arrays end up as Python Lists, while YAML maps end up as JSON objects end up as Python dicts.  I guess I didn't do a very good job of that. :smile: How else can we express that?  Or just skip it and let people figure out that the $PLATFORM_VARIABLES value is what it is?  (Which I suppose is documented on another page already, so...)

src/development/variables.md
```diff
@@ -4,7 +4,9 @@ Platform.sh allows a high degree of control over both the build process and the
 
 ## Types of variables
 
-There are three different types of variable: User-provided project variables, user-provided environment variables, and Platform.sh-provided environment variables.  All of those may be simple strings or base64-encoded JSON-serialized values.
+There are four different types of variable: User-provided project variables, user-provided environment variables, application-provided variables, and Platform.sh-provided environment variables.  All of those may be simple strings or base64-encoded JSON-serialized values.
```
ralt: "User" should be lowercase everywhere, no?

src/development/variables.md
```diff
@@ -36,7 +38,13 @@ That will set a variable on the currently active environment (that is, the branc
...
+Environment variables are a good place to store values that apply only on Platform.sh and not on your local development environment. This includes API credentials for 3rd party services, mode settings if your application has a separate "Dev" and "Prod" runtime toggle, etc.
+
+### Application-provided variables
+
+It is also possible to [set Environment variables in code](/configuration/app/variables.md), using the `.platform.app.yaml` file.  These values of course will be the same across all environments and present in the Git repository, which makes them a poor fit for API keys and such.  This capability is mainly to define values that an application expects via an environment variable that should be consistent across all environments.  For example, the Symfony framework has a `SYMFONY_ENV` property that users may wish to set to `prod` on all environments to ensure a consistent build, or it may be used to set [PHP configuration values](#php-specific-variables).
```
ralt: The "PHP Symfony framework"? If we're going language agnostic, let's not assume knowledge of other-than-PHP developers?

src/development/variables.md
```diff
@@ -209,8 +217,8 @@ As the above logic is defined in a file in your Git repository you are free to c
 
 You can also provide a `.environment` file as part of your application, in your application root (as a sibling of your `.platform.app.yaml` file, or files in the case of a multi-app configuration).  That file will be sourced as a bash script when the container starts and on all SSH logins.  It can be used to set any environment variables directly, such as the PATH variable.  For example, the following `.environment` file will allow any executable installed using Composer as part of a project to be run regardless of the current directory:
 
- ```yaml
+ ```bash
 export PATH=/app/vendor/bin:$PATH
  ```
 
-Note that the file is sourced after all other environment variables above are defined, so they will be available to the script.  
+Note that the file is sourced after all other environment variables above are defined, so they will be available to the script.  That also means the .environment script has the "last word" on environment variable values and can override anything it wants to.
```
ralt: You maybe want backticks around `.environment`?

src/configuration/app/variables.md
```diff
@@ -0,0 +1,115 @@
+# Variables
+
+Platform.sh provides a number of ways to set [variables](/development/variables.md), either globally or specific to a single environment.  For values that should be consistent between different environments (because they're configuring the application or runtime itself, generally) the easiest way to control them is to set them in the `.platform.app.yaml` file.
+
+Only prefixed variables may be set from the `.platform.app.yaml` file.  Some prefixes have specific meaning while others are only significant to a particular application.  Nested variables will be automatically converted into a nested array or list structure as appropriate to the language.
```
Crell: Yes.  Only variables with a colon prefix.  Are you saying it's unclear what it means, or "seriously, we only support prefixed vars"?

src/development/variables.md
Crell: The first User is because it's the first character after a colon, thus beginning a new sentence.

src/development/variables.md
Crell: Good call, will add.

src/configuration/app/variables.md
ralt: Your next example shows something without prefix

Crell: Where?  The syntax doesn't support non-prefixed vars, that's the point.

PR #555, opened 2017-05-25
Add style for code tabs

dudenhofer:


Crell:
Definitely an improvement!  Can/should we put a border/background on the inactive tabs to make it more obvious that they're tabs?

dudenhofer:
yeah, makes sense. I was just keeping it simple initially. I updated the inactive styles, so feel free to check them out now.

Crell:
I like it.  Thanks!

PR #565, opened 2017-06-16
Add note for creating Elasticsearh index

mrkschan:
We control `number_of_shards` and `number_of_replicas` based on the available computing resources.
Wrong value on these settings may cause the service fails to start.


Crell:
"Note" should be singular.

Avoid the use of "We".  Instead, "These values will be set automatically based on available resources."

mrkschan:
@Crell updated.

PR #568, opened 2017-06-19
Remove vestigial FAQ

Crell:
There's now a CLI command for this so let's get rid of the FAQ.


mrkschan:
Are we expecting everyone to use the CLI for `git push`?

pjcdawkins:
When they want `--no-wait` or auto-activation, then, yes.

mrkschan:
Why don't just mention the use of CLI at the beginning of this FAQ section rather than removing it...

We have non Platform CLI users.

PR #574, opened 2017-06-28
Timezone and sizing

Crell:
They seem to be missing.  Let's make them not missing.

Note: the FAQ currently says that setting the system timezone for app containers is impossible, which is not true according to the code.  This PR corrects that.

### DIFF
--- a/src/configuration/app/size.md
+++ b/src/configuration/app/size.md
@@ -1,0 +1,24 @@
+# Custom sizing
+
+By default, Platform.sh will automatically select an appropriate resources (CPU and memory) for a container when it's deployed based on the plan size and the number of other containers in the cluster.  The more containers in a project the fewer resources each one gets and vice versa, with similar containers getting similar resources.
+
+Usually that's fine, but sometimes it's undesireable.  You may, for instance, want to have a queue worker container that you know has low memory and CPU needs, so it's helpful to give it fewer resources and another container more.  Or a given service may be very heavily used in your architecture so it needs all the resources it can take.  In those cases you can provide sizing hints to the system on a per-service basis.
+
+Every application container as well as every service in `services.yaml` supports a `size` key, which provides a hint to the system for how many resources to allocate to it.  The exact CPU and memory allocated will depend on the application or service type, as well as the plan size.
+
+Legal values for the `size` key are `AUTO` (the default), `S`, `M`, `L`, `XL`.
+
+Note that in a development environment most combination of sizes will successfully deploy, but on production instances limits on the plan capacity will be enforced.
+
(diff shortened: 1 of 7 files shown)


src/configuration/app/timezone.md
```diff
@@ -0,0 +1,23 @@
+# System timezone
+
+All Platform.sh containers default to running in UTC time.  Applications and application runtimes may elect to use a different timezone but the container itself runs in UTC.  That is usually fine and a correct configuration but occasionally there is cause to run the server in a specific timezone.
+
+## Setting the system timezone
+
+The system timezone for any service or app container may be set with the `timezone` property, whose value is one of the [tz database region codes](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones) such as `Europe/Paris` or `America/New_York`.
```
mrkschan: this's just for cron, no?

Crell: I presume it affects logs, too.  Logs and cron are about the only things to be affected.  We should verify.

src/configuration/app/size.md
```diff
@@ -0,0 +1,24 @@
...
+
+Usually that's fine, but sometimes it's undesirable.  You may, for instance, want to have a queue worker container that you know has low memory and CPU needs, so it's helpful to give that one fewer resources and another container more.  Or a given service may be very heavily used in your architecture so it needs all the resources it can take.  In those cases you can provide sizing hints to the system on a per-service basis.
+
+Every application container as well as every service in `.platform/services.yaml` supports a `size` key, which instructs the system how many resources to allocate to it.  The exact CPU and memory allocated will depend on the application or service type.
+
+Legal values for the `size` key are `AUTO` (the default), `S`, `M`, `L`, `XL`.
+
+Note that in a development environment this value is ignored and always set to `S`.  It will only take effect in a production deployment.  If the total resources requested by all apps and services is smaller than what the plan size allows then a production deployment will fail with an error.
```
ralt: "production deployment" = "master environment of a non-development plan"

This may need to be cleared up.

src/configuration/app/timezone.md
ralt: Logs are unaffected. This is only for _crons_, not for the system. The container will still run in UTC. Only the crons are affected by this key.

src/configuration/app/size.md
Crell: I'm thinking ahead here to when production doesn't imply master branch specifically.  Is there a better way to phrase it, keeping that in mind?

src/configuration/app/timezone.md
Crell: Oh.  Well, that's exciting.  I'll correct.

PR #576, opened 2017-07-03
Added a section on how to install Bash on Win10

martinpierrefrenette:
I have linked to the article I followed on Windows Central, and gave the explanations myself, using my words. I also added to do use the same username as I had that problem myself.

### DIFF
--- a/src/overview/cli.md
+++ b/src/overview/cli.md
@@ -90,9 +90,13 @@ Options:
 
 ## Installing the CLI on Windows 10
 
-To install the CLI on a Windows 10, 64-bit machine, BASH needs to be installed. Otherwise, you will need to install a Virtual Machine running Linux. The Windows 10 Anniversary Update is needed for support for git and apt/apt-get.
+There are multiple ways to connect to the CLI from Windows 10. Platform.sh recommends using BASH for Windows (Windows Subsystem for Linux).
 
-To install Bash you need to:
+## installating BASH on Windows 10
+
+You can install BASH to use the CLI on a Windows 10, 64-bit machine. The Windows 10 Anniversary Update is needed for support for git and apt/apt-get.
+
+To install Bash on Windows 10 Anniversary Edition you need to:
 
 1. Activate the Developer Mode in "Update & Security" in the Windows Settings. This will prompt you to restart your computer.
 2. Activate the Windows Subsystem for Linux (Beta) in "Turn Windows features on or off" in the Programs and Features section of the Windows Control Panel. Once again, you will need to restart your computer.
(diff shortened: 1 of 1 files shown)


src/overview/cli.md
```diff
@@ -87,3 +87,19 @@ Options:
...
+
+To install Bashm you need to:
+
+1 ) Activate the Developer Mode in "Update & Security" in the Windows Settings. This will prompt you to restart your computer.
+2 ) Activate the Windows Subsystem for Linux (Beta) in "Turn Windows features on or off" in the Programs and Features section of the Windows Control Panel. Once again, you will need to restart your computer.
+3 ) In the start menu, search for the program "bash.exe", which will prompt you to install it from thw Windows Store.
+
+Bash is now installed. 
+
+You can read more here: https://www.windowscentral.com/how-install-bash-shell-command-line-windows-10
```
Crell: Let's make this an actual link.  Like "read more on [Widows Central](http:....)

Crell: Without the spaces so it renders, which I had to put in so the comment didn't render it... :smile:

src/overview/cli.md
```diff
@@ -87,3 +87,19 @@ Options:
  --no (-n)             Answer "no" to all prompts
  --shell (-s)          Launch the shell
 ```
+
+## Installing the CLI on Windows 10
+
+To install the CLI on a Windows 10, 64-bit machine, BASH needs to be installed. Otherwise, you will need to install a Virtual Machine running Linux. The Windows 10 Anniversary Update is needed for support for git and apt/apt-get.
```
pjcdawkins: Theoretically, there are a lot of other options. The CLI could work with Git Bash, Cygwin, MinGW, and even (reportedly) PowerShell. So this should be phrased more like:

```
To install the CLI on Windows 10, we recommend using Bash.
```

and mention the Anniversary Update under the steps below

src/overview/cli.md
```diff
@@ -87,3 +87,19 @@ Options:
...
+
+To install Bashm you need to:
+
+1 ) Activate the Developer Mode in "Update & Security" in the Windows Settings. This will prompt you to restart your computer.
+2 ) Activate the Windows Subsystem for Linux (Beta) in "Turn Windows features on or off" in the Programs and Features section of the Windows Control Panel. Once again, you will need to restart your computer.
+3 ) In the start menu, search for the program "bash.exe", which will prompt you to install it from thw Windows Store.
```
Crell: This isn't rendering as a proper list.  I think you need to use:

1.
2.
3.

(It may work with parens, but there needs to not be a space.)

PR #583, opened 2017-07-12
Stick gitbook-cli version to 2.3.0 to prevent build failure

GuGuss:
The most recent gitbook-cli version 2.3.1 throws the following error during build:
`Error loading version latest: Error: Cannot find module 'extend'`


GuGuss:
See https://github.com/GitbookIO/gitbook-cli/issues/68

PR #584, opened 2017-07-12
Instructions to get Drush

MacWeber:


src/frameworks/drupal8/drush.md
```diff
@@ -10,11 +10,11 @@ commands at [Drush.org](http://www.drush.org). If you have never used
 Drush before, you can learn more about it on the [Drush GitHub
 Repository](https://github.com/drush-ops/drush#description)
 
-Platform.sh has Drush installed by default, so all environments can utilize Drush commands in the development process. You can use the CLI to set up Drush aliases, to easily run Drush commands on
-specific remote Platform.sh environments.
+
+Platform.sh Drupal templates have Drush installed automatically. All custom environments can utilize Drush commands in the development process if they have `drush` configured as a composer requirement and added it to the environment's `$PATH`. In addition, you can use the [Platform.sh CLI](/overview/cli.md) to set up Drush aliases easily for all of your project's environments.
```
pjcdawkins: All of this is mentioned later in the page. If you want to explain it in prose, and then in more detail later, then that's OK, but IMO the prose should include hyperlinks (to in-page anchors).

MacWeber: The sections in this doc are only "how-to" recipes, they do not explain the reasoning and it has been a cause of some customers opening tickets because it is not clear that only templates get Drush automatically. I am improving it for even better clarity in the next commit and adding the suggested links, too.

src/frameworks/drupal8/drush.md
```diff
@@ -10,10 +10,14 @@ commands at [Drush.org](http://www.drush.org). If you have never used
 Drush before, you can learn more about it on the [Drush GitHub
 Repository](https://github.com/drush-ops/drush#description)
 
-Platform.sh has Drush installed by default, so all environments can utilize Drush commands in the development process. You can use the CLI to set up Drush aliases, to easily run Drush commands on
-specific remote Platform.sh environments.
+- Platform.sh Drupal templates have Drush installed automatically.
```
Crell: Possessive on Platform.sh's.

src/frameworks/drupal8/drush.md
```diff
@@ -10,10 +10,14 @@ commands at [Drush.org](http://www.drush.org). If you have never used
...
+
+- All custom environments can utilize Drush commands in the development process if they have `drush` configured as a composer requirement and added it to the environment's `$PATH`.[¹](#install-drush-in-custom-projects)
```
Crell: Internal linking with the doc: Great.  But we don't use footnote-ish style anywhere else in the docs.  They're bad for accessibility.  Instead, make a part of the previous sentence the link or add a sentence like "See the section below on 
 (link) installing Drush in custom projects (/link)"

src/frameworks/drupal8/drush.md
```diff
@@ -10,10 +10,14 @@ commands at [Drush.org](http://www.drush.org). If you have never used
...
+
+- In addition, you can use the [Platform.sh CLI](/overview/cli.md) to set up Drush aliases easily for all of your project's environments.[²](#use-drush-aliases)
```
Crell: Same as above.

Crell:
Looks good overall, just some small things to clean up.

PR #191, opened 2015-09-25
Cli docs  - autogenerated cli docs.

OriPekelman:
Very hacky. I'll refactor later.

PR #590, opened 2017-07-20
Document available CLI commands and refactor CLI page.

GuGuss:
This is a follow up of the previous PR #191.

The CLI commands has been generated with: `platform list --format md`


Crell:
Is there any way we could make that regeneration process automatic?  We could probably pull in the CLI in a build hook and just run `platform list`.  That doesn't need to authenticate...

pjcdawkins:
There is. And the `list` command needs to be fixed first to account for aliases and hidden commands.

.platform.app.yaml
```diff
@@ -21,6 +21,10 @@ dependencies:
 # by default.
 hooks:
     build: |
+      # Generate the CLI command index page.
+      curl -sS https://platform.sh/cli/installer | php
+      platform list --format md > src/overview/cli-commands.md
```
Crell: This is generating an empty file, although because it happens in the build env I'm not entirely sure why.  This part works when I run `platform build` locally.  @pjcdawkins, please advise.

pjcdawkins: You'd need `source .environment` between those two lines: the installer adds `platform` to the PATH via the `.environment` file for new shells, but it can't change the PATH in the running shell.

Crell:
Hm.  Still producing an empty page. :disappointed: 

The other thing to note is that the output of the `platform list` command includes paths that are specific to the user running it.  I don't know the degree to which we care about that.  Or what it will look like once we get it running here, for that matter. :smile:

Crell:
Rebased on Master to resolve conflicts from the Getting Started guide.

GuGuss:
Looks good to me.

pjcdawkins:
Looks incredibly long and unreadable to me... we can do so much better. We could give the CLI another `--format` option (`gitbook-md`?) which produces a file with a table of contents, etc.

Crell:
I'm still skeptical of the value of including the full commands list in the docs to begin with.  To me the biggest value is having an example of running the CLI on the platform to reference. :smile: If we're going to include it, then yes a dedicated better output method sounds better.

Crell:
I talked with Patrick in Slack earlier.  We're both on board with just dropping the generated list entirely.  It doesn't offer any value over just running `platform list` yourself, and you'd need to install the CLI anyway before using anything the page would document.

gnufred:
@Crell @GuGuss @pjcdawkins I suggest a meeting in order to discuss how we are going forward with those changes so we can proceed. Unless someone can give us a definitive direction to take.

Crell:
OK, I'm making an executive decision.  Command list is gone.  Branch has been rebased.  If the remaining text is good, let's merge and move on with life.

PR #592, opened 2017-07-20
Update Solr config example

JGrubb:


Crell:
I think this may be Solr-version-dependent.  That's the problem.  @MacWeber please weigh in here.

PR #603, opened 2017-07-24
Add more PHP extensions.

Crell:
Because who doesn't love PHP extensions?

Only merge once the relevant containers have been released.


xtfer:
Looks good

PR #607, opened 2017-07-26
Fix snippet for PE

MacWeber:


GuGuss:
@MacWeber Enterprise has its own documentation.

PR #611, opened 2017-07-27
Node.js 8.2 is out.

GuGuss:


src/languages/nodejs.md
```diff
@@ -7,11 +7,9 @@ Javascript and PHP applications.
 
 ## Supported versions
 
-* 0.12
-* 4.7
 * 4.8
-* 6.9
-* 6.10
+* 6.11
+* 8.2
```
Crell: We actually removed the 6.10 image?  I thought we didn't do that...  Is no one using it?

Crell: In Slack, @GuGuss said we didn't actually remove 6.10.  Therefore we shouldn't remove it from the docs, either.  Let's restore it.

PR #614, opened 2017-07-28
Added clarifications that these are examples

martinpierrefrenette:
We have a client who actually used the DNS examples from the documentation instead of their own DNS entries. This brought their site down, as they thought our documentation was showing the actual DNS entries to put.

I added disclaimers to make it clear.

### DIFF
--- a/src/development/going-live.md
+++ b/src/development/going-live.md
@@ -111,18 +111,18 @@ First, configure your `routes.yaml` file like so:
   to: "https://www.{default}/"
 ```
 
-That will result in two domains being created on Platform.sh: `master-def456-abc123.us.platform.sh` and `www---master-def456-abc123.us.platform.sh`.  The former will automatically redirect to the latter.  In the `routes.yaml` file, `{default}` will automatically be replaced with `master-def456-abc123.us.platform.sh`.  In domain prefixes (like `www`), the `.` will be replaced with `---`.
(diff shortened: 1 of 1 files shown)


sruggier:
I think this change adds a lot of unnecessary visual noise, given that the document already clearly states that it's an example at the beginning of the section.

If the goal is to prevent people from taking their site down by putting bogus values in the DNS, I think a better way to reduce the chance of that would be to move away from using example values that could be mistakenly used, such as `master-def456-abc123.us.platform.sh`, and instead use placeholders, like `master-7rqtwti-<your project ID>.us.platform.sh`. That should reduce the likelihood that a user would use the example value without customizing it for their site.

xtfer:
I agree with @sruggier on this one. I saw the ticket and read the docs, and didn't think from reading the ENTIRE page that there was a problem with the example.

What it might need is a clearer set of basic instructions at the beginning of the page. eg. To set your DNS, do X, rather than a detailed discussion on how it all works.

Crell:
Related: I'm considering turning the Go-live page into its own section with a series of short pages rather than one long page.  We should fix the examples anyway as Simon suggests, but would that help or hinder the point @xtfer notes?

xtfer:
@Crell I think that would help. One of the issues with the current page is it reads as a series of steps. "Going live" would cover all those steps, but the instructions for each should probably work on their own.

Aside, it would be good to include instructions on the domain page in the UI, showing the actual CNAME to use...

Crell:
Closing in favor of #623

PR #621, opened 2017-08-07
Update cache.md

xtfer:
@Crell I found the HTTP cache documentation confusing. The order seemed counterintuitive, so I've taken a shot at making it clearer. If you find this to be useful, we can work towards merging it. Otherwise please just close.


src/configuration/routes/cache.md
```diff
@@ -85,96 +91,79 @@ cache:
...
+* Responses with the `Cache-Control` header set to `Private`, `No-Cache`, or `No-Store` are not cached. 
+* Responses with the `Set-Cookie` header set are also not cached.
 
-## Cache duration
+### Cache duration
 
-The cache duration is decided based on the `Cache-Control` response
-header value. If no `Cache-Control` header is in the response, then the
-value of `default_ttl` key is used.
+The cache duration is decided based on the `Cache-Control` response header value. If no `Cache-Control` header is in the response, then the value of `default_ttl` key is used.
```
Crell: We should mention somewhere that there is no flush-all command for the router.  Is this a good place for it?  If not, what would be?

src/configuration/routes/cache.md
```diff
@@ -85,96 +91,79 @@ cache:
...
 
-## Cache serving
+### Cache serving
 
-Our web server does not honor the `Pragma` request header.
-Conditional requests using `If-Modified-Since` and `If-None-Match`
-are both supported.
+Our web server does not honor the `Pragma` request header. Conditional requests using `If-Modified-Since` and `If-None-Match` are both supported.
```
Crell: Let's reverse these 2 sentences.  Reading quickly it makes it sound like we don't support any of the 3, which is not the case.

src/configuration/routes/cache.md
```diff
@@ -85,96 +91,79 @@ cache:
...
-in the response, the web server would send a request to your
-application with `If-Modified-Since` header. Also, `If-None-Match` header
-is sent in the conditional request when `Etag` header is set in the cached
-response. Your application can extend the validity of the cache by replying
-`HTTP 304 Not Modified`.
+When the cache is expired, indicated by `Last-Modified` header in the response, the web server would send a request to your
+application with `If-Modified-Since` header. Also, `If-None-Match` header is sent in the conditional request when `Etag` header is set in the cached response. Your application can extend the validity of the cache by replying `HTTP 304 Not Modified`.
 
-## Cache Attributes
+## Caching strategies
```
Crell: This section seems more like "advanced caching use cases" or something like that.  Although it has only one item, so the plural seems a bit out of place until we add more...

src/configuration/routes/cache.md
```diff
@@ -85,96 +91,79 @@ cache:
...
 
-### `enabled`
+### Cache per route
 
-When set to `true`, enable the cache for this route. When set to
-`false`, disable the cache for this route.
+If you need fine-grained caching, you can set up caching rules for several routes separately.
 
-### `headers`
+Here is an example:
```
Crell: "For example" rather than "Here is an example".

src/configuration/routes/cache.md
```diff
@@ -25,56 +23,64 @@ https://{default}/:
...
+#### `default_ttl`
 
-## Cache key
+If the response does not have a `Cache-Control` header, the `default_ttl` key is used to define the cache duration, in seconds. The default value is `0`.
 
-To decide how to cache a response, Platform.sh will build a cache key
-depending on several factors and store the response associated with this
-key. When a request comes with the same cache key, the cached response will
-be reused.
+If the application code returns a Cache-Control header, or if your `.platform.app.yaml` file is configured to set a cache lifetime for static assets, then this value is ignored in favor of the application container's headers.
```
pjcdawkins: Also, the `default_ttl` only applies to non-static responses, I believe - all static assets will have a Cache-Control header anyway with a max age defaulting to 0 (the default for `expires` in the web config).

And "application container's headers" => "application's headers"

src/configuration/routes/cache.md
```diff
@@ -25,56 +23,64 @@ https://{default}/:
 
 In this example, requests will be cached based on the URI, the `Accept` header, `Accept-Language` header, and `X-Language-Locale` header; Any response that lacks a `Cache-Control` header will be cached for 60 seconds; and the presence of any cookie in the response will disable caching of that response.
```
Crell: This should also note that any header listed in Vary will be part of the cache key, too.

src/configuration/routes/cache.md
```diff
@@ -1,16 +1,35 @@
-# HTTP Cache
+# HTTP cache
 
-## Rationale
+Platform.sh supports HTTP caching at the server level. The cache is only applied to `GET` and `HEAD` requests.
 
-Many hosting solutions are adding an additional layer (Varnish...) to handle caching. Platform.sh supports HTTP caching right at the web server level.
+The cache must be explicitly enabled using the `cache` key in your `.platform/routes.yaml` file. Platform.sh serves the files or requests generated by the application directly, unless the cache is enabled.
```
pjcdawkins: Actually the cache is enabled by default (if the `cache` key does not exist for the route), with the following settings:

```yaml
cache:
    enabled: true
    default_ttl: 0
    cookies: ['*']
    headers: ['Accept', 'Accept-Language']
```

src/configuration/routes/cache.md
```diff
@@ -1,16 +1,35 @@
...
-By default, the router will respect whatever cache headers are sent by the application (either from your application code or from the `.platform.app.yaml` configuration), but the presence of any cookie headers will cause the cache to be bypassed.
+To decide how to cache a response, Platform.sh will build a cache key depending on several factors and store the response associated with this key. When a request comes with the same cache key, the cached response will be reused.
 
-If you disable caching, Platform.sh serves the files or requests generated by the application directly.
+If you enable caching...
 
-## Reference
+* you can configure cache behaviour for different location blocks in your `.platform.app.yaml`
+* the router will respect whatever cache headers are sent by the application
+* cookies will not bypass the cache unless the `cookie` property is set or the `Set-Cookie` header is present
```
pjcdawkins: I don't understand this point. Suggest:

* the cache key will vary based on the cookies that are configured in `cookies`

src/configuration/routes/cache.md
```diff
@@ -1,16 +1,35 @@
...
+* responses with the `Cache-Control` header set to `Private`, `No-Cache`, or `No-Store` are not cached. 
 
-Cache can be enabled in your `.platform/routes.yaml` file like below:
+The Platform.sh HTTP cache is an implementation of the Nginx HTTP cache.
```
pjcdawkins: I don't think it's helpful to mention this (and it might change).

src/configuration/routes/cache.md
```diff
@@ -23,58 +42,80 @@ https://{default}/:
...
+Turns the cache on or off for a route.
 
-If you need fine-grained caching, you can set up caching rules for
-several routes separately.
+**Type:** Boolean
 
-Here is an example:
+**Values**
+* `true`: enable the cache for this route
+* `false`: disable the cache for this route [default]
```
pjcdawkins: There is no default here: `enabled` is a required key if you have `cache` (and if you don't have `cache`, the cache will be enabled)

src/configuration/routes/cache.md
```diff
@@ -23,58 +42,80 @@ https://{default}/:
...
+
+### `headers`
+
+Adds specific header fields to the cache key, enabling caching of separate responses for those headers.
+
+For example, if the `headers` key is the following, Platform.sh will cache a different response for each value of the `Accept` HTTP request header:
```
pjcdawkins: The default is `["Accept", "Accept-Language"]` so maybe pick a different example

src/configuration/routes/cache.md
```diff
@@ -23,58 +42,80 @@ https://{default}/:
...
+The cache key will depend on the value of the `foo` cookie in the request.
 
-With this configuration, the following routes are cached:
+**Type:** List
 
--   `https://{default}/`
--   `https://{default}/foo/bar/`
--   `https://{default}/foo/bar/baz/`
+**Values:**
+* `*`: any request with a cookie will bypass the cache [default]
```
pjcdawkins: it's a list, so I suggest notating as `["*"]` or `['*']` (it only has an effect if it's the only item in the list)

src/configuration/routes/cache.md
```diff
@@ -23,58 +42,80 @@ https://{default}/:
...
 
-## Cache key
+**Type:** integer
 
-To decide how to cache a response, Platform.sh will build a cache key
-depending on several factors and store the response associated with this
-key. When a request comes with the same cache key, the cached response will
-be reused.
+**Values:**
+* `0`: Cache indefinitely [default]
```
Crell: A max-age of 0 is is don't cache, not cache indefinitely.  Unless we special case reverse that, too, which we really shouldn't. :smile:

src/configuration/routes/cache.md
```diff
@@ -85,96 +126,98 @@ cache:
...
-## Cache Attributes
+When the cache is expired, indicated by `Last-Modified` header in the response, the web server would send a request to your
+application with `If-Modified-Since` header. Also, `If-None-Match` header is sent in the conditional request when `Etag` header is set in the cached response, your application can extend the validity of the cache by replying `HTTP 304 Not Modified`.
 
-### `enabled`
+### Flushing
 
-When set to `true`, enable the cache for this route. When set to
-`false`, disable the cache for this route.
+The HTTP cache does not support a complete cache flush, however, you can invalidate the cache by setting `cache:false`.
```
Crell: Ooo, clever.  Just toggle it off and on again. :smile:

src/configuration/routes/cache.md
```diff
@@ -85,96 +126,98 @@ cache:
...
+
+Set the Vary header to `X-Forwarded-Proto` [custom request header](/development/faq.md#do-you-add-custom-http-headers) to render content based on the request protocol (i.e. HTTP or HTTPS). By adding `Vary: X-Forwarded-Proto` to the response header, HTTP and HTTPS content would be cached separately.
 
-If the response does not have a `Cache-Control` header, the
-`default_ttl` key is used to define the cache duration, in seconds. The
-default value is `0`.
+### Cache zipped content separately
 
-If the application code returns a Cache-Control header, or if your `.platform.app.yaml` file is configured to set a cache lifetime for static assets, then this value is ignored in favor of the application container's headers.
+Use `Vary: Accept-Encoding` to serve different content depending on the encoding. Useful for ensuring that gzipped content is not served to clients that can read it.
```
Crell: Shouldn't there be another negative in there?  We don't want to make sure non-gzip clients only get gzip content, I assume...

src/configuration/routes/cache.md
xtfer: Woah! That's quite a big difference. Good catch. Will update.

src/configuration/routes/cache.md
xtfer: Fixed.

src/configuration/routes/cache.md
xtfer: Changed to `Platform.sh will cache a different response for each value of the `Accept` HTTP request header "only"`.

src/configuration/routes/cache.md
xtfer: Updated

src/configuration/routes/cache.md
xtfer: Fixed

src/configuration/routes/cache.md
xtfer: Fixed

Crell:
Patrick, I'm happy here as soon as you are.  You have an outstanding review listed, so as soon as you hit green on that we're good to go.  Feel free to merge.

PR #623, opened 2017-08-07
Break up Go Live page into its own section

OriPekelman:


OriPekelman:
Possibly going with something like https://www.npmjs.com/package/gitbook-plugin-page-toc will be  much nicer. But we can do this later.

Crell:
I think we tried that TOC plugin and it was a total fail.  There's another we're already using currently on a few pages.  I like the breakup, though.

src/golive/README.md
```diff
@@ -0,0 +1,8 @@
+# Going Live
+
+You successfully deployed your project to Platform.sh? Time to go live. This is a simple process, but it's worth reading the docs the first time around.
+
+* [Pre-Launch Checklist](/golive/checklist.md) 
+* [Going Live Steps](/golive/steps.md) 
+* [Troubleshooting](/golive/troubleshoot.md) 
+* [Example](/golive/example.md)
```
Crell: This file is never seen or linked from.  It also doesn't really offer anything, so I will just remove it.

xtfer:
References to `<environment>-<project>.<region>.platform.sh` are confusing... You can't construct these backward from those parts, due to an additional component in the project. The real format is more like `<environment>-<?>-<project>.<region>.platform.site` (note `site` on the end too).

Crell:
Well, technically `<environment>` is correct.  `master-7rqtwti` is the environment name that corresponds to branch `master`.  Yes, it's confusing, but that's how the system works at the moment.  I'm struggling with how to describe those generally across the docs.

As for platform.sh vs platformsh.site, bah, I didn't even notice that.  And having the domain name different like that on different regions... are we trying to make it harder to document? :smile: I guess for now we should be consistent on the old style, until we switch over to all new-style regions then shift the docs.  Ugh.

xtfer:
Perhaps <branch>-<slug>-<project>...

Crell:
Let's leave the URL pattern for later; this PR is already an improvement.  Any other issues here before we merge?

PR #626, opened 2017-08-10
Algolia search option

OriPekelman:
This seems to work better than addsearch. I would go for Aloglia. Needs to be implemented properly do not merge.

### DIFF
--- a/book.json
+++ b/book.json
@@ -13,8 +13,7 @@
     "toc",
     "-sharing",
     "-lunr",
-    "-search",
-    "search-plus"
+    "-search"
   ],
   "pluginsConfig": {
     "edit-link": {
--- a/src/_layouts/layout.html
+++ b/src/_layouts/layout.html
@@ -23,6 +23,7 @@
 <script src="/scripts/collapsible-menu.js"></script>
 <!-- at the end of the BODY -->
 <script type="text/javascript" src="https://cdn.jsdelivr.net/docsearch.js/2/docsearch.min.js"></script>
+<script src="/scripts/search.js"></script>
 <!-- at the end of the HEAD -->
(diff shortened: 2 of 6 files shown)


Crell:
image

Neither Troubleshooting nor PHP Support are under "The Big Picture", and Redis is not under MySQL.

If those sections were accurate that would be a pretty nice search engine, but it's rather misleading as is.

maxiloc:
I updated the config, should be fixed now.

Crell:
I spent some time looking into running the indexer ourselves.  Unfortunately it depends on a couple of additional C packages that are not installed on Psh containers, and adding the headers for them is more work than I want to put into the docs for right now.  Which is a bit unfortunate as it seems that would allow us to have at least basic analytics for docs search, which would be nice.

In any case, turning this over to Aaron for now to style.  Aaron, let's just try to swap the current search box out for the Algolia one.  Nothing fancy needed.

maxiloc:
We can provide analytics for the index on the DocSearch account. @OriPekelman, you can send me the list of emails that need access, and I'll invite them.

dudenhofer:
Updated the styles and removed the old search plugin we were using. Seems to be working just fine to me.

OriPekelman:
This should also be test _after_ merging... remember the google custom search bug... As far as I concerned can be merged

Crell:
Everything checks out for me in spot checks on desktop.  However, I just checked it on mobile for the first time (I should have checked it earlier) and the pop-over is not resizing for mobile, making it very difficult to use.  Is that something that could be fixed on our end or is that on Algolia to fix on their end?  

Paging @maxiloc for input on that front...

Crell:
Much nicer.  Thanks, Aaron!

PR #627, opened 2017-08-11
Document Composer Manager for Drupal 7.

Crell:
This needs to be vetted to make sure I didn't miss a step somewhere.  Referring to @pjcdawkins.

### DIFF
--- a/src/frameworks/drupal7/composer-manager.md
+++ b/src/frameworks/drupal7/composer-manager.md
@@ -22,23 +22,19 @@ Composer Manager works by using a Drush command to aggregate all module-provided
 
 ```php
 $conf['composer_manager_vendor_dir'] = 'composer/vendor';
-$conf['composer_manager_file_dir'] = 'composer/';
+$conf['composer_manager_file_dir'] = 'composer';
 ```
 
 The above lines will direct Composer Manager to put the generated `composer.json` file into `composer/`, and to autoload Composer-based packages from the `composer/vendor` directory.  The paths are relative to the Drupal root.  You may use another location if desired (such as `../composer`), provided that they are not in the file mount and the `vendor` directory is a sibling of wherever the `composer.json` file will be.
 
-Then, manually create the `composer` and `composer/vendor` directories and place an empty file named `.gitkeep` into the vendor directory, then commit it to git.  Composer Manager needs that directory to exist but due to a bug will not create it itself.  The file name technically does not matter but `.gitkeep` is a very common convention.
-
-## 3.  Configure `.gitignore`
-
-If you have a `.gitignore` file, The vendor directory should be added to it.  However, due to the `.gitkeep` file it is a bit trickier than just listing the directory name.  Adding the following two lines (adjusting for the path to the `composer.json` file as needed) will have the desired result, once `.gitkeep` is added to the repository:
+Then, manually create the `composer` directory and place a `.gitignore` file inside it, containing the following, and commit it to Git:
 
 ```
-composer/vendor/*/
-composer/vendor/autoload.php
+# Exclude Composer dependencies.
+/vendor
 ```
 
-## 4. Update the build hook
+## 3. Update the build hook
 
 If the `composer.json` file is in the application root (as a sibling of the `.platform.app.yaml` file) then it's simplest to just use the `composer` build flavor and let it install Composer dependencies:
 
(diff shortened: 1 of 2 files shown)


src/frameworks/drupal7/composer-manager.md
```diff
@@ -0,0 +1,67 @@
...
+projects[composer_manager][version] = "1.8"
+projects[composer_manager][patch][] = "https://www.drupal.org/files/issues/composer_manager-2620348-3.patch"
+projects[composer_manager][patch][] = "https://www.drupal.org/files/issues/composer_manager-relative_realpath-2864297-2.patch"
+```
+
+If you're checking the entire codebase into Git, do so with Composer Manager as well.  Then download and apply the two patches in the issues above and commit the result.
+
+## 2. Configure file locations
+
+Composer Manager works by using a Drush command to aggregate all module-provided `composer.json` files into a single file, which can then be installed via a normal Composer command.  Both the generated file and the resulting `vendor` directory must be in the application portion of the file system, that is, not in a writeable file mount.  As that is not the default configuration for Composer Manager it will need to be changed.  Add the following lines to your `settings.php` file (not `settings.platformsh.php`):
```
pjcdawkins: It could be in either `settings.php` or `settings.platformsh.php` really, so it's probably simpler just to say `settings.php` here

src/frameworks/drupal7/composer-manager.md
```diff
@@ -22,23 +22,19 @@ Composer Manager works by using a Drush command to aggregate all module-provided
...
-## 3.  Configure `.gitignore`
-
-If you have a `.gitignore` file, The vendor directory should be added to it.  However, due to the `.gitkeep` file it is a bit trickier than just listing the directory name.  Adding the following two lines (adjusting for the path to the `composer.json` file as needed) will have the desired result, once `.gitkeep` is added to the repository:
+Then, manually create the `composer` directory and place a `.gitignore` file inside it, containing the following, and commit it to Git:
 
 ```
-composer/vendor/*/
-composer/vendor/autoload.php
+# Exclude Composer dependencies.
+/vendor
```
Crell: This approach won't create a vendor directory, which in my testing needs to pre-exist.  Composer Manager won't create it. (This seems like a bug in Composer Manager, but I didn't dig too deeply into it.)

src/frameworks/drupal7/composer-manager.md
```diff
@@ -36,28 +36,34 @@ Then, manually create the `composer` directory and place a `.gitignore` file ins
 
 ## 3. Update the build hook
 
-If the `composer.json` file is in the application root (as a sibling of the `.platform.app.yaml` file) then it's simplest to just use the `composer` build flavor and let it install Composer dependencies:
+Create a build hook in your `.platform.app.yaml` file that will install Composer dependencies:
 
 ```yaml
-build:
-   flavor: composer
+hooks:
```
Crell: Why not mention the composer flavor as an option?

pjcdawkins: I can't think of a situation where one would put the `composer.json` at the root of the repository, so I thought it was an unnecessary distraction.

src/frameworks/drupal7/composer-manager.md
pjcdawkins: The vendor directory will be created as soon as you run `composer install` - does it need to exist before that?

pjcdawkins: ... ah you're right, it does. Humph

src/frameworks/drupal7/composer-manager.md
```diff
@@ -0,0 +1,69 @@
...
+The above lines will direct Composer Manager to put the generated `composer.json` file into `../composer/`, and to autoload Composer-based packages from the `../composer/vendor` directory.  The paths are relative to the Drupal root.  You may use another location if desired provided that they are not in a writable file mount, and provided the `vendor` directory is a sibling of wherever the `composer.json` file will be.
+
+Then, manually create the `composer` directory and place a `.gitignore` file inside it, containing the following, and commit it to Git:
+
+```
+# Exclude Composer dependencies.
+/vendor
+```
+
+## 3. Update the build hook
```
Crell: Let's add a note here about adding a local `vendor` directory, at least, and note that it's caused by an upstream bug.

pjcdawkins: Resolved in d55060b

PR #633, opened 2017-08-17
Add legacy_schema to importer endpoint

JGrubb:


Crell:
We should included a mention in the text that the CLI et al won't work if you don't list it. A few paragraphs down we say:

```
If a given endpoint has access to multiple databases you should also specify which will be listed by default in the relationships array. If one isn't specified the path property of the relationship will be null. If there is only one database listed it will be used and default_schema is unnecessary.
```

Which we've just determined is not true. :smile:

JGrubb:
The trick is, I'm not sure it's just the CLI that wouldn't work.  The issue the other day was with a Drupal project and drush was unable to properly connect to the DB using `drush sqlc --database=legacydb` even though the relationship information was all there.  When connecting via `drush sqlc` it would open a connection to the DB server, but would connect to `SCHEMA_INFORMATION` database, instead of the intended DB.

Is this a bug/regression?  I don't know.

PR #634, opened 2017-08-17
Overhaul local development page.

Crell:


src/development/local.md
```diff
@@ -121,3 +116,10 @@ and you can close tunnels with:
 ```bash
 platform tunnel:close
 ```
+
+## Untethered Local
+
+Alternatively, you can also run the entire site locally on your computer.  That is more performant as there's no lag time to connect to a remote database and doesn't require an active Internet connection to work.  However, it does require running all necessary services (databases, search servers, etc.) locally.  These can be setup however you prefer, although Platform.sh recommends using a virtual machine to make it easier to share configuration between developers.
```
Fidelix: "lag time"
"extra latency" or "latency"

src/development/local.md
```diff
@@ -1,13 +1,14 @@
...
+You must have an [SSH key](/overview/tools.md#ssh) already configured on your account, and have both [Git](/overview/tools.md#git) and the [Platform.sh CLI](/overview/cli.md) installed before continuing.
 
-In order to copy your project locally, you'll need to know its project
-ID. The command `platform projects` will list all of the projects in
-your account.
+## Download the code
+
+If you don't already have a local Git clone of your code, you can easily download it.  From the Platform.sh UI copy the `platform get` command listed just under the action buttons on the right and run it locally on your system.
+
+Alternatively, you can use the CLI to get a copy of the code directly.  Run `platform projects` to list all of the projects in your account.
```
pjcdawkins: IMO referring to the UI at this point adds a lot of extra thinking steps (1. understand what UI means, 2. find the UI URL for a particular project, 3. go to it, 4. maybe log in, 5. understand where the action buttons are, 6. copy and paste). Typing `platform get` is quicker... so I suggest:

> If you don't already have a local copy of your project's code, run `platform get` to download one. You can also run `platform projects` to list all of the projects in your account.

src/development/local.md
```diff
@@ -95,15 +90,16 @@ Close tunnels with: platform tunnel:close
...
 
-Now you can connect to the remote database normally, as if it were
-local.
+Now you can connect to the remote database normally, as if it were local.
 
 ```
 $ mysql --host=127.0.0.1 --port=30001 --user='user' --password='' --database='main'
 ```
 
+The specific port that each service uses may vary project to project, but it will not change within a single project unless you change your services or relationship definition.  That makes it safe to add a local-configuration file for your application that connects to, in this case, `localhost:30001` for the SQL database and `localhost:30000` for Redis.
```
pjcdawkins: The local port may change at any time, it just depends whether it's already occupied by something else.

To stay up to date you could use the result of `platform tunnel:info --encode` as PLATFORM_RELATIONSHIPS, in PHP maybe:

```php
if ($relationships_encoded = shell_exec('platform tunnel:info --encode')) {
  $relationships = json_decode(base64_decode($relationships_encoded, TRUE);
  // ...
}
```

or run the server with `export PLATFORM_RELATIONSHIPS="$(platform tunnel:info --encode)"`

or something similar

pjcdawkins:
I'm not convinced that SSH tunnelling is more often easier than setting up local services. I think tunnelling is far too slow with Drupal for example. Your experience may vary, but I'd suggest presenting the two options as equals (equally difficult...) with their own pros and cons.

src/development/local.md
danmilon: I agree. I was confused when I read that sentence.

Crell:
@pjcdawkins Tethered is simpler, which doesn't mean easier. :smile: Just that it has fewer steps.  How would you suggest rewording it to nudge people toward using Untethered?

PR #635, opened 2017-08-17
Getting started guide

Crell:
Reorganizing some pages, adding a few more, creating a new and improved getting started guide.


OriPekelman:
I could not get markdown in table to work. There is an old plugin for that. Does not seem to function. You could always render that part as HTML and include it as such.

Crell:
Yeah, I just did the HTML manually for the table.  Seemed to work well enough.  I'm open to other formatting approaches as well, but figured just a bunch of lists would look lame.

src/gettingstarted/first-project.md
```diff
@@ -0,0 +1,91 @@
+# Your first project
```
pjcdawkins: IMO this is too verbose, without enough breaks or reference points.

Perhaps there is some unnecessary detail. Regardless, I think adding hyperlinks from each concept within the prose (to their relevant detailed docs) would break it up for the eye.

src/gettingstarted/tools.md
```diff
@@ -0,0 +1,49 @@
...
+
+> **see also**
+>
+> * [Install Git](https://help.github.com/articles/set-up-git/)
+> * [Learn more about Git](https://git-scm.com/)
+
+
+## SSH
+
+One of the ways [Platform.sh](https://platform.sh/) keeps things secure is by using SSH behind the scenes. Users connect to their Platform.sh Git repository and to their Command Line Interface (CLI) using SSH.
```
pjcdawkins: The CLI (`platform login`) mainly authenticates with OAuth2: it only passes through SSH when it's connecting to your Git or app servers. Suggest:

> Users connect to their Platform.sh Git repository and to their applications and services using SSH.

src/gettingstarted/first-project.md
Crell: Fixed by converting it to a shorter set of bullet points.

PR #637, opened 2017-08-24
Misleading --fetch-branches value in GitHub integration documentation

yannickoo:
In order to avoid confusion for developers we should update the documentation and set `--fetch-branches` to it's default value so copying the full line doesn't lead to unexpected behavior.


Crell:
Yep, makes sense to me.  Thanks.

PR #640, opened 2017-09-01
Update cache.md

danharper83:
Added extra detail on sentence to make it clear what adding the cookies to the array does.


pjcdawkins:
Thank you!

PR #644, opened 2017-09-18
Document pre/post-merge PR toggle and data cloning toggle for GitHub

Crell:
Also did some formatting cleanup while I was in there.


pjcdawkins:
> Still waiting for confirmation that it works on BitBucket, too.

It doesn't, it's a GitHub-only option, right now.

Bitbucket PRs are built post-merge via a platform.sh merge process (Bitbucket itself doesn't expose a post-merged commit).

Crell:
@pjcdawkins We should prioritize adding this to the BitBucket integration then for parity.  We have customer requests for it.

Crell:
I added another new feature here as well, since it's a one line change to the same file. It's in its own commit at least. :smile:

PR #649, opened 2017-09-20
Document how compression is handled

Crell:
Since we now do some compression on the fly automatically.


src/configuration/app-containers.md
```diff
@@ -40,6 +40,18 @@ The `.platform.app.yaml` file needs at least one of the following to define an i
...
+
+Platform.sh does not compress any dynamic responses generated by your application due to a [well known security issue](https://en.wikipedia.org/wiki/BREACH_%28security_exploit%29).  While your application can compress its own response, doing so when the response includes any user-specific information, including a session cookie, opens up an attack vector over SSL connections.  For that reason we recommend against compressing any generated responses.
+
+Requests for static files on disk are compressed automatically using either gzip or brotli compression if:
+
+* The request headers for the file support gzip or brotli.
+* The file would be served with a cache expiration time in the future.
+* The file type is one of: html, javascript, json, pdf, postscript, svg, css, csv, plain text, or XML.
+
+Additionally, if a file with a ".gz" or ".bl" extension exists that will be served instead for the appropriate compression type regardless of the file type.  That is, a request for `styles.css` that accepts a gzipped file (according to the request headers) will automatically return the contents of `styles.css.gz` if it exists.
```
ralt: `s/\.bl/\.br/`

ralt: Maybe mention that this is a CPU optimization?

src/configuration/app-containers.md
```diff
@@ -40,6 +40,18 @@ The `.platform.app.yaml` file needs at least one of the following to define an i
 
 Each web or worker instance is its own running container, which takes its own resources.  The `size` key allows some control over how many resources each container gets and if omitted the system will select one of a few fixed sizes for each container automatically.  All application and service containers are given resources out of a common pool defined by your plan size.  That means the more containers you define, the fewer resources each one will get and you may need to increase your plan size.
 
+## Compression
+
+Platform.sh does not compress any dynamic responses generated by your application due to a [well known security issue](https://en.wikipedia.org/wiki/BREACH_%28security_exploit%29).  While your application can compress its own response, doing so when the response includes any user-specific information, including a session cookie, opens up an attack vector over SSL connections.  For that reason we recommend against compressing any generated responses.
```
ralt: `s/SSL/TLS/`?

src/configuration/app-containers.md
```diff
@@ -40,6 +40,18 @@ The `.platform.app.yaml` file needs at least one of the following to define an i
...
+
+Requests for static files on disk are compressed automatically using either gzip or brotli compression if:
```
ralt: Maybe note that this depends on the `script` and `passthru` properties in our `web.locations`?

Crell: Elaborate?  Not quite sure I follow here.

ralt: Technically speaking, you can serve "static files on disk" through your application. We only auto compress when the locations rules of your `.platform.app.yaml` respect a few constraints.

src/configuration/app-containers.md
Crell: I'll put both.  TLS may be more accurate but most people only know "SSL is that secure thingie that makes my connection safe, right?"

src/configuration/app-containers.md
Crell: Oh, I see.  I'll adjust the wording.

PR #657, opened 2017-10-02
Add note for manually renewing Let's Encrypt certificates for projects that are not regularly updated.

barryrld:
I came across an issue today where if a project is dormant for 3 months then the Let's Encrypt certificate will expire because the renewal attempt process is run which the environment is rebuilt.

This PR is my proposal to update the docs to reflect this caveat. Sorry for the extra line endings that have come out in the diff. I thought that you may want to alter the wording anyhow, so perhaps this should have been an issue (for further discussion) instead of a PR. If so, please let me know and I can convert this to an issue instead.

The key change to note in `src/configuration/routes.md`:

> **note**
> Let's Encrypt certificate renewals are attempted each time your environment is deployed. If your project does not receive regular code commits, you will need to manually issue a re-deployment to ensure the certificate remains valid. We suggest that you do when your project doesn't receive any updates for over 1 month. This can be done by pushing a code change via git or issuing the following command from your **local** environment : `platform variable:set -W -y force-le-renewal true && platform variable:delete -W -y force-le-renewal`

UPDATE: I also wasn't sure about the inclusion of the workaround command, but it's here now for discussion.

Thanks,

Barry

### DIFF
--- a/src/configuration/routes.md
+++ b/src/configuration/routes.md
@@ -31,7 +31,7 @@ routes will be resolved to `http://www.example.com/` and
 `https://example.com/blog` in the master environment.
 
 Platform.sh will also generate a domain for every active development environment.
-It will receive a domain name based on the region, project ID, branch name, and 
+It will receive a domain name based on the region, project ID, branch name, and
 a per-environment random string. The domain name itself is not guaranteed stable,
 although the pattern is consistent.
 
@@ -57,7 +57,7 @@ Each route can be configured separately. It has the following properties
   For more information: see [SSI](/configuration/routes/ssi.html).
 * `redirects` controls [redirect rules](/configuration/routes/redirects.html) associated with the
   route.
-  
+
 ![Routes files](/images/config_diagrams/routes2.svg)
 
 > **note**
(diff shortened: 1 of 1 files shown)


Crell:
Good call, thanks Barry!

I like the suggested workaround, but can you include an explanation of what it does?  What it's actually doing is fairly trivial, but technically causes two deploys, I believe.  It's worth explaining what the command does so that it's not presented as just a magic incantation.

Crell:
Also, grammar tweak: "We suggest that you do when your project" - Change to "you do so when..."  (Without the "so" it's missing what you should do.)

The whitespace fixes are fine; I need to smack my IDE for not being more aggressive in cleaning those out to begin with.

barryrld:
Thanks for the review Larry!

I've updated the grammar tweak you suggested. I've also addressed the double-build issue. On reflection, it would be better to set the date and time when `variable-set` is called. This offers the benefit of issuing only one rebuild and the ability to inspect the rebuild time from the variable itself. The only possible perceived disadvantage is that the variable persists.

One last thing to mention is that the command is:

`NOW=$(date +"%F_%H:%M:%S") && platform variable:set -W -y force-le-renewal $NOW`

This is explained in the amended note as to what it does, but I should mention that I've only been able to test this on my Arch Linux machine and my brother's Macbook - i.e. I don't know about those poor souls using Windows. I know the date command is fairly prevalent on *nix type systems, but it feels like a possible weakness with this solution. Perhaps, I'm being overly cautious. Do you think it's worth including this caveat e.g. YMMV. Understandably, we probably don't want to over-populate the docs too much.

What do you think?

xtfer:
Looks good to me. I suggest we push the update and wait for someone who uses Windows to discover it doesn't work. That won't break their build or anything serious, so then it can be fixed up.

Crell:
Makes sense to me.  Thanks, @barryrld and @xtfer!

PR #662, opened 2017-10-11
Add page on accessing the site over SFTP/SSH/Rsync

Crell:
Open question: Should we recommend any particular SFTP client on Windows/Mac, or just let people Google rather than getting into that question?


xtfer:
Looks good. Merging.

PR #665, opened 2017-10-16
New TLS features

Crell:
Coming in Git 2.7.  Getting it out there ahead of time for review.

### DIFF
--- a/src/configuration/routes/https.md
+++ b/src/configuration/routes/https.md
@@ -100,22 +100,28 @@ Note: If multiple routes for the same domain specify different HSTS settings, th
 
 In some non-browser applications (such as mobile applications, IoT devices, or other restricted-client-list use cases), it is beneficial to restrict access to selected devices using TLS.  This process is known as client-authenticated TLS, and functions effectively as a more secure alternative to HTTP Basic Auth.
 
-Platform.sh allows you to set the server-side portion of such a configuration via the `routes.yaml` file.  Doing so entails setting client authentication to required, and then providing a signed certificate either inline or via a file reference.
-
-```yaml
-tls:
-    client_authentication: "require"
-    client_certificate_authorities: !file cert.key
-```
-
-In this case, `cert.key` is resolved relative to the `.platform` directory.  Alternatively, the key can be specified inline in the file:
+By default, any valid SSL cert issued by one of the common certificate issuing authorities will be accepted.  Alternatively, you can restrict access to SSL certs issued by just those certificate authorities you specify, including a custom authority.  (The latter is generally only applicable if you are building a mass-market IoT device or similar.)  To do so, set `client_authentication` required and then provide a list of the certificates of the certificate authorities you wish to allow.
 
 ```yaml
 tls:
     client_authentication: "require"
     client_certificate_authorities:
-    - |
-        -----BEGIN CERTIFICATE-----
-        ### Several lines of random characters here###
-        -----END CERTIFICATE-----
+        - !file cert1.key
+        - !file cert2.key
+```
+
+In this case, the key files are resolved relative to the `.platform` directory.  Alternatively, the keys can be specified inline in the file:
+
+```yaml
+tls:
+    client_authentication: "require"
+    client_certificate_authorities:
+        - |
+            -----BEGIN CERTIFICATE-----
+            ### Several lines of random characters here ###
+            -----END CERTIFICATE-----
+        - |
+            -----BEGIN CERTIFICATE-----
+            ### Several lines of different random characters here ###
+            -----END CERTIFICATE-----
 ```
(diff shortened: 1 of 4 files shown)


src/configuration/routes/https.md
```diff
@@ -0,0 +1,117 @@
...
+        # ...
+```
+
+### `min_version`
+
+Setting a minimum version of TLS will cause the server to automatically reject any connections using an older version of TLS.  While the vast majority of modern browsers will default to TLS 1.2, there are some older browsers that still use insecure older versions of TLS.  Rejecting older versions with known security vulnerabilities is necessary for some security compliance processes.
+
+```yaml
+tls:
+    min_version: TLSv1.2
```
ralt: Valid values are `TLSv1.0`, `TLSv1.1` and `TLSv1.2`.

src/configuration/routes/https.md
```diff
@@ -0,0 +1,117 @@
...
+tls:
+    strict_transport_security:
+        enabled: true
+        include_subdomains: true
+        preload: true
+```
+
+There are three sub-properties for the `strict_transport_security` property:
+
+* `enabled`: Can be `true` or `false`.  Defaults to `false`.  If `false`, the other properties wil be ignored.
```
ralt: It can be `true`, `false`, or `null`. Let's discuss this in slack.

src/configuration/routes/https.md
```diff
@@ -0,0 +1,117 @@
...
+### Client authenticated TLS
+
+In some non-browser applications (such as mobile applications, IoT devices, or other restricted-client-list use cases), it is beneficial to restrict access to selected devices using TLS.  This process is known as client-authenticated TLS, and functions effectively as a more secure alternative to HTTP Basic Auth.
+
+Platform.sh allows you to set the server-side portion of such a configuration via the `routes.yaml` file.  Doing so entails setting client authentication to required, and then providing a signed certificate either inline or via a file reference.
+
+```yaml
+tls:
+    client_authentication: "require"
+    client_certificate_authorities: !file cert.key
```
ralt: If not provided, the default list of root CA certificates is used.

Crell: I... still don't understand what that means.

src/configuration/routes/https.md
Crell: Updated

barryrld:
My comment may be in the wrong place as this is the docs repo. However...

We recently had a security audit on one of own sites (not on Platform) and some of the recommendations were to remove some of the insecure and deprecated cipher algorithms as well as the actual TLS protocols.

We're looking at an option to move this particular system to Platform and would ideally need to be able to configure the ciphers to achieve the best [Qualys](https://www.ssllabs.com/ssltest/) grades.

Where would be the best place to log this as a feature request (or is something already slated?)

Thanks! Barry

src/configuration/routes/https.md
```diff
@@ -0,0 +1,127 @@
...
+        preload: true
+```
+
+There are three sub-properties for the `strict_transport_security` property:
+
+* `enabled`: Can be `true`, `false`, or `null`.  Defaults to `null`.  If `false`, the other properties wil be ignored.
+* `include_subdomains`: Can be `true` or `false`.  Defaults to `false`. If `true`, browsers will be instructed to apply HSTS restrictions to all subdomains as well.
+* `preload`: Can be `true` or `false`.  Defaults to `false`.  If `true`, Google and others may add your site to a lookup reference of sites that should only ever be connected to over HTTPS.  Many although not all browsers will consult this list before connecting to a site over HTTP and switch to HTTPS if instructed.  Although not part of the HSTS specification it is supported by most browsers.
+
+If enabled, the `Strict-Transport-Security` header will always be sent with a lifetime of 1 year.  The [Mozilla Developer Network]((https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security)) has more detailed information on HSTS.
```
ralt: There are superfluous parentheses there, see how it's rendered on the site.

Crell:
@barryrld This functionality has been implemented but is not yet available.  It will be available soon, when we release the next version of the per-project backend application.  As soon as that's released the ability to disable older protocols as described here will be available to all projects.

So, sit tight, new toys coming soon. :smile:

barryrld:
@Crell wahoo!

Good news. Look forward to the release.

damz:
@barryrld Our configuration is already a `A` on Qualys SSLabs in all the environments. Removing TLS 1.0 is specifically for PCI-DSS, it doesn't result in any significant security improvement.

src/configuration/routes/https.md
stefandoorn: It would be interesting to be able to set the lifetime, in order to start testing with a low lifetime, checking that all works properly, and then adjusting it up.

PR #666, opened 2017-10-17
Link Magento 1 example project to correct page

dbrekelmans:
The Magento 1 link currently redirects to the Laravel example project, but should redirect to the Magento 1 example project.


Crell:
Also, congratulations on having PR #666. :smile:

dbrekelmans:
Thanks! 😈

PR #670, opened 2017-10-18
Document CLI `mount:list`, `mount:upload` and `mount:download` commands

pjcdawkins:
Released in CLI v3.21.0


src/configuration/app/storage.md
```diff
@@ -60,3 +60,58 @@ web:
 ```
 
 See the [web locations](/configuration/app/web.md) documentation for more details.
+
+## Using the Platform.sh CLI with mounts
```
Crell: Let's move this section up above the "How do I?" sections.  Those should come last on each page.

src/configuration/app/storage.md
```diff
@@ -60,3 +60,58 @@ web:
...
+Downloading files from the remote mount /app/private to /Users/alice/Projects/foo/private
+  receiving file list ...   done
+
+  sent 16 bytes  received 3.73K bytes  2.50K bytes/sec
+  total size is 1.77M  speedup is 471.78
+  time: 0.91s
+The download completed successfully.
+```
+
+Uploading files to a mount is similar:
```
Crell: Should we note in here somewhere that it uses rsync under the hood, so is fast for incremental updates?

src/development/access-site.md
```diff
@@ -49,3 +49,9 @@ rsync -az `platform ssh --pipe`:web/uploads/ ./uploads/
 ```
 
 This command will copy all files in the `web/uploads` directory on the current environment to the `uploads` directory locally.  Note that `rsync` is very sensitive about trailing `/` characters, so that may change the meaning of a command.  Consult the `rsync` documentation for more details.  Also see our [migrating](/tutorials/migrating.md) and [exporting](/tutorials/exporting.md) guides for more examples using `rsync`.
+
+If you have the [Platform.sh CLI](/gettingstarted/cli.md) installed, rsync can be used via the `mount:download` and `mount:upload` commands. This example achieves the same result as above:
+
+```bash
+platform mount:download --mount web/uploads --target ./uploads
+```
```
Crell: This chunk should

1) Link to the larger description on the storage page
2) Also exist in some form on the exporting and migrating pages under tutorials.

Crell:
Question.  Should the long-form description appear on the web/storage page, or on the accessing-site page?  Wherever it lives the others should link to it.  I'm not sure which place is best for it to live, but I'm leaning toward access-site.

src/configuration/app/storage.md
pjcdawkins: Then I'd have to change the examples too: currently I have `For example, to download files from the 'private' mount`...

pjcdawkins:
Agreed, I'll move the long-form description into `access-site.md`.

pjcdawkins:
I feel the longer description should be mainly under `storage.md` because that's where mounts are introduced and described. And perhaps the `Rsync` section should be moved there as well? The `access-site.md` page is a higher level thing, which could link to it.

src/development/access-site.md
```diff
@@ -4,48 +4,124 @@ Once you have an environment running, there are many ways to access it to perfor
...
-From the command line, you can also use the `scp` command in combination with the `--pipe` switch.  For example:
+### SCP
+
+SCP is a simple command-line utility to copy files to and from a remote environment.
+
+For example, this command:
 
 ```bash
-scp `platform ssh --pipe`:web/uploads/diagram.png .
+scp "$(platform ssh --pipe)":web/uploads/diagram.png .
```
Crell: Why this change?

src/development/access-site.md
```diff
@@ -4,48 +4,124 @@ Once you have an environment running, there are many ways to access it to perfor
 
 By design, the only way to deploy new code is to push to the corresponding branch.  That ensures a consistent, repeatable, auditable application instance at all times.
 
-## SSH
+## Visiting the site on the web
 
-Most interactions with Platform.sh require SSH key authentication, and you will need to [setup your SSH keys](/development/ssh.md) before working on a site.
+The web URL(s) for the site are listed in the (web UI)[/administration/web.md] under "Access site".
```
Crell: The link brackets are backwards here.

Crell: [ link text ] ( link url )

src/development/access-site.md
```diff
@@ -4,48 +4,124 @@ Once you have an environment running, there are many ways to access it to perfor
...
 ```
 
-That will output the connection string for SSH, including the username and host for the current project and environment.  It will look something like `<project ID>-<environment ID>--app@ssh.us.platform.sh`.  The part before the `@` is the username, the part after is the host.  Enter both of those into your SFTP client.  No password is necessary, but your SFTP client will need to have access to the SSH private key that corresponds to the public key on Platform.sh.
+That will output the connection string for SSH, including the username and host for the current project and environment.  It will look something like `<project ID>-<environment ID>--app@ssh.us.platform.sh`.  The part before the `@` is the username, the part after is the host.  Enter both of those into your SSH/SFTP client.  No password is necessary, but your client will need to have access to the SSH private key that corresponds to the public key on Platform.sh.
+
+### SFTP
+
+SFTP is another way to upload and download files to and from a remote environment. There are many SFTP clients available for every possible operating system - graphical and command-line - use whichever one works for you.
```
Crell: "use whichever..." is its own sentence.  It should either have a period or semi-colon before it, not a dash.  Also, the dash between system and graphical should be an mdash HTML code.

Crell:
A few minor points but this is looking way better!

src/development/access-site.md
pjcdawkins: Just as a good habit (https://stackoverflow.com/questions/9449778/what-is-the-benefit-of-using-instead-of-backticks-in-shell-scripts), but it doesn't really matter in this context.

PR #676, opened 2017-10-23
Change 3rd Party to Tandem.

serundeputy:


Crell:
How about "by Tandem, a 3rd party agency, ..." ? :smile:

JGrubb:
I personally think this is totally fine.

serundeputy:
@Crell @JGrubb what say you all?  Leave or change it? Should I stay or should I go now? :) 🎵

Crell:
I'd include the 3rd party mention, as otherwise the sentence is a bit clunky.  "It's maintained by Tandem, but it's really OK!" vs "It's maintained by Tandem who is a 3rd party, but it's really OK!"  (It's not that it's Tandem that requires the "no really, it's OK" but the fact that it's not-Platform but is still OK.)

serundeputy:
@Crell thanks; updated the PR accordingly.

Crell:
Merci bien.

PR #683, opened 2017-10-26
Removing experimental features from the screenshot.

gnufred:
We might not want those feature exposed first hand to new users.
Check ticket #21509 for an example.


Crell:
Seems this should be done as part of #590, which needs to move forward, no?

gnufred:
@Crell It should be part of the CLI refactor. But to simplify things can we merge or close this MR and move on to #590?

Crell:
Closing in favor of #590

PR #686, opened 2017-10-30
Add a page detailing how YAML works, and our extensions to it

Crell:
This could definitely use some feedback to make sure it's logical and flows well.


src/configuration/yaml.md
```diff
@@ -0,0 +1,173 @@
...
+type: 'php:7.1'
+build:
+    flavor: 'composer'
+
+disk: 1024
+```
+
+This example defines a key `name` with value `app`, a key `type` with value `php:7.1`, a key `disk` with a value 1024, and a key `build` that is itself a nested set of key/value pairs, of which there is only one: `flavor`, whose value is `composer`.  Nested values are often referenced using a dotted syntax, such as `build.flavor`.
+
+Keys are always strings, and may be quoted or not.  Values may be strings, numbers, or booleans.  Alphanumeric strings may be quoted or not.  More complex strings (with punctuation, etc.) must be quoted.  Numbers should not be quoted.  The boolean values `true` and `false` should never be quoted.
```
ralt: Values can be dates too, although we don't use them.

ralt: Single or double-quoted? :-)

src/configuration/yaml.md
```diff
@@ -0,0 +1,173 @@
...
+
+In general the order of keys in a YAML file does not matter.  Neither do blank lines.  Indentation may be with any number of spaces, as long as it is consistent throughout the file.  Platform.sh examples by convention use four-space indentation.
```
ralt: We actually do care about the order.

src/configuration/yaml.md
```diff
@@ -0,0 +1,173 @@
...
+
+creates a nested property `hooks.build`, which has the value `set -e\ncp a.txt b.txt`.  (That is, a string with a line break in it.)  That is useful primarily for hooks, which allow the user to enter small shell scripts within the YAML file.
+
+## Local tags
+
+YAML allows for special "tags" on values that change their meaning.  These tags may be customized for individual applications so may vary from one system to another.  For Platform.sh there are two that are useful to know about.
+
+### !archive
+
+The `!archive` tag specifies that the value it tags is a reference to a directory on disk, relative to the location of the YAML file.  Essentially it defines the value of key as "this entire directory".  For example:
```
ralt: Let's only document `!include`?

src/configuration/yaml.md
Crell: Worth mentioning?  Are dates not just strings?

Crell: I knew someone would ask about that...  Now I have to get into interpolation rules!

src/configuration/yaml.md
Crell: We do?  Where/how?

src/configuration/yaml.md
Crell: Er.  Why?  Anyone using Solr with Psh right now is using !archive.  Not mentioning it seems very obtuse.

src/configuration/yaml.md
ralt: Not sure it's worth mentioning. And no, they're not quoted, ans they're parsed as dates.

src/configuration/yaml.md
ralt: Or at least mention that it's deprecated? Dunno, if we're superseding `!archive`, not giving it a big place in the doc makes sense to me.

src/configuration/yaml.md
Crell: I'll leave it out for now; this isn't intended as a comprehensive guide, just as a Platform-relevant guide.

Crell: Interpolation rules added.

pjcdawkins: As implied in the previous paragraph, values can be key-value objects too.

src/configuration/yaml.md
```diff
@@ -0,0 +1,169 @@
...
+```yaml
+name: 'app'
+type: 'php:7.1'
+build:
+    flavor: 'composer'
+
+disk: 1024
+```
+
+This example defines a key `name` with value `app`, a key `type` with value `php:7.1`, a key `disk` with a value 1024, and a key `build` that is itself a nested set of key/value pairs, of which there is only one: `flavor`, whose value is `composer`.  Nested values are often referenced using a dotted syntax, such as `build.flavor`.
```
pjcdawkins: The dotted syntax isn't YAML, so, maybe "In our documentation, nested values are...".

src/configuration/yaml.md
```diff
@@ -0,0 +1,169 @@
...
+```
+
+```yaml
+hooks:
+    build: !include
+      type: string
+      path: build.sh
+```
+
+That is primarily useful for breaking longer build scripts or inlined configuration files out to a separate file for easier maintenance.
```
pjcdawkins: But this would be simpler, if build.sh is executable:

```yaml
hooks:
    build: ./build.sh
```

Crell: ... True, that would be even better.  I don't want to get into the Varnish example here, though, especially as that's not published yet.  Have you a better suggestion for an example to use?

PR #687, opened 2017-11-03
Remove note about old subdomain format (---)

pjcdawkins:
I think it is now the default everywhere to use a `.` instead of `---`?


Crell:
I don't believe that's the case.  @ralt please confirm?

ralt:
It will be after Git 2.7 goes GA. Though existing projects won't be affected.

Crell:
Oh lovely.  That means we need to account for both patterns in the Docs, not just remove the note.

Crell:
Closing in favor of #689

PR #690, opened 2017-11-03
Git 2.7 change logs

Crell:
The changes are spread across a couple of PRs, so I'm putting all the changelog stuff together here to avoid merge conflicts.

### DIFF
--- a/src/changelog/2017.md
+++ b/src/changelog/2017.md
@@ -14,7 +14,7 @@ A new syntax has been added for defining [mount points](/configuration/app/stora
 
 ## Blocking older TLS versions
 
-It is now possible to disable support for [HTTPS requests](/configuration/routes/https.md) using older versions of TLS.  TLS 1.0 is known to be insecure in some circumstances and is being phased out by websites and some security requirement policies.
+It is now possible to disable support for [HTTPS requests](/configuration/routes/https.md) using older versions of TLS.  TLS 1.0 is known to be insecure in some circumstances and some compliance standards require a higher minimum supported version.
 
 ## `{all}` placeholder for routes
 
--- a/src/changelog/2017.md
+++ b/src/changelog/2017.md
@@ -1,4 +1,4 @@
-# November 2017
+# December 2017
 
 ## New project subdomains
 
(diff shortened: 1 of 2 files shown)


src/changelog/2017.md
```diff
@@ -1,3 +1,25 @@
...
+
+All YAML configuration files now support a generic [`!include`](/configuration/yaml.md) tag that can be used to embed one file within another.
+
+## Extended mount definitions
+
+A new syntax has been added for defining [mount points](/configuration/app/storage.md) that is more self-descriptive and makes future extension easier.
+
+## Blocking older TLS versions
+
+It is now possible to explicitly reject [HTTPS requests](/configuration/routes/https.md) using older versions of TLS with known insecurities.
```
ralt: I would avoid the "with known insecurities" part. Looks like FUD more than anything else.

sruggier: If we want to seem more objective, we can say something like "it's now possible to disable support for HTTPS requests over TLS 1.0. TLS 1.0 is enabled by default for backward compatibility. However, it's known to be insecure, because it's vulnerable to the BEAST attack, and many websites have already discontinued support for it, or are planning to do so later this year."

Web servers using TLS 1.0 have a choice between using more secure ciphers that are vulnerable to BEAST due to a protocol vulnerability, or using RC4, which is a pretty broken cipher at this point, but isn't vulnerable to BEAST. Most commonly used web clients have now implemented mitigations for BEAST, but TLS 1.0 is still objectively broken.

Sources:
 https://www.helpnetsecurity.com/2011/10/18/mitigating-the-beast-attack-on-tls/
https://blog.qualys.com/ssllabs/2013/03/19/rc4-in-tls-is-broken-now-what
https://security.stackexchange.com/a/32908

Crell: I split the difference at:  "It is now possible to disable support for HTTPS requests using older versions of TLS.  TLS 1.0 is known to be insecure in some circumstances and is being phased out by websites and some security requirement policies."

ralt: That still seems too much out of line. Saying "Some compliance standards require a minimum supported version" should be enough, imho.

PR #691, opened 2017-11-05
Update mail config to use PLATFORM_SMTP_HOST

stefandoorn:
After discussion on Slack with @damz, some changes to the mail configuration for Symfony + added a line regarding the usage of the environment variable that holds the SMTP host.


ralt:
Thanks!

PR #694, opened 2017-11-18
[Redis] Remove Confusion with Relationship name

Plopix:
The documentation is unclear to me, it is confusing when it comes to the relationship name.

The `Assuming your Redis relationship is named 'rediscache'` is weird as in the previous configuration `rediscache` is just the name of the service.

Letting `redis` as the name is also a bit confusing to me. Then I have added the configuration again to make it clear.

Let me know

Best,

### DIFF
--- a/src/configuration/services/redis.md
+++ b/src/configuration/services/redis.md
@@ -77,7 +77,7 @@ runtime:
         - redis
 
 relationships:
-    myredis: "rediscache:redis"
+    fooredis: "rediscache:redis"
 ```
 
 You can then use the service in a configuration file of your application with something like:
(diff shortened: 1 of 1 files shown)


pjcdawkins:
Thanks, this is definitely an improvement. I've added a commit to update the previous example to `myredis` for consistency.

Crell:
I've been trying to actively avoid "my*" as a name for services, as it's not particularly descriptive or extensible.  "my*" examples generally are not very good.

The rest of the changes here I like, especially adding how to use Redis for native sessions.  Let's try for a better relationship name, though.  I'm open to suggestions.

Plopix:
fair enough @Crell! I renamed with `fooredis`.

Let me know! ping @pjcdawkins

Crell:
eh, `foo` isn't really much better.  It's just `my` with an extra character. :smile: The intent should be to have default names that people can really use, because most will be copy-pasting it so we want meaningful examples.

pjcdawkins:
I suggest `redis`!

Plopix:
@Crell how about now? `applicationcache` and `sessionstorage`
@pjcdawkins To me using `redis` would introduce another confusion with the `endpoint`

> I think if these names are still not good, @Crell you should decide and tell us ;) OR we should merge and decide/update later (creating an issue)

Crell:
applicationcache and sessionstorage actually sound great.  Specific, descriptive, all the good things.  :+1: 

Thanks, @Plopix!

PR #696, opened 2017-11-28
Install Apache Tika on Platform.sh

serundeputy:


serundeputy:
@JGrubb I think this would be a useful addition to the docs and you could point people that inquire about Tika to this post.

JGrubb:
Agreed, and good call @serundeputy

PR #704, opened 2017-11-30
fix broken environment link

DanielSchwiperich:
without slash at the beginning it will be appended to overview path which does not work. 

With this it's also  in line with contribution guidelines
> Internal links should be absolute (starting with /) and link to a file ending in .md, not .html. That makes them easier to find when needed to update.

See broken version here: https://docs.platform.sh/overview/build-deploy.html


Crell:
You are correct!  Good catch, thanks.

PR #707, opened 2017-12-05
Update nodejs supported versions list

sruggier:


Crell:
Per discussion yesterday, instead of removing 8.2 outright let's add a second list for "deprecated" and put 8.2 there as well.  We can then follow-up to do the same for PHP and any other services/languages needed.

Crell:
Closing in favor of #712

PR #708, opened 2017-12-06
Complete the 2FA question on FAQ

mrkschan:


Crell:
LOL.  How'd we miss that one for so long?  Thanks.

PR #709, opened 2017-12-08
Quick Start

OriPekelman:


Crell:
This needs a lot more explanation of what it's actually doing.  It's not even slightly obvious.  And we may want to remove some other parts of the page if we're going to recommend this approach.  (Once I figured out what it was doing I like it better, but it's not even a little self-evident.)

I'll revise this.

Crell:
Fleshed this out further.  I'm still not sure I like the "quick start" section, but it does seem useful to offer a one-liner.  Open to alternate suggestions there.

Crell:
Actually no, with further testing the code here doesn't actually work.  PHP's local webserver isn't picking up the PLATFORM_RELATIONSHIPS env var.  Investigating further.

Crell:
Figured it out.  PHP's webserver doesn't read env vars by default.  This is annoying.  I have documented that, as well as the other problem I spent too long chasing.  This is now tested and IMO mergable.

pjcdawkins:
Technically env vars are available by default, but via `getenv()`, not via the $_ENV superglobal.

Crell:
Damnit, PHP...

We are currently using $_ENV in all of our sample code; not for any particularly special reason, I guess.  We could change that but it wouldn't help current users if they read a command on this page that would only work the other way.  I'm open to suggestions on how we want to resolve that.

pjcdawkins:
Release `platform server:start --tunnel`? That starts PHP web server(s) with the right variables_order. But it's not for everyone; it's not going to support fancy things like custom `start` commands etc.

Crell:
That sounds complicated...  

Proposal: Merge the docs as-is, and then also switch our sample code over to use `getenv()` so it's less of an issue in the future.  Can I get an amen?  (By which I mean a merge.)

PR #711, opened 2017-12-08
Simplify and standardize redeploy-variable-set technique examples.

Crell:
Because there's no need to use big words when a diminutive one will suffice.


src/configuration/routes/https.md
```diff
@@ -7,7 +7,7 @@ All environments on Platform.sh support both HTTP and HTTPS automatically.  Prod
 > **note**
 > Let's Encrypt certificate renewals are attempted each time your environment is deployed. If your project does not receive regular code commits, you will need to manually issue a re-deployment to ensure the certificate remains valid. We suggest that you do so when your project doesn't receive any updates for over 1 month. This can be done by pushing a code change via git or issuing the following command from your **local** environment:
 > ```
-> NOW=$(date +"%F_%H:%M:%S") && platform variable:set -W -y force-le-renewal $NOW
+> platform variable:set -W force-le-renewal $(date)
```
pjcdawkins: Needs to be inside ""

PR #712, opened 2017-12-12
Deprecated app versions

Crell:
Supersedes #707, and gives the same split-list treatment to PHP.

### DIFF
--- a/src/languages/nodejs.md
+++ b/src/languages/nodejs.md
@@ -13,7 +13,7 @@ Javascript and PHP applications.
 * 6.9
 * 6.10
 * 6.11
-* 8.2
+* 8.9
 
 See https://github.com/platformsh/platformsh-example-nodejs/tree/mongodb for a
 full example with MongoDB support.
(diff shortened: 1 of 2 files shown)


src/languages/nodejs.md
```diff
@@ -7,12 +7,18 @@ Javascript and PHP applications.
 
 ## Supported versions
 
+* 4.8
+* 6.11
+* 8.9
+
+## Deprecated versions
+
+The following versions are available but are not receiving security updates from upstream, so their use is not recommended.  They will be removed at some point in the future.
```
vincenzo: You've got a double space between `recommended.` and `They`.

src/languages/php.md
```diff
@@ -35,6 +33,15 @@ To select a PHP version, specify a `type` such as `php:7.1` or `hhvm:3.9`:
 type: "php:7.1"
 ```
 
+## Deprecated versions
+
+The following versions are available but are not receiving security updates from upstream, so their use is not recommended.  They will be removed at some point in the future.
```
vincenzo: You've got a double space between `recommended.` and `They`.

PR #714, opened 2017-12-13
Add memcached-for-Drupal 8 instructions

Crell:
Developed from working with a customer who was trying to narrow down some odd cache-related issues.  Not recommended but no sense not documenting it.

### DIFF
--- a/src/frameworks/drupal8/memcached.md
+++ b/src/frameworks/drupal8/memcached.md
@@ -13,7 +13,7 @@ cacheservice:
     type: memcached:1.4
 ```
 
-That will create a service named `cache`, of type `memcached`, specifically version `1.4`.
+That will create a service named `cacheservice`, of type `memcached`, specifically version `1.4`.
 
 ### Expose the Memcached service to your application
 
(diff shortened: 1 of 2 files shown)


src/frameworks/drupal8/memcached.md
```diff
@@ -0,0 +1,147 @@
...
+  }
+}
+```
+
+The `example.services.yml` file noted above will also use Redis for the lock and flood
+control systems.
+
+The redis module is able to use Redis as a queue backend, however, that should not be done on an ephemeral Redis instance as that could result in lost items when the Redis service instance is restarted or fills up.  If you wish to use Redis for the queue we recommend using a separate persistent Redis instance.  See the [Redis documentation page](/configuration/services/redis.md) for more information.
+
+### Verifying Redis is running
```
ralt: ???

src/frameworks/drupal8/memcached.md
```diff
@@ -0,0 +1,147 @@
...
+```
+
+That will create a service named `cache`, of type `memcached`, specifically version `1.4`.
+
+### Expose the Memcached service to your application
+
+In your `.platform.app.yaml` file, we now need to open a connection to the new Redis service.  Under the `relationships` section, add the following:
+
+```yaml
+relationships:
```
ralt: Choose a different name for the key?

Crell: New name chosen.

src/frameworks/drupal8/memcached.md
```diff
@@ -0,0 +1,135 @@
...
+    cache: "cacheservice:memcached"
+```
+
+The key (left side) is the name that will be exposed to the application in the `PLATFORM_RELATIONSHIPS` [variable](/development/variables.md).  The right hand side is the name of the service we specified above (`cache`) and the endpoint (`memcached`).  If you named the service something different above, change `cache` to that.
```
ralt: > change `cacheservice` to that.

PR #715, opened 2017-12-13
Solr tunnel instructions

Crell:
I'm removing these from the PE docs, but they're valid for PS so may as well include it.

### DIFF
--- a/src/configuration/services/solr.md
+++ b/src/configuration/services/solr.md
@@ -196,3 +196,28 @@ solrsearch:
 ```
 
 The Solr 6.x Drupal 8 configuration files are reasonably generic and should work in many other circumstances, but explicitly defining a core, configuration, and endpoint is generally recommended.
+
+## Accessing the Solr admin interface
+
+Because Solr uses HTTP for both its API and admin interface it's possible to access the admin interface over an SSH tunnel.
+
+```
+platform tunnel:open
+```
+
+```
+That will open an SSH tunnel to all services on the current environment, and give an output similar to:
+
+SSH tunnel opened on port 30000 to relationship: solr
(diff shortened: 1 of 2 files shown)


Crell:
Tagging @MacWeber for review/merge since Solr's his bag. :smile:

src/frameworks/drupal8/solr.md
```diff
@@ -108,3 +108,28 @@ if (isset($_ENV['PLATFORM_RELATIONSHIPS'])) {
...
+
+List tunnels with: platform tunnels
+View tunnel details with: platform tunnel:info
+Close tunnels with: platform tunnel:close
+```
+
+In this example, you can now open `http://localhost:30000/solr/` in a browser to access the Solr admin interface.  Note that you cannot create indexes or users this way, but you can browse the existing indexes and manipulate the stored data.
+
+> **Note**
+> Platform.sh Enterprise users can use `ssh -L 8888:localhost:<port> <user>@<cluster-name>.ent.platform.sh` to open a tunnel instead.
```
MacWeber: AFAIK, `<port>` here is always `8983`.

src/configuration/services/solr.md
```diff
@@ -196,3 +196,28 @@ solrsearch:
 ```
 
 The Solr 6.x Drupal 8 configuration files are reasonably generic and should work in many other circumstances, but explicitly defining a core, configuration, and endpoint is generally recommended.
+
+## Accessing the Solr admin interface
```
MacWeber: Maybe we better write "Accessing the Solr server's admin interface" to avoid confusion with the framewok's UI admin pages.

PR #718, opened 2017-12-18
Automated backups

Crell:
This includes both how to setup a CLI token, as well as how to use that to trigger nightly backups automatically.

@pjcdawkins Please confirm I'm not setting us up for a bad time. :smile:


Crell:
This is also worth a blog post once it's merged to docs, methinks.

src/administration/snapshot-and-restore.md
```diff
@@ -42,25 +41,51 @@ Finding snapshots for the environment master
...
-$ platform snapshot:create --yes --no-wait
+Now add a build hook to your `.platform.app.yaml` file to download the CLI as part of the build process.  
+
+```yaml
+hooks:
+    build: |
+        curl -sS https://platform.sh/cli/installer | php
 ```
 
+That will download the CLI to a known directory within the `.platform` directory, which is available in the path.  Because the API Token is available it will now be able to run authenticated commands against any project available to the user who created the token.
```
pjcdawkins: This will download the CLI to a known directory, `.platformsh/bin`, which will be added to the `PATH` at runtime (via the `.environment` file). Because the API token is available, the CLI will now be able to run authenticated commands, acting as the user who created the token.

src/gettingstarted/cli.md
```diff
@@ -97,3 +97,30 @@ Once Bash for Windows is installed, you can install the Platform.sh CLI with the
 ```bash
 curl -sS https://platform.sh/cli/installer | php
 ```
+
+## API Tokens
+
+The CLI can also be used from 3rd party CI services or other automation tools, and supports an API Token authentication option for that purpose.
```
pjcdawkins: The "3rd party" bit is not important (it can also be used with 1st-party automation tools, as demonstrated above).

src/gettingstarted/cli.md
```diff
@@ -97,3 +97,30 @@ Once Bash for Windows is installed, you can install the Platform.sh CLI with the
...
+
+An API token can be created through the UI.  Go to the "User" page from your accounts drop-down, then select the "Account Settings" tab, then the "API Tokens".
```
pjcdawkins: Is "your accounts drop-down" clear?

src/gettingstarted/cli.md
```diff
@@ -97,3 +97,30 @@ Once Bash for Windows is installed, you can install the Platform.sh CLI with the
...
+
+Click the "Create an API Token" link.  You may be asked to reverify your password.
+
+Enter a unique application name to identify the token.  The API token can authorize the CLI to access project information, or project information and account information.  In most cases it's not necessary to give the CLI access to change account settings so that checkbox can be unchecked.
```
pjcdawkins: Hm. I think this scope option is actually totally irrelevant, for now.

Crell: Well, the checkbox is there so we should say something about it.  What should we say if not the above?

src/gettingstarted/cli.md
Crell: Removed.

src/gettingstarted/cli.md
Crell: I'm not sure.  I didn't want to go through and add a dozen extra screenshots, but text is hard. :smile: 

@josephdpurcell Your thoughts?  Was this sufficiently clear?

josephdpurcell: It is sufficiently clear after having gone to the "User" page before. I admit when I very first logged into Platform.sh and clicked my profile image drop down and didn't see a page that said "My Profile" or "Settings" or something similar, I was confused. I thought the simple word "User" was ambiguous, oddly enough.

All that said, I think the way this is written is fine. People who get to this page will most likely have already been to the "User" page before and will know what the phrase `Go to the "User" page from your accounts drop-down` means.

src/gettingstarted/cli.md
```diff
@@ -97,3 +97,28 @@ Once Bash for Windows is installed, you can install the Platform.sh CLI with the
...
+
+The CLI can also be used from CI services or other automation tools, and supports an API Token authentication option for that purpose.
+
+An API token can be created through the UI.  Go to the "User" page from your accounts drop-down, then select the "Account Settings" tab, then the "API Tokens".
+
+Click the "Create an API Token" link.  You may be asked to reverify your password, then enter a unique application name to identify the token.
+
+![API Token screen](/images/api-token-list.png)
+
+After creating the token, you may view it by clicking the "view" link next to the token name.  You will be asked to reverify your password as well.
```
pjcdawkins: Also, directly after creating the token, you'll be shown the token value in a temporary message (a `drupal_set_message()`, if you see what I mean). To see it again, as you say, you will need to click "view" and you will need to re-enter your password.

Crell: I completely missed that because it's all the way at the top of the page. :smile: Will update.

src/administration/snapshot-and-restore.md
```diff
@@ -1,34 +1,33 @@
...
-## Snapshot
+## Snapshots
 
-### Manual snapshot
+Snapshots are triggered directly via the Web Interface or via the CLI. The snapshot creates a complete backup of the environment. It includes all persistent data from all running services (MySQL, SOLR,...) and any files stored on the mounted volumes.
 
-Snapshots are triggered directly via the Web Interface or via the CLI. The snapshot
-creates a complete backup of the environment. It includes all persistent data from
-all running services (MySQL, SOLR,...) and any files stored on the mounted volumes.
+You will need the "admin" role on the project/environment to be snapshotted in order to do so.
```
pjcdawkins: I couldn't scan this at first. Perhaps:

> You need to have the "admin" role in order to create a snapshot of an environment.

Crell: Fixed.

pjcdawkins:
An aspect of this I've forgotten until now (sorry)... we should recommend that you create a machine user specifically for accessing project(s) to create automated backups (create API tokens as that user). The machine user would need admin access to the projects in question, of course, but it's still a worthwhile separation.

src/administration/snapshot-and-restore.md
```diff
@@ -42,25 +41,41 @@ Finding snapshots for the environment master
...
-### Automated snapshots
+## Automated snapshots
 
-No snapshot is triggered automatically on Platform.sh Standard. You can
-trigger your snapshot via the Web Interface or via the CLI.
+No snapshot is triggered automatically on Platform.sh Standard. You can trigger your snapshot via the Web Interface or via the CLI.
 
-If you want to automate your snapshots, you can use Jenkins and trigger
-with the CLI command as follows:
+Snapshots may be triggered from an automated system, such as cron, Jenkins, or another CI service by calling the CLI.  To enable backups automatically using just Platform.sh, see the section on [API tokens](/gettingstarted/cli/api-tokens.md).  For installing the CLI in the app container.
```
josephdpurcell: The last sentence is not a complete sentence.

src/administration/snapshot-and-restore.md
```diff
@@ -42,25 +41,41 @@ Finding snapshots for the environment master
...
 
-```bash
-$ platform snapshot:create --yes --no-wait
+Once the CLI is installed an API token configured you can add a cron task to run once a day and trigger a snapshot.  The CLI will read the existing environment variables in the container and default to the project and environment it is running on.  However, in most cases such backups are only useful on the `master` production environment.  That can be achieved like so:
```
josephdpurcell: Once the CLI "has" installed?

Crell: Nope, missing an "and" later in the sentence. :smile:

src/gettingstarted/cli/api-tokens.md
```diff
@@ -0,0 +1,63 @@
...
+
+> **note**
+>
+> If running CLI commands from any automated system, including a Platform.sh cron task, we urge you to use the `--no-wait` flag on any commands that may take more than a second or two to avoid blocking the process.
+
+## Automation users
+
+For projects with more than one user it is often a good idea to create a dedicated "automation user" to run automation tasks such as backups.  That way, that user can be given a very restrictive set of permissions limited to just its needed tasks.  It will also show up in logs and activity streams as a separate entry from human users.
+
+Every user account needs its own unique email address.  This account will not need an SSH key, however.  The permissions it needs on a given project will depend on what automation tasks it will be running.  Backups, for instance, require `admin` access, while checking out code from a CI server to run tests on it would require only `reader` access.
```
pjcdawkins: "will not need" -> "may not need" (it would need a key for `git clone`, for example, or `platform ssh ...`)

src/gettingstarted/cli/api-tokens.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Obtaining a token
+
+The Platform.sh CLI can also be used from CI services or other automation tools, and supports an API Token authentication option for that purpose.
+
+An API token can be created through the UI.  Go to the "User" page from your accounts drop-down, then select the "Account Settings" tab, then the "API Tokens".
+
+Click the "Create an API Token" link.  You may be asked to reverify your password, then enter a unique application name to identify the token.
+
+![API Token screen](/images/api-token-list.png)
```
josephdpurcell: I believe this needs to be `src/images/api-token-list.png`

Looking at the preview: https://github.com/platformsh/platformsh-docs/blob/cli-backups/src/gettingstarted/cli/api-tokens.md

You can see the image does not display.

pjcdawkins: It's correct - have a look at https://pr-718-7beqdoa-t2llqeifuhpzg.eu.platform.sh/gettingstarted/cli/api-tokens.html

PR #719, opened 2017-12-19
Add Python 3.7 support

Crell:
@sruggier, please confirm if we should list any versions as deprecated at this point.

### DIFF
--- a/src/changelog/2018.md
+++ b/src/changelog/2018.md
@@ -1,3 +1,9 @@
+# September 2018
+
+## Python 3.7 support
+
+We now support [Python 3.7](/languages/python.md).
+
 # July 2018
 
 ## Security and Compliance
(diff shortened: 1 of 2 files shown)


sruggier:
We need to clarify that it's a pre-release of Python before this PR can be merged.

mcyprian:
Final release of Python 3.7 is available now. I guess there should be a small mention [like this](https://github.com/platformsh/platformsh-docs/pull/913/commits/1947d23281d09c2f08b1d2113b1e0cf952b2d50d#diff-28dcd1657b1f461dbbe35002ce9e6b7cR1) added to the changelog as well.

mcyprian:
Python 3.5 is still the default Python 3 on Debian Stretch, I think it is too soon for deprecation,

PR #720, opened 2017-12-19
Add RootNotFoundException error and how to fix it.

Crell:
As discussed in https://github.com/platformsh/platformsh-cli/issues/440#issuecomment-352832032

### DIFF
--- a/src/development/troubleshoot.md
+++ b/src/development/troubleshoot.md
@@ -154,7 +154,7 @@ Then the CLI hasn't been able to determine the project to use.  To fix that, run
 platform project:set-remote <project_id>
 ```
 
-where `<project_id>` is the random-character ID of the project.  That can be found in the UI after the `platform get` command shown or in the URL of the UI or project domain.
+where `<project_id>` is the random-character ID of the project.  That can be found by running `platform projects` from the command line to list all accessible projects.  Alternatively, it can be found in the UI after the `platform get` command shown or in the URL of the UI or project domain.
 
 ## "File not found" in Drupal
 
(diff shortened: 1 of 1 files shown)


src/development/troubleshoot.md
```diff
@@ -140,6 +140,21 @@ In Platform.sh, you cannot just "hack production".  It is a constraint, but it i
...
+[RootNotFoundException] Project root not found. This can only be run from inside a project directory.
+```
+
+Then the CLI hasn't been able to determine the project to use.  To fix that, run:
+
+```
+platform project:set-remote <project_id>
+```
+
+where `<project_id>` is the random-character ID of the project.  That can be found in the UI after the `platform get` command shown or in the URL of the UI or project domain.
```
pjcdawkins: or from `platform projects`

pjcdawkins:
As a follow-up... maybe the CLI should detect if you're in a (non-project) Git repository, and prompt you to use `project:set-remote`.

src/development/troubleshoot.md
Crell: Added.

Crell:
Better self-help in the CLI would be great, but I think it's fine to do both.

pjcdawkins:
I have committed https://github.com/platformsh/platformsh-cli/commit/7b48bb71d99f7070ba1362728d18b32da25b3c1e to `development`, it will reduce (it may or may not obviate) the need for documentation, _when_ it's released and _if/when_ everyone has upgraded to that version. So... I think this PR is still useful.

Crell:
OK.  Go ahead and merge this, then.  We can always remove it later if it makes sense.

PR #733, opened 2018-01-08
Remove inbound IP addresses for DE-1 and NL-1 regions

GuGuss:


GuGuss:
Inbound IP addresses on DE-1 and NL-1 are static. Closing PR.

PR #734, opened 2018-01-09
Add strong notice on foregrounding processes

OriPekelman:


src/configuration/app/web.md
```diff
@@ -25,6 +25,12 @@ web:
         start: 'uwsgi --ini conf/server.ini'
 ```
 
+> **NB: Never Background Processes!**
+>
+> Your start command _must_ run as a **foreground process** or all hell will break loose. Platform.sh supervises your started
+> process. If it is backgrounded (your command ends with `&` or starts with `bg`) our superviser will consider it is dead
```
ralt: s/superviser/supervisor/

ralt: Soft wrap please.

src/configuration/app/web.md
```diff
@@ -25,6 +25,10 @@ web:
         start: 'uwsgi --ini conf/server.ini'
 ```
 
+> **NB: Never Background Processes!**
+>
+> Your start command _must_ run as a **foreground process** or all hell will break loose. Platform.sh supervises your started process. If it is backgrounded (your command ends with `&` or starts with `bg`) our supervisor will consider it is dead and try to restart it. You will soon be out of resources. And the whole environment will become unresponsive.
```
Crell: Good thing to call out, but "all hell will break loose" is a bit too casual for the docs.

PR #736, opened 2018-01-09
Fix typo when requiring Drush via Composer.

achton:
This fixes a typo in the docs where you would be requiring a 4 year old Drush version from the Drupal package repo, instead of the fresh one from the Drush namespace.


pjcdawkins:
Thanks @achton

PR #738, opened 2018-01-11
Add info about sshing into workers

OriPekelman:


Crell:
This'll do until the CLI is able to handle it natively.  Thanks.

PR #742, opened 2018-01-18
Add eZ Platform page

Crell:
Includes:

* Redis setup instructions
* Fastly setup instructions
* Deep links to elsewhere.

Note that this references a page that won't exist until #744 is merged, but the result should be a tight, non-duplicating set of instructions.

Are there any other code files in the eZ repo we should deep-link to for example purposes?


src/frameworks/ez.md
```diff
@@ -0,0 +1,37 @@
...
+
+eZ Platform is a Composer-based PHP CMS, and as such fits well with the Platform.sh model.  As a Symfony-based application its setup is very similar to Symfony.
+
+eZ Platform comes pre-configured for use with Platform.sh in version 1.13 and later and version 2.0.1 and later.  Those are the only versions that are supported.  Appropriate Platform.sh configuration files are included in the eZ Platform application itself, but of course may be modified to suit your particular site if needed.
+
+## Cache and sessions
+
+By default, eZ Platform is configured to use a single Redis instance for both the application cache and session storage.  You may optionally choose to use a separate Redis instance for session storage in case you have a lot of authenticated traffic (and thus there would be many session records).
+
+To do so, uncomment the `redissession` entry in the `.platform/services.yaml` file and the corresponding relationship in the `.platform.app.yaml` file.  The bridge code that is provided with eZ Platform 1.13 and later will automatically detect the additional Redis service and use it for session storage.
```
vidarl: Looking forward for your input on https://github.com/ezsystems/ezplatform/pull/254#discussion_r162566352

Unless you have good reasons not to, I would recommend we go for two redis instances by default.

Crell:
Varnish-through-Fastly is only supported on eZ Platform Enterprise, as I understand it.  Someone using eZ Platform Enterprise I presume will be on Platform.sh Enterprise, where we can/should link to those docs.

PR #744, opened 2018-01-22
Fastly setup

Crell:
This is a split off of the eZ Platform documentation, since there's nothing eZ specific about this part.

I also moved the Custom SSL Cert section to its own page since most users won't be doing that, and renamed it to TLS for accuracy.  Because reasons.


andrerom:
What i'm missing here is best practice on how to deal with VCL.

It's then probably specific to enterprise, and afaik there are two options that needs to be described:
- Usage of Faslty API to add snippets, and if there is no access to this out of the box how to do that. Also how to deal with moving snippets from staging to prod, but this might only be relevant for those with separate staging env so might be unnecessary complex to describe her, which brings us to:
- Self managed Faslty, full access to deal with it yourself. How to ask for it, and how it affects monitoring of the install _(the SLA, but I would avoid saying the SLA does not imply or anything similar scare here, cause afaik the reality is just that the SLA and the monitoring rather then applies to backend) _

 As @JGrubb is also involved now he might have some input on that.

src/golive/steps/fastly.md
```diff
@@ -0,0 +1,26 @@
...
+
+Platform.sh does not offer an integrated CDN on Platform.sh Professional at this time.  However, Fastly is a common choice for our customers.
+
+Launching a Platform.sh site with Fastly in front of it is nearly the same as launching normally.  There are only two notable differences.
+
+Note that individual applications may have their own Fastly setup instructions or additional modules.  Consult the documentation for your application for specific details.
+
+## Set the Platform.sh domain on Fastly
+
+Rather than create a DNS CNAME for your Platform.sh master branch (for instance `master-7rqtwti-qwertyqwerty.eu.platform.sh`), point your Fastly account to the same domain.  Then configure your DNS to point your domain at Fastly instead.  See the [Fastly documentation](https://docs.fastly.com/guides/basic-configuration/connecting-to-origins) for further details.  Be sure to enable TLS for the backend connection to Platform.sh.
```
vidarl: I am not a native English speaker, so I have a hard time understanding what you exactly refer to in "point your Fastly account to the same domain"

What I think you mean is that when configuring native host at Fastly, you should use the CNAME generated by Platform.sh ( for instance master-7rqtwti-qwertyqwerty.eu.platform.sh )

Also maybe mention how to tell Fastly what your real domain is by refering to https://docs.fastly.com/guides/basic-configuration/working-with-domains

Crell:
@vidarl Updated text.  Give us a 👍 if you think this is ready to merge and I'll bug someone on our side to do so. 😄

andrerom:
@Crell Any chance that will be simplified in the future so it will be possible to give some consistent recommendations/guidlines? _(one for PE and one for PS would be ok for some cases, even if not optimal)_

src/golive/steps/tls.md
```diff
@@ -0,0 +1,33 @@
...
+
+You can add a custom certificate via the Platform.sh [Web Interface](/administration/web.md). Just go to the [project configuration page](/administration/web/configure-project.md) in the web interface and click on Domains. If you already have a domain, you can edit the domain and then click on the Add TLS certificate button. You can then add your private key, public key certificate and optional certificate chain.
+
+![UI configuration for TLS](/images/ui-ssl.png)
+
+### Adding a custom certificate through the CLI
+
+Example:
+```bash
+platform domain:add secure.example.com --cert=/etc/TLS/private/secure-example-com.crt --key=/etc/TLS/private/secure-example-com.key
```
pjcdawkins: I'm not clear what's better between `domain:add` and `certificate:add`.. we should check that.

PR #745, opened 2018-01-24
Add leading slash back into Dropbox mount names

xtfer:
Fixes reversion of https://github.com/platformsh/platformsh-docs/pull/543/commits/d35c7efe2936ea81d1b6eadd79977f911a7f8f8e.

### DIFF
--- a/src/tutorials/dropbox.md
+++ b/src/tutorials/dropbox.md
@@ -28,10 +28,10 @@ Because Dropbox files are mirrored on your writable disk mount, your application
 disk: 2048
 # the needed mounts for the Dropbox daemon
 mounts:
-    Dropbox:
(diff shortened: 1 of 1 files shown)


pjcdawkins:
I like that you're restoring the quotes, but [why] is the slash necessary?

xtfer:
I thought they were required... but now that I look around I see a mix. Perhaps we should agree on a standard (which, IMHO, should be leading slash's and quotes)? @Crell @pjcdawkins ?

Crell:
I thought I had standardized the docs when we switched to the new syntax.  Is it still inconsistent?

Crell:
Per discussion, file paths will not have a leading slash.

PR #746, opened 2018-01-24
Add back original mount syntax

xtfer:
Users of old mount syntax mix-and-match and it no workee.

### DIFF
--- a/src/tutorials/dropbox.md
+++ b/src/tutorials/dropbox.md
@@ -36,11 +36,12 @@ mounts:
         source_path: 'dropbox-meta'
 ```
 
-If your .platform.app.yaml file still uses the old syntax for mounts, this is equivalent to:
+If your `.platform.app.yaml` file still uses the old syntax for mounts, this is equivalent to:
 
 ```yaml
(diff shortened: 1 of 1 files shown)


Crell:
Hm.  Why is this failing?  It's... not like there's anything else going on with that page lately in master...

src/tutorials/dropbox.md
```diff
@@ -36,6 +36,13 @@ mounts:
         source_path: 'dropbox-meta'
 ```
 
+If your .platform.app.yaml file still uses the old syntax for mounts, this is equivalent to:
```
Crell: The file name should always be surrounded in backticks.

src/tutorials/dropbox.md
```diff
@@ -36,6 +36,13 @@ mounts:
...
+
+```yaml
+"/Dropbox": "shared:files/dropbox"
+"/.dropbox": "shared:files/dropbox-meta"
+```
```
Crell: Let's include the full `mounts` block so that the indentation makes sense.

PR #747, opened 2018-01-24
Document how to find disk usage by hand

Crell:
The particular filesystem ID here was scrambled so doesn't correspond to any particular project.

### DIFF
--- a/src/development/troubleshoot.md
+++ b/src/development/troubleshoot.md
@@ -89,7 +89,7 @@ Filesystem                                                       Size  Used Avai
 * The first entry shows the storage device that is shared by all of your disk mounts.  Only one path will be shown under `Mounted on` but the disk space reported is common to all defined mounts in a single pool.  In this example, there are 2 GB of total disk allocated to the app container of which only 2% (37 MB) has been used total by all defined mounts.
 * The second entry is the operating system temp directory, which is always the same size.  While you can write to this directory files there are not guaranteed to persist and may be deleted on deploy.
 
-For a MariaDB database, the command `platform db:size` will give approximate disk usage as reported by MariaDB.  However, be aware that due to the way MySQL/MariaDB store and pack data this number is not always accurate, and may be off by as much as 10 percent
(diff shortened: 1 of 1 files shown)


src/development/troubleshoot.md
```diff
@@ -74,6 +74,35 @@ The memory usage of your container exceeds the limit, the kernel thus kills the
 * Use [sizing hints](https://docs.platform.sh/languages/php.html#php-worker-sizing-hints) to reduce the amount of PHP workers which reduces the memory footprint.
 * Upgrade your subscription on Platform.sh to get more computing resources. To do so, log into your [account](https://accounts.platform.sh) and edit the project.
 
+## Low disk space
+
+If you suspect you are running low on disk space on your application container, the easiest way to check it is to login using `platform ssh` and run the `df` command.  `df` has numerous options to tweak its output, but for just checking the available writeable space the simplest option is: `df -h -x tmpfs -x squashfs | grep -v /run/shared`
```
pjcdawkins: If you add `-a` you'll get to see all the mounts, not just the first on each volume, but then you have to filter out more:

```sh
df -a -h -x proc -x squashfs -x tmpfs -x devpts -x sysfs | grep -v /run/shared
```

Might seem a bit ridiculous describing that as the "simplest option" (when `df` theoretically shows the necessary % on its own).

Crell: Perhaps change it to "most direct"?  My first version used `df -haT` to get all possible data, but as @sruggier pointed out that's TMI for most people who just want to know how much disk mount  space they are using.

src/development/troubleshoot.md
```diff
@@ -19,7 +19,7 @@ git push platform <branch-name>
 
 ## HTTP 502 response
 
-If you receive HTTP 502 responses from your site, it means that your site is running out of workers, or the application process is crashing.  Here are the typical causes:
+If you receive HTTP 502 responses from your site, it means that your site is running out of workers or the application process is crashing.  Here are the typical causes:
```
Crell: This additional comma is actually wrong and a comma splice. 😄   Please remove it, then go ahead and merge this PR.

Crell: Oh, duh, I totally read that backwards for some reason.  We're good, please merge.

PR #750, opened 2018-01-26
Clarify Platform - HipChat Integration setup steps

jameswilson:
The HipChat administrative UI provides two ways to create API tokens.  Most other services we use work with the application-level tokens generated at url `https://mycompany.hipchat.com/admin/api`  however Platform.sh only works with room level tokens at url `https://mycompany.hipchat.com/admin/rooms/tokens/<ROOM-ID>`

This cleanup outlines these steps more clearly, and also adds missing optional parameters for `environments` and `exclude-environments` which appear to have been added since the documentation was originally written.

### DIFF
--- a/src/administration/integrations/hipchat.md
+++ b/src/administration/integrations/hipchat.md
@@ -3,14 +3,32 @@
 The [HipChat](https://hipchat.com) integration allows you to send
 notifications about your Platform.sh activity directly to HipChat.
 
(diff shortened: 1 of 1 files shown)


Crell:
Looks good.  Thanks!

jameswilson:
Thanks for the shoutout in the Feb newsletter that dropped today!

PR #753, opened 2018-01-29
Add example of supporting an SPA config file in a deploy hook.

Crell:


allgood2386:
Also found that make-config.js should actually be node make-config.js.

Here is my issue that I'm realizing for us the killer. Fundamentally our particular application isn't addressed in here.

Our grunt build basically NEEDS access to the route. 

So we have a chicken and egg problem. I need to do the following during build and have a writeable mount.

```
npm install
bower install
grunt build```

grunt build will fail due to a missing config.js

then during deploy we run 

```
node make-config.js
```
but this is too late for the application. I get that this presents a pain and isn't the idea behind building but this is the situation we are in. If the answer is, don't do this and if you do it's not supported that's fine and I get that, I'm not happy with this config as well, but if not, I'll provide some additional config if/when I get this working. I don't want to encourage a bad practice but I can't be the only person who has been handed an application like this?

**Comment by Crell** [comment]
Your situation is... unusual. 😄 For now let's target this PR at the more common case of just needing the one file.  As discussed in Slack your specific app is going to be its own animal.

Fixed the bugs noted above.

**Comment by pjcdawkins** [comment]
I think this needs a clearer use case (the example as it is would be better off using `/`)

**Comment by yobottehg** [comment]
I also tried to get this working and then saw that we can only do this in the deploy hook :)

It's not uncommon what @allgood2386 is saying. 
Basic example: 
- multi app
  - static page generator
  - laravel backend
- while building the static page in the build step you need to inject the routes into the builder for the backend.

I'm all for getting this in the docs but it's not enough for the most use cases.

**Comment by yobottehg** [comment]
@Crell thanks again for the talk at Drupal Camp Transylvania.

Here the packages i suggested for testing:

Backend:
https://github.com/contentacms/contenta_jsonapi

Frontend:
https://github.com/contentacms/contenta_vue_nuxt

As you can see here:
https://github.com/contentacms/contenta_vue_nuxt/blob/master/nuxt.config.js#L5

We can add a callback here which gives us the corresponding backend url. We just needed a function which returns the name deterministic.

If you have further questions just ask me.

Greetings,

Rainer

**Comment by yobottehg** [comment]
@JGrubb we have a workaround in place. We just added the master backend url as a project environment variable. Most pull requests don't need the separate api. If you need just change the backend url in the variables for the PR

**Comment by yobottehg** [comment]
I recently found this here:
https://github.com/systemseed/drupal_reactjs_boilerplate

I think what they are doing here should work flawlessy:
https://github.com/systemseed/drupal_reactjs_boilerplate/pull/7/files

They are doing it on circleci but it should also be possible in a build or deploy hook.

`platform project:set-remote $PLATFORMSH_PROJECT_NAME`
```
BACKEND_URL=$(platform route:get -P url https://admin.\{default\}/)
# Set variable without trailing / symbol.
platform vset env:BACKEND_URL ${BACKEND_URL%/}
```

Crell:
We're going to have to rethink this eventually, but for now I'm trying to clean up old threads.  I've saved the information here offline and am closing this thread.  Thank you to those who participated.

PR #755, opened 2018-01-30
document routes precedence

alexmadon:


src/configuration/routes.md
```diff
@@ -163,3 +163,35 @@ Here is an example to define a route that serves WebSocket:
...
+To have `https://www.domain1.com` redirected to `https://www.domain2.com` you would need to change the order of the rules like this:
+```yaml
+"https://www.{all}/":
+    type: upstream
+    upstream: "app:http"
+    
+"https://www.domain1.com":
+    type: redirect
+    to: "https://www.domain2.com"
+
```
ralt: Superfluous newline

src/configuration/routes.md
```diff
@@ -163,3 +163,35 @@ Here is an example to define a route that serves WebSocket:
...
+"https://www.{all}/":
+    type: upstream
+    upstream: "app:http"
+```
+
+when asking for the URL `https://www.domain1.com` there will be no redirect applied as the first half of the file would suggest but rather an application hit and the second rule will also match and override the fist rule.
+
+To have `https://www.domain1.com` redirected to `https://www.domain2.com` you would need to change the order of the rules like this:
+```yaml
```
ralt: Add a newline before this one please

damz:
This is backwards: routes generated by `{all}` do not override any existing route.

Crell:
With the new changes to `{all}` is this even still relevant?  If so, @alexmadon please continue.

PR #758, opened 2018-01-31
Adapt source for consistency with current conventions

Crell:
This started as just updating the path usage as we discussed (see the first two commits).  However, that involved touching a lot of files that were stupid old and still had hard word wraps, badly specified code samples, and other formatting fails.  So I decided to go ahead and fix All The Things, which I've been meaning to do for a while.

I'm pretty sure this doesn't conflict with any other outstanding PRs, but we should get it in quickly to avoid conflicting with future PRs.

Who's feeling adventurous? 😄

### DIFF
--- a/src/administration/users.md
+++ b/src/administration/users.md
@@ -2,59 +2,38 @@
 
 ##User roles
 
-Every Platform.sh user has a role which controls access and improves
-security on your project. Different roles are authorized to do
-different things with your applications, environments and users. You can
-use your collection of Roles to manage how users interact with
-Platform.sh.
+Every Platform.sh user has a role which controls access and improves security on your project. Different roles are authorized to do different things with your applications, environments and users. You can use your collection of Roles to manage how users interact with Platform.sh.
 
 At the project level:
 
--   **Project Administrator** - A project administrator can change
-    settings and execute actions on any environment.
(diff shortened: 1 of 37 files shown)


src/administration/users.md
```diff
@@ -2,59 +2,38 @@
...
-> SSH access to the environment. You can change this by [specifying
-> user types](/configuration/app/access.md) with ssh access.
+>
+> An environment contributor can push code to the environment and has SSH access to the environment. You can change this by [specifying user types](/configuration/app/access.md) with ssh access.
 
 > **note**
-> The project owner - the person licensed to use Platform.sh - doesn't have
-> special powers. A project owner usually has a project administrator role.
+>
+> The project owner - the person licensed to use Platform.sh - doesn't havespecial powers. A project owner usually has a project administrator role.
```
pjcdawkins: havespecial

Crell: I KNEW that would happen somewhere. Fixed.

src/configuration/app/storage.md
```diff
@@ -55,12 +55,12 @@ mounts:
     private:
         source: local
         source_path: private
```
pjcdawkins: Add quotes, possibly

PR #760, opened 2018-02-02
Nodejs nvm

xtfer:
Adds instructions to change the Node.js version.


src/SUMMARY.md
```diff
@@ -56,6 +56,7 @@
 
 * [PHP](languages/php.md)
 * [Node.js](languages/nodejs.md)
+  * [Changing the Node.js version](languages/nodejs-nvm.md)
```
Crell: Structural note: This file should be `languages/nodejs/nvm.md`.  That's how we're handling subdirectories in the docs.

That said, I'm uncertain if this is the best place for the file.  Most of the language pages are for using that language as the primary language, whereas this is talking about changing the node.js version when it's a secondary language.  I'm not sure what would be better, though.  Thoughts?

Crell:
Users are going to love this. 😄

src/SUMMARY.md
xtfer: Neither was I, I was hoping you would have an idea. 

My thoughts were that someone looking to change the NodeJS version would probably start at these pages, however its not strictly the App container, so I sub paged it.

PR #761, opened 2018-02-02
Adding Pagespeed in the services docs

shfranc:


src/configuration/services/pagespeed.md
```diff
@@ -0,0 +1,41 @@
...
+
+* 1.12
+
+## Setup
+
+In your ``.platform/services.yaml``:
+
+```yaml
+pagespeed:
+    type: pagespeed:1.12-rc
```
Crell: Hey @damz, any idea when this is going to become non-rc?

Crell: If we're documenting it as RC, then this page should say clearly in the header section that it's experimental/unsupported/beta/whatever.

src/configuration/services/pagespeed.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+The pagespeed proxy exposes an endpoint app and this endpoint uses the upstream relationship named app.
+
+In your .platform/routes.yaml, use pagespeed:http-app instead of app:http:
+
+```yaml
+https://{default}/":
+    type: upstream
+    upstream: "pagespeed:http-app"
```
Crell: http-app?  Is that the endpoint that Damien gave?  That's... very misleading, as I would expect it to be the HTTP connection to the app container.  Why's it not just `http`, like app containers and the experimental Varnish container?

src/configuration/services/pagespeed.md
```diff
@@ -0,0 +1,41 @@
...
+
+See the [PageSpeed docs](https://developers.google.com/speed/) for more information.
+
+## Supported versions
+
+* 1.12
+
+## Setup
+
+In your ``.platform/services.yaml``:
```
Crell: Only one set of `

src/configuration/services/pagespeed.md
```diff
@@ -0,0 +1,41 @@
...
+    configuration:
+        endpoints:
+            app: {}
+    relationships:
+        app: "myapp:http"
+```
+
+The pagespeed proxy exposes an endpoint app and this endpoint uses the upstream relationship named app.
+
+In your .platform/routes.yaml, use pagespeed:http-app instead of app:http:
```
Crell: `.platform/routes.yaml` should be backticked.

src/configuration/services/pagespeed.md
ralt: It's `http-<relationship to app>`.

src/configuration/services/pagespeed.md
```diff
@@ -0,0 +1,41 @@
...
+pagespeed:
+    type: pagespeed:1.12-rc
+    configuration:
+        endpoints:
+            app: {}
+    relationships:
+        app: "myapp:http"
+```
+
+The pagespeed proxy exposes an endpoint app and this endpoint uses the upstream relationship named app.
```
Crell: Technically it's not `app`, it's whatever your app container is named. `app` is just what most of the examples use.

This page should probably explain, at least a little, about how we're sticking an interstitial service between the router and the app.  That's a weird concept that we don't have anywhere else yet. 😄

src/configuration/services/pagespeed.md
```diff
@@ -0,0 +1,41 @@
+# PageSpeed Module
+
+PageSpeed is a family of 4 tools developped by Google Inc to optimize of websites performance.
```
Crell: The "Inc" is unnecessary.  Or if it is, it needs a comma before it and period after.  (English is weird.)

OriPekelman:
(do not merge until we are sure we don't want to tweak the container further before release).

src/configuration/services/pagespeed.md
OriPekelman: So this needs much clearer docs...

xtfer: Yes, so it it correct, but it is a bit confusing.

src/configuration/services/pagespeed.md
xtfer: Just unnecessary. Remove :)

src/configuration/services/pagespeed.md
xtfer: @Crell We haven't documented that, even in draft, for Varnish, yet?

src/configuration/services/pagespeed.md
```diff
@@ -0,0 +1,37 @@
...
+    relationships:
+        <relationship to your app>: "<name of your app>:http"
+```
+
+This is basically a proxy service that sits between our router (the web server) and your application. It exposes an end-point (here confusingly named `app`) and has a relationship to the application also called app.
+
+Usually in your  `.platform/routes.yaml` you would have an `upstream` composed of the name of the application (the one you put in `.platform.app.yaml` and the protocol (currently always `http`). To use the PageSpeed service we make the router use it as its upstream rather than directly the app. In this example: `pagespeed:http-<relationship to your app>` instead of `app:http`. 
+
+```yaml
+https://{default}/":
```
pheadeaux: Note: A starting `"` is missing

src/configuration/services/pagespeed.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Setup
+
+To describe the relationship between Pagespeed and your application, you need to set up an endpoint and a relationship to your application. Both will carry the same name, `<relationship to app>` in the services.yaml example.
+
+Example of a `.platform/services.yaml`, where you need to fill `<relationship to your app>` and `<name of your app>` with your own:
+
+```yaml
+pagespeed:
+    type: pagespeed:1.12
```
pheadeaux: Quotes Missing: `type: "pagespeed:1.12"`

PR #763, opened 2018-02-05
Cloudflare page

JGrubb:


src/golive/steps/cloudflare.md
```diff
@@ -0,0 +1,17 @@
+# (Optional) Cloudflare configuration
+
+One of the main features that a modern DNS provider needs to have in order to work with Platform.sh is somethat that's colloquially known as "CName Flattening".  This solves an age-old problem (in internet years) - being able to point your "root domain" to a domain name rather then an IP address.  [This post](https://blog.cloudflare.com/introducing-cname-flattening-rfc-compliant-cnames-at-a-domains-root/) explains it well.
+
+## Prevent the infinite redirect loop
+
+Cloudflare offers a bit of a relief from the headache of setting up HTTPS behind a CDN with "flexible SSL".  That means that your site can use Cloudflare's SSL cert to encrypt the traffic from user to Cloudflare (remember, cloudflare is sitting in between your users and your website's server).  Traffic from Cloudflare to your website then travels unencrypted over plain old HTTP.  This is "suboptimal", but it does alleviate some of the attack vectors on your users.
```
Crell: Why is "suboptimal" in quotes?  It is suboptimal, period.

src/golive/steps/cloudflare.md
```diff
@@ -0,0 +1,17 @@
...
+
+```
+Cloudflare's "Flexible SSL" option
+        HTTPS                   HTTP
+User ----->----> Cloudflare ----->----> "Origin" server (your website)
+```
+
+The effect of this setting on your site is that while you may request a page over HTTPS, the origin (your platform.sh application) receives a request over HTTP, which it then attempts to redirect to HTTPS and accidentally kicks off a redirect loop.
+
+The solution - _always enable "Full SSL" in Cloudflare_.
```
Crell: I don't follow this section.  It talks about "flexible SSL", but the title is about infinite loops.  But the end conclusion is "enable Full SSL".  None of this seems related to the rest.

I think it will work better to lead with the "correct" configuration.  Then you can explain what the fail case is that you're trying to avoid.  The title should also be more action-oriented rather than something about a loop that isn't explained until the very end of the section.

src/golive/steps/cloudflare.md
```diff
@@ -0,0 +1,35 @@
...
+```
+# Cloudflare's Flexible SSL option
+		   https                       http
+User <---------------> Cloudflare <-------------> Platform.sh
+
+```
+
+This will cause all traffic from Cloudflare to your project to be redirected to HTTPS, which will set off an endless loop as HTTPS traffic will be presented as HTTP to your project no matter what.
+
+**Always use "Full SSL"**
```
pjcdawkins: `Full (strict)` should normally be fine too - but it will fail if/when the project's certificate expires. Provided your certificates don't expire, `Full (strict)` would be good and arguably better than `Full`, so I feel it deserves a mention?

pjcdawkins: And... `Flexible` would also be fine if you define an `http` upstream on your project (instead of the default `http`->`https` redirect) and then you can optionally use Cloudflare's https redirection ("Always use HTTPS").

IMO the absolutism of this statement is at odds with the verbosity of the rest of our docs.

src/golive/steps/cloudflare.md
```diff
@@ -0,0 +1,35 @@
+# (Optional) Cloudflare configuration
+
+One of the main features that a modern DNS provider needs to have in order to work well with Platform.sh is somethat that's colloquially known as "Cname Flattening".  This solves the problem of being able to point your "root domain" (example.com) to a domain name (cname) rather than an IP address (A record).  [This post](https://blog.cloudflare.com/introducing-cname-flattening-rfc-compliant-cnames-at-a-domains-root/) explains it well.
```
Crell: Throughout the page, "CNAME" should be in all-caps.

src/golive/steps/cloudflare.md
```diff
@@ -0,0 +1,35 @@
...
+
+In order to correctly point DNS to your Platform.sh project, you'll need at the very least the master environment cname, in other words the domain of your site before you add a custom domain in the project admin UI (or otherwise in the CLI).  It will look something like 
+
+TODO - get example
```
Crell: I was just discussing this with @Fidelix in channel.  Shouldn't this point back to the main GoLive docs where we talk about getting that domain?  (Which is currently harder than it should be, but if we're going to change anything in docs for it we should do so centrally.)

src/golive/steps/cloudflare.md
```diff
@@ -0,0 +1,35 @@
...
+
+Assuming that you are using both a www. subdomain as well as the bare domain, you'll want to point both of those DNS entries to the same place. Whether you choose the bare domain version or the www subdomain doesn't make any practical difference, as they both will reach Platform.sh and be handled correctly.
```
Crell: www. should have backticks around it.

src/golive/steps/cloudflare.md
```diff
@@ -0,0 +1,35 @@
...
+Cloudflare also makes it very simple to use their free TLS/SSL service to secure your site via HTTPS, while also being behind their CDN if you so choose.  If you decide to use Cloudflare's CDN functionality in addition to their DNS service, you should be sure to choose the "Full SSL" option in the Cloudflare admin.
+
+This means that traffic to your site is encrypted from the client (browser) to Cloudflare's servers using their certificate, and also between Cloudflare's servers and your project hosting here at Platform.sh, mostly like using your project's Let's Encrypt certificate.
+
+```
+# Cloudflare's Full SSL option
+		   https                       https
+User <---------------> Cloudflare <-------------> Platform.sh
+
+```
```
Crell: No blank line inside the end of the code sample.  (Same below, as well.)

src/golive/steps/cloudflare.md
Crell: Since there will always be an SSL cert offered from the Platform.sh side, shouldn't we be documenting based on that assumption, and then including a warning "yo, don't let your LE cert expire"?  I can't see a reason we should ever let users run Flexible SSL...

Fidelix: @Crell we may not be able to think about a scenario now, but that doesn't mean it doesn't exist. We shouldn't take the away the option from users when we can afford to do so.

Fidelix: I just thought of one scenario:
If you have a fully working application on Platform.sh, using HTTP-only, but you want to do a trial of a CDN to see how your application behaves using HTTPS without actually changing the code on Platform.sh.

Crell: I'm open to suggested alternative wordings here to encourage without demand.  I don't know Cloudflare well enough to write that.

Crell:
I think this is good to go as soon as the question around Full SSL is resolved.  I don't know Cloudflare well enough to write that guideline.  @pjcdawkins @Fidelix, please assist.

src/golive/steps/cloudflare.md
Crell: Revised the wording to soften it a bit.  If this is acceptable now hit the Big Green Button(tm).

PR #764, opened 2018-02-06
Add page detailing why not to use Backup and Migrate for Drupal.

Crell:
Because reasons. 😄


xtfer:
I would seriously consider putting this in the FAQ or a Known Issues section. Having a page for it seems like overkill.

PR #766, opened 2018-02-10
Fix typo - rename Redis to Memcached

boyanborisov:


Crell:
Oops!  That's what I get for copy-paste-modifying.  Thanks!

PR #769, opened 2018-02-12
Let's Encrypt auto-renewal

Crell:
We tell people they can do this, so let's document it.

While I was at it, I also added both cron setups to the go-live instructions since they're not easy to find otherwise.

### DIFF
--- a/src/configuration/routes/https.md
+++ b/src/configuration/routes/https.md
@@ -147,7 +147,7 @@ crons:
         spec: '0 10 3 * *'
         cmd: |
             if [ "$PLATFORM_BRANCH" = master ]; then
-                platform variable:set _redeploy "$(date)" --no-wait
+                platform variable:set _redeploy "$(date)" --yes --no-wait
             fi
 ```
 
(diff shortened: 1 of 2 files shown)


src/configuration/routes/https.md
```diff
@@ -129,3 +131,30 @@ tls:
...
+Once the CLI is installed and an API token configured you can add a cron task to run once a month and set a placeholder variable.  For example:
+
+```yaml
+crons:
+    renewcert:
+        # Force a redeploy at 10 am (UTC) on the 3rd of every month.
+        spec: '0 10 3 * *'
+        cmd: |
+            if [ "$PLATFORM_BRANCH" = master ]; then
+                platform variable:set _redeploy "$(date)" --no-wait
```
pjcdawkins: Use `--yes` to turn off interactivity (including update checks, and any confirm steps that the command may ask)

Crell:
Changes seem fine, although a few are off topic.  (Meh, I'm not usually that picky about those if they're in their own commits.)  I'm good to merge here if you are.

PR #772, opened 2018-02-15
Update the robo-blocker now that we use a header, not a robots.txt file

Crell:


src/frameworks/drupal7/faq.md
```diff
@@ -19,7 +19,7 @@ deployment hooks to automatically run the updates.
 
 ## How can I override the default robots.txt?
```
pjcdawkins: There is no longer a default robots.txt, so this question and probably the first sentence of the answer need to change.

Crell: Revised to clarify that it's really only about where to put `robots.txt` in the `drupal` build mode.

PR #773, opened 2018-02-15
Don't use Symfony-specific syntax in PHP examples

Crell:
This is just Rabbit MQ for now. Others will come in this PR once the pattern is agreed upon.


Fidelix:
megusta this PR.

PR #775, opened 2018-02-20
typo

mrkschan:


Crell:
Is the _ more common in the docs right now?  We never did discuss/decide if we wanted to include that by default or not.

PR #776, opened 2018-02-21
Clean up Dropbox instructions

OriPekelman:
Current docs are wrong on mounts. Clarified. Simplified. But most importantly made functional.


JGrubb:
I think I might rather delete it.  It seems only to work for php 7.0, and no others.

OriPekelman:
Please leave this one for the moment .. there are a bunch of enhancements here ... I need to find time to merge..

PR #777, opened 2018-02-21
Add notice on no hidden mounts

OriPekelman:


Crell:
Closing as we determined this isn't actually true, they just need a leading /, per #778

PR #780, opened 2018-02-25
Improve Drush information

xtfer:
This clarifies when you should need to install Drush, and removes some references to it being available "automatically" which seems to confuse people.

I also copied the revised information from D8 over to D7 so that is all pretty much the same, with the exception of the make stuff.


src/frameworks/drupal7/drush.md
```diff
@@ -10,11 +10,41 @@ commands at [Drush.org](http://www.drush.org). If you have never used
...
+
+## Installing Drush
+
+## Install Drush on Platform.sh-provided Drupal examples
+
+Platform.sh's Drupal templates have Drush installed automatically via Composer and should not require you to add it.
+
+## Install Drush in custom projects
+
+### Make Drush a project requirement
```
pjcdawkins: Could this also describe using the `dependencies` section of the . platform.app.yaml?

xtfer: Well, do we have a preference for dependencies over composer.json?

src/frameworks/drupal7/drush.md
```diff
@@ -10,11 +10,41 @@ commands at [Drush.org](http://www.drush.org). If you have never used
...
+### Add Drush to the project's $PATH
+
+Add a new file named `.environment` to the root of your your project's git repository with this code:
+```bash
+# Statements in this file will be executed (sourced) by the shell in SSH
+# sessions, in deploy hooks, in cron jobs, and in the application's runtime
+# environment.
+
+# Allow executable app dependencies from Composer to be run from the path.
+export PATH=/app/vendor/bin:$PATH
```
MacWeber: Let's use the same code from the example that lets the use of the same `.environment` file in PE.

src/frameworks/drupal8/drush.md
```diff
@@ -62,6 +50,24 @@ Add a new file named `.environment` to the root of your your project's git repos
 export PATH=/app/vendor/bin:$PATH
```
MacWeber: Let's use the same code from the example that lets the use of the same `.environment` file in PE: https://github.com/platformsh/platformsh-example-drupal8/blob/master/.environment#L8-L11

src/frameworks/drupal7/drush.md
Crell: On D8 we favor composer.  On D7, we're using dependencies.

src/frameworks/drupal7/drush.md
Crell: Concur.

src/frameworks/drupal7/drush.md
```diff
@@ -32,15 +62,19 @@ $ drush
...
+> -   [Drush repository on Github](https://github.com/drush-ops/drush)
 
 ## Use drush aliases
 
 ### Create Drush aliases
 
-[Drush aliases](http://drush.readthedocs.org/en/master/usage/index.html#site-aliases) make it easy to manage your development websites. Here's an example of a [Drush alias file](https://github.com/drush-ops/drush/blob/master/examples/example.aliases.drushrc.php).
+[Drush aliases](http://drush.readthedocs.org/en/master/usage/index.html#site-aliases) make it easy to manage your development websites. Here's an
+example of a [Drush alias
+file](https://github.com/drush-ops/drush/blob/master/examples/example.aliases.drushrc.php).
```
Crell: Why the extra newlines here?

src/frameworks/drupal7/drush.md
```diff
@@ -51,6 +85,8 @@ Aliases for My Site (tqmd2kvitnoly):
     @my-site.sprint1
 ```
 
+To recreate existing aliases, or after pushing a new branch via git to create the new alias, run `platform drush-aliases -r`
```
Crell: Oh dear god thank you.  We should have added that a long time ago.

src/frameworks/drupal7/drush.md
xtfer: Keeps it interesting. (I'll remove them)

src/frameworks/drupal8/drush.md
```diff
@@ -20,54 +12,72 @@ Repository](https://github.com/drush-ops/drush#description)
...
-$ drush
+$ composer require drush/drush
 ```
+Then, commit and push.
 
-And see a list of available commands.
+Drush will then be available at `vendor/bin/drush`, in the exact same version on your local system and on Platform.sh.
 
-> -   [Drush repository on Github](https://github.com/drush-ops/drush)
+### Install Drush in custom projects using _Build Dependencies_
```
Crell: Do we even want to document this option?  For Drupal 8 users should pretty much always use the composer approach.

Crell:
I'm confused.  Why are you folding both Drush pages into a single page?  The installation process is different enough that seems like it only complicates matters.

xtfer:
Most of the instructions are identical? Would you prefer to maintain two largely identical documents? Up to you.

src/frameworks/drupal8/drush.md
xtfer: What if they aren't using Composer?

Crell:
I decided to go ahead here, but it needed updates with master.  I also removed more duplication (if we're going down that route, go all the way) and switched to embedding example files rather than copying them in.

I think this is mergeable.  @pjcdawkins @MacWeber ?

src/frameworks/drupal7/drush.md
```diff
@@ -1,82 +1,22 @@
...
-Repository](https://github.com/drush-ops/drush#description)
+Drush is a command-line shell and scripting interface for Drupal, a veritable Swiss Army knife designed to make life easier for those who spend their working hours hacking away at the command prompt. 
 
-You can use the CLI to set up Drush aliases, to easily run Drush commands on
-specific remote Platform.sh environments.
+You can use the CLI to set up Drush aliases, to easily run Drush commands on specific remote Platform.sh environments.
 
-> **note**
-> Platform's CLI requires **Drush 6 or greater**.
+See the [documentation on Drush in Drupal 8](frameworks/drupal8/drush.html) for installation, drush aliases, and other general information.  The installation procedure is the same for both Drupal 7 and 8.
```
pjcdawkins: I think the link needs a `/` prefix, is that right?

Crell: Yep.  Fixed.

PR #781, opened 2018-02-26
Access control supports CIDR

mrkschan:


Crell:
Just mentioning CIDR on its own is not especially helpful for most people.  It's also not entirely accurate.  Rather, we should say that we support "CIDR format", and give a super-duper-brief explanation of what it means.  (And then link to Wikipedia for the rest.)

mrkschan:
@Crell how's this now?

Crell:
Updated to include an example of CIDR format, then merged.

PR #782, opened 2018-02-26
No longer need to add the gitlab webhook manually

mrkschan:
The latest CLI will create the webhook when we have enough permission to do so.

### DIFF
--- a/src/administration/integrations/gitlab.md
+++ b/src/administration/integrations/gitlab.md
@@ -43,16 +43,18 @@ Optional parameters:
 
 ### 3. Add the webhook
 
-The previous command, if succesful should output the configuration of the integration. The last element would look like:
+The previous command, if successful should output the configuration of the integration. The last element would look like:
 
 ```
 | hook_url | https://{region}.platform.sh/api/projects/{projectid}/integrations/{hook_id}/hook |
 ```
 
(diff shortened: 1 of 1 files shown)


src/administration/integrations/gitlab.md
```diff
@@ -49,7 +49,7 @@ The previous command, if succesful should output the configuration of the integr
 | hook_url | https://{region}.platform.sh/api/projects/{projectid}/integrations/{hook_id}/hook |
 ```
 
-Now, copy the hook URL.
+If you don't see the CLI outputs `Webhook created successfully`, you need to add the webhook manually. Copy the hook URL and continue the below.
```
Crell: "continue the below" is clumsy grammar.  What's the change here exactly?

pjcdawkins: The change is that the CLI will automatically ensure webhooks, if it can (i.e. if the token is correct and has enough permissions).

Suggestion:

> If you see the message "Failed to read or write webhooks", you will need to add a webhook manually:
> 
> - Copy the hook URL shown in the message.
> - Go to your GitLab repository and click `Settings > Integrations`.
> - Paste the hook URL. In the Triggers section choose `Push events`, `Tag push events` and `Merge Request events`. Click on `Add webhook`.

mrkschan:
@Crell @pjcdawkins revised.

PR #786, opened 2018-02-27
Add new variable options

Crell:
@pjcdawkins I'm assuming what the CLI toggles will be named.  This is blocked until you release a CLI version with the updated flags.  Let me know if I should change the CLI switches at that time, or just do it yourself before merging.  I'm good either way.

### DIFF
--- a/src/development/variables.md
+++ b/src/development/variables.md
@@ -6,12 +6,12 @@ Platform.sh allows a high degree of control over both the build process and the 
 
 There are four different types of variable: 
 
+* application-provided variables
 * user-provided project variables
 * user-provided environment variables
-* application-provided variables
 * Platform.sh-provided environment variables
 
-All of those may be simple strings or base64-encoded JSON-serialized values.  In case of name collisions, Platform.sh-provided values override user-provided environment variables, which override project-level variables, which override application-provided variables.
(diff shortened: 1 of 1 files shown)


pjcdawkins:
I'm proposing some bigger CLI changes to support this:
https://github.com/platformsh/platformsh-cli/pull/694

1. Old variables commands will still work, hopefully as before, but they will be hidden from the command list.
2. The old, hidden `variable:set` (`vset`) command will not support the new features.
3. The `variable:create` command will support something like:
  ```sh
  platform variable:create --name paypal_id --sensitive true
  ```
  But it also presents an interactive form for filling out any of the options not specified on the command-line (if in interactive mode, otherwise they are set to defaults).
4. The `variable:update` command will support changing values or other flags like:
  ```sh
  platform variable:update paypal_id --inheritable false
  ```
5. There's a new `variable:list` command, aliased to `variables`.
6. The `variable:` commands can be used to list, view, update, or delete variables at both the project and environment levels. If you don't specify a `--level` (`project` or `environment` or an abbreviation), it will look to see if a variable exists on either level and select that. If a variable exists at _both_ levels, it will exit with an error (and ask you to specify a `--level`).

pjcdawkins:
I'll make a PR for above changes if you're happy with that @Crell

Crell:
I would have said just take over this PR, but either way...

So should we close this one in favor of the other?

Crell:
I would want to keep the first commit from this branch.  I can cherry-pick that out if we decide to just toss the PR.

pjcdawkins:
Yes sorry I should have continued this PR. The first commit is wanted, I agree.

PR #788, opened 2018-02-28
Switch all examples to use the new redeploy command.

Crell:
Patrick, please merge as soon as the CLI is released.


Crell:
@pjcdawkins Are we clear for launch on this one then, or is it still blocked by something?

pjcdawkins:
@Crell it requires platform.sh [git] version 2.8.x, which is not available on all projects yet

OriPekelman:
I believe now it is ? (2.8 deployed everywhere ..)

PR #792, opened 2018-03-05
Tweak Blackfire page

fabpot:
Blackfire being now independant, the SensioLabs mention should be removed.
Removed also the companion version as it's not related to platform.sh (and it is automatically updated client-side anyway).
Added a mention of the Blackfire web browser companion.


Crell:
Merci bien!

PR #805, opened 2018-03-19
Fix cookie example so it doesn't break Drupal

pjcdawkins:
Even though it doesn't say it's a Drupal example, it looks like one, so people will want to copy and paste it into a Drupal configuration anyway. The current example breaks Drupal sites on https.


Crell:
Makes sense to me.

PR #816, opened 2018-04-09
Python gunicorn example fix

Oc7at:
Hello,

Gunicorn is used with the bind option which cannot only take a port - it has to be a whole address (either `$(HOST)`, `$(HOST):$(PORT)`, or `unix:$(PATH)`) ; using `-b $PORT` makes Gunicorn calculate the port as a host (and breaks things), please see [this page](http://docs.gunicorn.org/en/latest/run.html#commonly-used-arguments).

Lucas

### DIFF
--- a/src/languages/python.md
+++ b/src/languages/python.md
@@ -36,7 +36,7 @@ In this example, we use Gunicorn to run our WSGI application.  Configure the `.p
    ```yaml
    web:
      commands:
-       start: "gunicorn -b $PORT project.wsgi:application"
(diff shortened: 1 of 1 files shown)


Crell:
Shouldn't we be providing a real IP?  Or localhost?  (I don't know from Gunicorn so I can't speak to what is "right" here.)

mrkschan:
Both localhost / 0.0.0.0 will work as it's talking to Nginx in the app
container.

PR #818, opened 2018-04-17
PHP max children

Crell:
A lot of users are reporting the "hit child limit" error from their logs and unsure what to do about it.  So let's answer them and improve the instructions for what to do with that knowledge.

This is a precursor to reorganizing the whole PHP section and adding an optimization section, but one step at a time...

### DIFF
--- a/src/development/troubleshoot.md
+++ b/src/development/troubleshoot.md
@@ -76,7 +76,7 @@ You may see a line like the following in the `/var/log/app.log` file:
 
 `WARNING: [pool web] server reached max_children setting (2), consider raising it`
 
-That indicates that the server is receiving more concurrent requests than it has PHP processes allocated to, which means some requests will have to wait until another finishes.  In this example there are 2 PHP processes that can run concurrently.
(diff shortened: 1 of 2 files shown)


src/development/troubleshoot.md
```diff
@@ -64,12 +64,25 @@ If your PHP process is killed by the kernel, you should see log lines in `/var/l
...
 * Use [sizing hints](https://docs.platform.sh/languages/php.html#php-worker-sizing-hints) to reduce the amount of PHP workers which reduces the memory footprint.
 * Upgrade your subscription on Platform.sh to get more computing resources. To do so, log into your [account](https://accounts.platform.sh) and edit the project.
 
+## PHP "server reached max_children" error in logs
+
+You may see a line like the following in the `/var/log/app.log` file:
+
+`WARNING: [pool web] server reached max_children setting (2), consider raising it`
+
+That indicates that the server is receiving more concurrent requests than it has PHP processes allocated to, which means some requests will have to wait until another finishes.  In this example there are 2 PHP processes that can run concurrently.
```
JGrubb: > That indicates that the server is receiving more concurrent requests than it has PHP processes allocated to

maybe "allocated to _handle_"?  "allocated to," don't read good.

PR #821, opened 2018-04-18
change index to different structure

Crell:
Making a PR for this branch so that we can discuss it.

For an overview of the structure see: 

https://github.com/platformsh/platformsh-docs/blob/alternative-structure/src/SUMMARY.md


.idea/modules.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Crell: @xtfer, you forgot to add .idea to your global .gitignore file. :smile:  Please rebase this out.

Crell:
@xtfer Interesting.  Can you elaborate on your reasoning here?  Some of it I can see the use of, others less so.

.idea/modules.xml
xtfer: Whoops. Probably the result of changing machines.

xtfer:
The original intent was to simply merge Development and Administration, as we discussed, however as I dug further I found some things which seemed logical while I was refactoring (they may not be, but they did at the time)...

_I've tried to expose at the top level the features that people might be looking for_, since scanning the top level menu is the only logical way to find information (other than search). Hence, Services gets its own section, integrations gets its own section.

Everything about using Platform.sh has gone in a single "Working with Platform.sh" section, largely alphabetized for scan-ability but grouped by concept (account concepts, configuration concepts, everything else).

Some pages were merged or moved based on their content as well. This could go further, for example the Fastly page could go under Integrations.

Other sections I didn't touch but should be reviewed in this model include the Getting Started section and Resources.

PR #822, opened 2018-04-23
update data retention to include google analytics

rinchen:
COM-372.  "Google analytics has new GDPR controls and the minimum amount we can keep analytics data is 14 months. We need to articulate this in our privacy policy. Add something like "we are going to keep analytics data for 14 from the time of last use"

### DIFF
--- a/src/administration/data-retention.md
+++ b/src/administration/data-retention.md
@@ -40,4 +40,4 @@ These "tombstone" backups are retained for between _7 days_ and _6 months_.
 
 ## Analytics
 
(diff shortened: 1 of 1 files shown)


src/administration/data-retention.md
```diff
@@ -38,3 +38,6 @@ When a project is deleted Platform.sh takes a final snapshot of active environme
 
 These "tombstone" backups are retained for between _7 days_ and _6 months_.
 
+## Analytics
+
+Platform.sh uses Google Analytics on various web pages, and therefore Google Analytics will store collected data for a period of time.  We have configured our Google Analytics account to store data for 14 months from the time you last accessed our site, which is the minimum Google allows.
```
Crell: Let's bold "14 months", since we're bolding all other timeframes in this page.

Crell:
@damz Did you want to object here?  It sounded like you had concerns in Slack.  If I don't hear otherwise I'll merge this in a day or two.

PR #823, opened 2018-04-27
Add a section on database access

Crell:
Rather than relying on bits scattered around in different places.


xtfer:
Shouldn't information on accessing services be in the relevant Services section? e.g. "Accessing Services". This page is not named in a way that suggests the database connectivity will be part of it.

pjcdawkins:
Similarly, "Accessing the database" implies there always is one.

Crell:
Turns out the Services page already had half of this information, so I folded the two together onto the Services main page.

PR #827, opened 2018-05-07
Drop unrelated section

mrkschan:
php.access.log has nothing to do with stuck build / deployment.


Crell:
I think the intent there was another "long running tasks block deploy" example, but yeah, if your page takes long enough that you'll notice from deploy you've got bigger problems. :smile:

PR #829, opened 2018-05-09
Remove NodeJS 4 from supported versions

xtfer:
Node 4 is EOL. See https://medium.com/the-node-js-collection/april-2018-release-updates-from-the-node-js-project-71687e1f7742.


Crell:
We should *not* remove Node 4 from the docs unless we're also removing the container images.  Instead we should add an "unsupported" section, like we have for PHP.

Please revert and/or update with an unsupported section.

PR #834, opened 2018-05-16
Added a leading forward slash

michielkenis:
This is required to create the writable mount in order to properly run the Dropbox installer.


pjcdawkins:
Thanks @michielkenis!

PR #835, opened 2018-05-17
Extra headsup in used version

michielkenis:
Added a small footnote to alert users to use the correct php-fpm (if not, apps will keep crashing).


Crell:
Merci, @michielkenis!

PR #842, opened 2018-05-31
Add support form MongoDB v3.6

gimudia:
Add support form MongoDB v3.2, v3.4 and v3.6 images

### DIFF
--- a/src/configuration/services/mongodb.md
+++ b/src/configuration/services/mongodb.md
@@ -7,6 +7,7 @@ For more information on using MongoDB, see [MongoDB's own documentation](https:/
 ## Supported versions
 
 * 3.0
+* 3.6
 
 ## Relationship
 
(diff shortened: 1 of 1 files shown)


src/configuration/services/mongodb.md
```diff
@@ -115,3 +118,7 @@ mongodump --port 30000 -u main -p main --authenticationDatabase main --db main
 As with any other shell command it can be piped to another command to compress the output or redirect it to a specific file.
 
 For further references please see the [official mongodump documentation](https://docs.mongodb.com/manual/reference/program/mongodump/#bin.mongodump).
+
+## Upgrading
```
ricardokirkner: I would also mention that upgrading only consists of updating the image version tag, and no further steps should be necessary, provided (this is important) the application does not make use of any version specific features (ie, when upgrading to 3.2, the app is not using any 3.0 specific features that are not compatible with 3.2, and so on).

PR #855, opened 2018-06-18
Troubleshooting cleanup

Crell:
There's some duplication in the troubleshooting guide, and it's a bit disorganized.  So let's organize it.

Open to suggestions on how to better order things here, too.  Maybe put the PHP section at the end?

### DIFF
--- a/src/languages/php/fpm.md
+++ b/src/languages/php/fpm.md
@@ -49,9 +49,11 @@ In the fifth column, you'll see the peak memory usage that occurred while each r
 
 A good way to determine an optimal request memory is with the following command:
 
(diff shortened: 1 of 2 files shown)


src/languages/php/fpm.md
```diff
@@ -49,9 +49,11 @@ In the fifth column, you'll see the peak memory usage that occurred while each r
 
 A good way to determine an optimal request memory is with the following command:
 
-`cat /var/log/php.access.log | awk '{print $6}' | sort -n | uniq -c`
+```
+grep $(date +%Y-%m-%dT%H --date='-7 days') /var/log/php.access.log | awk '{print $6}' | sort -n | uniq -c
```
xtfer: We should add a CLI command for this

PR #856, opened 2018-06-18
Add a PHP performance tuning guide.

Crell:
I've wanted to add one of these for a while.  Finally got around to it.

There's other non-PHP stuff that we could document as well but this is a start.

### DIFF
--- a/src/languages/php/tuning.md
+++ b/src/languages/php/tuning.md
@@ -42,10 +42,10 @@ The most important values to set are
 To determine how many files you have, run this command from the root of your application:
 
 ```bash
-find . -type f -print | grep php | wc -l
+find . -type f -name '*.php' | wc -l
 ```
 
-That will report the number of files in your file tree that contain the string "php" in their name.  That may not be perfectly accurate (some applications have PHP code in files that don't end in `.php`, some files may contain "php" in their name but not actually be PHP code, etc.) but it's a reasonable approximation.  Set the `opcache.max_accelerated_files` option to a value slightly higher than this.
+That will report the number of files in your file tree that end in `.php`.  That may not be perfectly accurate (some applications have PHP code in files that don't end in `.php`, it may not catch generated files that haven't been generated yet, etc.) but it's a reasonable approximation.  Set the `opcache.max_accelerated_files` option to a value slightly higher than this.  Note that PHP will automatically round the value you specify up to the next highest prime number, for reasons long lost to the sands of time.
 
 Determining an optimal `opcache.memory_consumption` is a bit harder, unfortunately, as it requires executing code via a web request to get adequate statistics.  Additionally, pushing new code clears the opcache so the statistics will be non-representative for a while.
 
--- a/src/languages/php/tuning.md
+++ b/src/languages/php/tuning.md
@@ -16,6 +16,8 @@ Although not PHP-specific, a common source of performance issues is a misconfigu
 
 You will also need to ensure that your application is sending the correct `cache-control` header.  The router cache will obey whatever cache headers your application sends, so send it good ones.
 
+Static assets cache headers are set using the `expires` key in `.platform.app.yaml`.  See the [`web.locations`](/configuration/app/web.md#locations) documentation for more details.
+
 ## Optimize the FPM worker count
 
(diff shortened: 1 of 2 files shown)


src/languages/php/tuning.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Upgrade to PHP 7.1+
+
+There is very little purpose to trying to optimize a PHP application on PHP 5.  PHP 7 is generally twice as fast and uses half as much memory as PHP 5, making it unquestionably the first step to take when trying to make a PHP-based site run faster.
+
+To change your PHP version, simply change the `type` key in your `.platform.app.yaml` to the desired PHP version.  As always, test it on a branch first before merging to `master`.
+
+## Ensure that the router cache is properly configured
+
+Although not PHP-specific, a common source of performance issues is a misconfigured cache.  The most common issue is not whitelisting session cookies, which results in a site with any cookies at all, including from analytics tools, never being cached.  See the [router cache](/configuration/routes/cache.md) documentation, and the [cookie entry](/configuration/routes/cache.md#cookies) specifically.
```
ralt: Maybe mention the cache settings of the `.platform.app.yaml` for static assets too?

src/languages/php/tuning.md
```diff
@@ -0,0 +1,110 @@
...
+
+PHP-FPM reserves a fixed number of simultaneous worker processes to handle incoming requests.  If more simultaneous requests are received than the number of workers then some requests will wait.  The default worker count is deliberately set rather conservative but can be improved in many cases.  See the [PHP-FPM sizing](/languages/php/fpm.md) page for how to determine and set a more optimal value.
+
+If your `/var/log/app.log` file contains a lot of entries like 
+
+```
+WARNING: [pool web] server reached max_children setting (2), consider raising it
+```
+
+That means your worker count is too low.  If you have already set an optimal worker size and still see that message frequently consider upgrading your project resources.
```
ralt: The common case is not a lack of resources, but requests being blocked on an external resource or something similar. Asking to upgrade the plan first is probably disingenuous.

src/languages/php/tuning.md
```diff
@@ -0,0 +1,110 @@
...
+
+The most important values to set are 
+
+* `opcache.max_accelerated_files`: The max number of files that the opcache may cache at once.  If this is lower than the number of files in the application it will begin thrashing and become less effective.
+* `opcache.memory_consumption`: The total memory that the opcache may use.  If the application is larger than this the cache will start thrashing and become less effective.
+
+To determine how many files you have, run this command from the root of your application:
+
+```bash
+find . -type f -print | grep php | wc -l
```
ralt: `find . -type f -name '*.php' | wc -l`

src/languages/php/tuning.md
```diff
@@ -0,0 +1,110 @@
...
+    php:
+        'opcache.max_accelerated_files': 22000
+        'opcache.memory_consumtion': 96
+```
+
+(Memory consumption is set in megabytes.)
+
+## Optimize your code
+
+It's also possible that your own code is doing more work than it needs to.  Profiling and optimizing a PHP application is a much larger topic than will fit here, but Platform.sh recommends enabling [Blackfire.io](/administration/integrations/blackfire.md) on your project to determine what slow spots can be found and addressed.
```
ralt: We don't have any guide about using xdebug profiler?

src/languages/php/tuning.md
Crell: Is it?  I've not seen external resource-dependencies causing a max_children issue.  (And telling people to upgrade is what we already do in about 15 other places in the docs.)

src/languages/php/tuning.md
Crell: No, xdebug's profiler is weak sauce in comparison.  Blackfire is superior in just about every way.  At some point I want to document using xdebug for remote step debugging but trying to document its profiler support is a waste of everyone's time.

src/languages/php/tuning.md
Crell: Added.

src/languages/php/tuning.md
Fidelix: ```I've not seen external resource-dependencies causing a max_children issue.```` I have - but it's rare. Most of the time the request is just not very well cached and is taking a bunch of seconds to reply, and has a bit of concurrency.

PR #861, opened 2018-06-22
Improve structure of Variables page

xtfer:
Adds ToC
Adds a table comparing variable types
Orders variable types to correspond to weighting
Cross-link
Add links to access environment variables in different supported languages


src/development/variables.md
```diff
@@ -1,17 +1,35 @@
...
+* [application-provided variables](/development/variables.html#application-provided-variables)
+* [user-provided project variables](/development/variables.html#project-variables)
+* [user-provided environment variables](/development/variables.html#environment-variables)
+* [Platform.sh-provided environment variables](/development/variables.html#platformsh-provided-variables)
+
+All of those may be simple strings or base64-encoded JSON-serialized values.  In case of name collisions, Platform.sh-provided values override user-provided environment variables, which override user-provided project-level variables, which override application-provided variables.  (That is, lower items in the list above take precedence.)
+
+## Comparison of variable types
+
+| Type          | Definer     | Scope       | Inheritance | Build | Runtime  |
```
Crell: Should the list above be removed in favor of just the table, and giving the table the links?  I feel like that would be less-redundant.

src/development/variables.md
```diff
@@ -1,17 +1,35 @@
...
+| ------------- | ----------- | ----------- | ----------- |:-----:|:--------:|
+| Application   | Application | Application | n/a         | Yes   | Yes      |
+| Project       | User        | Project     | n/a         | Yes   | Yes      |
+| Environment   | User        | Environment | Optional    | Yes   | Yes      |
```
Crell: Environment variables are not Build-time accessible.

src/development/variables.md
```diff
@@ -1,17 +1,35 @@
...
+| Platform.sh   | Pre-defined | Project     | n/a         | Some  | Yes      |
```
Crell: Platform.sh-defined variables are env-specific, or at least some are.  We should probably list it as Environment scope.

src/development/variables.md
```diff
@@ -1,17 +1,35 @@
...
+| Application   | Application | Application | n/a         | Yes   | Yes      |
+| Project       | User        | Project     | n/a         | Yes   | Yes      |
+| Environment   | User        | Environment | Optional    | Yes   | Yes      |
+| Platform.sh   | Pre-defined | Project     | n/a         | Some  | Yes      |   
+
+## Types
+
+### Application-provided variables
+
+It is also possible to [set Environment variables in code](/configuration/app/variables.md), using the `.platform.app.yaml` file.  These values of course will be the same across all environments and present in the Git repository, which makes them a poor fit for API keys and such.  This capability is mainly to define values that an application expects via an environment variable that should be consistent across all environments.  For example, the PHP Symfony framework has a `SYMFONY_ENV` property that users may wish to set to `prod` on all environments to ensure a consistent build, or it may be used to set [PHP configuration values](#php-specific-variables).
```
Crell: The first sentence should be reworded now that app-provided variables are the first item, meaning the "also" doesn't make sense.

src/development/variables.md
```diff
@@ -214,6 +226,15 @@ While the same command on the `feature-x` branch would produce:
...
 ```
 
+### In your application
+
+Check the individual documentation pages for accessing environment variables for your given application language...
+
+* [PHP: the getenv() function](http://php.net/manual/en/function.getenv.php)
+* [Node.js: the process.env object](https://nodejs.org/api/process.html#process_process_env)
+* [Python: the os.environ object](https://docs.python.org/3/library/os.html#os.environ)
+* [Ruby: the ENV accessor](https://ruby-doc.org/core-2.1.4/ENV.html)
```
Crell: We should do actual code samples here, methinks.  We have the code formatter for it.

Crell:
Overall, I quite like!  I think we could tighten it up a bit further, though, as above.

PR #865, opened 2018-06-26
Move slow request troubleshooting to out of FPM worker section

mrkschan:
PHP execution timeout is solely application related.
Slow requests have nothing to do with it.
While, having slow requests will eventually lead to occupying all FPM workers.

### DIFF
--- a/src/development/troubleshoot.md
+++ b/src/development/troubleshoot.md
@@ -140,15 +140,24 @@ WARNING: [pool web] server reached max_children setting (2), consider raising it
 
 That indicates that the server is receiving more concurrent requests than it has PHP processes allocated, which means some requests will have to wait until another finishes.  In this example there are 2 PHP processes that can run concurrently.
 
-Platform.sh sets the number of workers based on the available memory of your container and the estimated average memory size of each process.  There are two ways to increase the number of workers:
+The following command will identify the 20 slowest requests in the last hour, which can provide an indication of what code paths to investigate.
 
-* Adjust the [worker sizing hints](/languages/php/fpm.md) for your project.
-* Upgrade your subscription on Platform.sh to get more computing resources. To do so, log into your [account](https://accounts.platform.sh) and edit the project.
+```bash
+grep $(date +%Y-%m-%dT%H --date='-1 hours') /var/log/php.access.log | sort -k 4 -r -n | head -20
+```
 
+If you see that the processing time of certain request is slow while its CPU usage is high (e.g. taking more than 500ms while 65% of time spent on CPU), you may consider using a profiler like [Blackfire](/administration/integrations/blackfire.md) to troubleshoot the application performance issue. You may refer to [FPM access log explained](/languages/php/fpm.md) for details on the log field.
+
+Otherwise there is not any code path being slow, you may find the most visited pages and see if that can be cached and/or put behind a CDN. Refer to [how caching works](/configuration/routes/cache.md) for further instructions.
+
+Also, increasing the number of PHP workers would help. Platform.sh sets the number of workers based on the available memory of your container and the estimated average memory size of each process. There are two ways to increase the number of workers:
+
+* Adjust the [worker sizing hints](/languages/php/fpm.md) for your project when the processing time of the application is low.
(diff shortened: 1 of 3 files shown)


src/development/troubleshoot.md
```diff
@@ -140,15 +140,24 @@ WARNING: [pool web] server reached max_children setting (2), consider raising it
...
-Platform.sh sets the number of workers based on the available memory of your container and the estimated average memory size of each process.  There are two ways to increase the number of workers:
+The following command will identify the 20 slowest requests in the last hour, which can provide an indication of what code paths to investigate.
 
-* Adjust the [worker sizing hints](/languages/php/fpm.md) for your project.
-* Upgrade your subscription on Platform.sh to get more computing resources. To do so, log into your [account](https://accounts.platform.sh) and edit the project.
+```bash
+grep $(date +%Y-%m-%dT%H --date='-1 hours') /var/log/php.access.log | sort -k 4 -r -n | head -20
+```
 
+If you see that the processing time of certain requests is slow while its CPU usage is high (e.g. taking more than 1000ms while 65% of time spent on CPU), you may consider using a profiler like [Blackfire](/administration/integrations/blackfire.md) to troubleshoot the application performance issue.
```
Crell: How can a user tell how much of a request is spent on CPU?  I don't know how to determine that quickly myself.  (At least not without using an app-specific tool like Drupal Devel.)

src/development/troubleshoot.md
```diff
@@ -140,15 +140,24 @@ WARNING: [pool web] server reached max_children setting (2), consider raising it
 
 That indicates that the server is receiving more concurrent requests than it has PHP processes allocated, which means some requests will have to wait until another finishes.  In this example there are 2 PHP processes that can run concurrently.
 
-Platform.sh sets the number of workers based on the available memory of your container and the estimated average memory size of each process.  There are two ways to increase the number of workers:
+The following command will identify the 20 slowest requests in the last hour, which can provide an indication of what code paths to investigate.
 
-* Adjust the [worker sizing hints](/languages/php/fpm.md) for your project.
-* Upgrade your subscription on Platform.sh to get more computing resources. To do so, log into your [account](https://accounts.platform.sh) and edit the project.
+```bash
```
Crell: I don't agree with collapsing these two.  Talking about "server reached max_children" without pointing people at the worker sizing seems pointless.  It's one of the more straightforward things to do, and if a process has a lot of non-static callbacks on a given page then every single request is going to end up blocking on max_children on every request.  Jumping from that straight to finding slow requests seems like a non-sequitor.

src/development/troubleshoot.md
mrkschan: Each line on the FPM access log (php.access.log) has a percentage of time spent on running PHP. That's the percentage of time spent running in the PHP zen engine compared to the total time spent on processing the request.

If the percentage is high, that request processing is basically CPU bound.

src/development/troubleshoot.md
mrkschan: See a883dc6daf548dd08494975ee495765714628ca4, that's the intention for adjusting number of workers in moat case.

When the application is always hitting the limit, it should upgrade subscription / make better caching.

When the workload is CPU bound, adding more FPM processes without adding computing resources will just queue requests in FPM (instead of Nginx), waiting for CPU to be available.

If the workload is I/O bound, that's another story. Adding more workers may help non I/O bound request getting a chance for processing but all FPMs will eventually be filled if the I/O is slow.

src/development/troubleshoot.md
Crell: Ah, then that should be explained.  It's not obvious what all of the columns in that log file mean.

mrkschan: I would like to add a mention of the php-fpm access log format to either https://docs.platform.sh/languages/php.html or https://docs.platform.sh/languages/php/fpm.html. Or ... we're just doing things like https://docs.platform.sh/languages/php/fpm.html#measuring-php-worker-memory-usage `In the fifth column, you'll see the peak memory usage that occurred while each request was handled` on this page?

Which one should we go ahead?

Crell: Hm.  That log is generated by FPM, so probably the FPM page.  Which means we then should rename that page to not be sizing-specific, and have a log section at the bottom or something.

mrkschan: Added a section to explain the FPM access log

PR #877, opened 2018-07-06
MongoDB and MariaDB troubleshooting

xtfer:
Adds troubleshooting steps for:

- MongoDB SCRAM issue

Improves troubleshooting for:

- MariaDB Definer/Invoker issues

### DIFF
--- a/src/configuration/services/mongodb.md
+++ b/src/configuration/services/mongodb.md
@@ -130,4 +130,4 @@ For more details on upgrading and how to handle potential application backward c
 
 ### The "SCRAM-SHA-1" authentication mechanism requires libmongoc built with --enable-ssl
 
-MongoDB deprecated password authentication in MongoDB 3.6 and replaced it with [SCRAM] (https://docs.mongodb.com/v3.6/core/security-scram/). As Platform.sh PHP containers are not currently built with this option, the workaround is to set the authentication mechanism (MONGO-CR) explicitly using https://docs.mongodb.com/v3.6/reference/parameters/#param.authenticationMechanisms, or use a lower version.
+MongoDB deprecated password authentication in MongoDB 3.6 and replaced it with [SCRAM] (https://docs.mongodb.com/v3.6/core/security-scram/). As Platform.sh PHP containers are not currently built with this option, the workaround is to [set the authentication mechanism](https://docs.mongodb.com/v3.6/reference/parameters/#param.authenticationMechanism) (MONGO-CR) explicitly, or use a lower version.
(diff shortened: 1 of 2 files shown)


src/configuration/services/mongodb.md
```diff
@@ -123,4 +123,11 @@ For further references please see the [official mongodump documentation](https:/
...
-For more details on upgrading and how to handle potential application backward compatibility issues, please see [Release Notes for MongoDB](https://docs.mongodb.com/manual/release-notes).
\ No newline at end of file
+For more details on upgrading and how to handle potential application backward compatibility issues, please see [Release Notes for MongoDB](https://docs.mongodb.com/manual/release-notes).
+
+
+## Troubleshooting
+
+### The "SCRAM-SHA-1" authentication mechanism requires libmongoc built with --enable-ssl
+
+MongoDB deprecated password authentication in MongoDB 3.6 and replaced it with [SCRAM] (https://docs.mongodb.com/v3.6/core/security-scram/). As Platform.sh PHP containers are not currently built with this option, the workaround is to set the authentication mechanism (MONGO-CR) explicitly using https://docs.mongodb.com/v3.6/reference/parameters/#param.authenticationMechanisms, or use a lower version.
```
Crell: Well that's unfortunate.  Do we know how long until that gets resolved?

The link to the Mongo Docs should also be link-ified.

src/configuration/services/mysql.md
```diff
@@ -257,3 +254,25 @@ platform sql --relationship database -e master < my_database_snapshot.sql
...
+CREATE OR REPLACE SQL SECURITY INVOKER 
+VIEW `view_name` AS 
+SELECT 
+```
+
+### MySQL server has gone away
+
+Errors such as "PDO Exception 'MySQL server has gone away'" are usually simply the result of exhausting your existing diskspace. Be sure you have sufficient space allocated to the service in [.platform/services.yaml](/configuration/services.md).
+
+The current disk usage can be checked using the CLI command `platform db:size`. It is usually advisable to have at least double the reported size available.
```
Crell: We should warn that command is not all that reliable.  I don't know what you mean by "advisable to have at least double the reported size available".  You should make sure you have twice the space available it tells you is available?  That's an infinite loop. :smile: 

This is also a good place to link people to the health notifications, which is the better solution.

src/configuration/services/mongodb.md
xtfer: I dont know how long that situation will persist.

PR #880, opened 2018-07-12
WIP: 844 getting started

xtfer:


Crell:
Chris, can we have this in smaller chunks please?  It looks from skimming it like there's several distinct sub-changes here we could easily discuss on their own.  (And they wouldn't have immediate conflicts.)  Let's do that instead, please.  It's way too hard to discuss larger refactors like this all at once.

PR #884, opened 2018-07-16
Encourage using health notification to prevent filling database disk

mrkschan:


src/configuration/services/mysql.md
```diff
@@ -275,4 +275,4 @@ Errors such as "PDO Exception 'MySQL server has gone away'" are usually simply t
 
 The current disk usage can be checked using the CLI command `platform db:size`. Because of issues with the way InnoDB reports its size, this can out by up to 20%. As table space can grow rapidly, *it is usually advisable to make your database mount size twice the size reported by the `db:size` command*.
 
-
+You are encouraged to create [low-disk warning notification](/administration/integrations/notifications.html#low-disk-warning) to proactively prevent disk usage introduced issue.
```
Crell: :+1: on adding this, but the grammar in the sentence is off.  Try "to proactively warn of low disk space before it becomes an issue."  Also "create" should probably be "add a".

mrkschan: updated, but it's a bit strange to see "warn" twice on the sentence

`You are encouraged to add a low-disk warning notification to proactively warn of low disk space before it becomes an issue.`

PR #885, opened 2018-07-17
Fixing type in PHP code snipper

L0rdJ:


Crell:
Oopsies.  Good catch, thanks!

PR #891, opened 2018-07-25
Vulnerability Scanning and Penetration Testing

rinchen:
Created vulnerability and pen testing guidance documentation. Content approved via email by Nikola and Damien.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -88,7 +88,7 @@
   * [Fastly](/golive/steps/fastly.md)
   * [Cloudflare](/golive/steps/cloudflare.md)
 * [Troubleshooting](/golive/troubleshoot.md)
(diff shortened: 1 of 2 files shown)


src/SUMMARY.md
```diff
@@ -118,6 +118,7 @@
 * [Data retention](security/data-retention.md)
 * [Encryption](security/encryption.md)
 * [Protective block](security/protective-block.md)
+* [Vulnerability Scanning and Penetration Testing](security/pen-test.md)
```
Crell: We should probably use a slightly shorter title here, as the sidebar is not wide and users won't be able to read the whole thing anyway. :smile: I suggest "Security scans" or similar.

rinchen: Pushed revised version

PR #893, opened 2018-07-27
Data deletion

rinchen:
Adding a new data deletion section to help answer standard questions from customers.

### DIFF
--- a/src/security/data-deletion.md
+++ b/src/security/data-deletion.md
@@ -1,8 +1,6 @@
-# Data Deletion
+# Data deletion
 
-## Data deletion
-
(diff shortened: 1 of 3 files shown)


src/security/data-deletion.md
```diff
@@ -0,0 +1,17 @@
+# Data Deletion
+
+## Data deletion
```
Crell: No need to have double headers here.  Just remove the second, and use "Sentence case" for the title header.

Crell:
are there offsite links we can point to for the NIST standards etc. that are referenced?

PR #895, opened 2018-08-09
Update the php extensions

Paul-Chabas:
Update of the list of supported extensions. By supported is intended:
- Can be added in the `.platform.app.yaml`
- Can be loaded from the PHP application (tested via `extension_loaded()` PHP assert)

I did not made distinction between built-in extensions and the others.

### DIFF
--- a/src/languages/php/extensions.md
+++ b/src/languages/php/extensions.md
@@ -58,84 +58,109 @@ You can disable those by adding them to the `disabled_extensions` list.
 
 This is the complete list of extensions that can be enabled:
 
-| Extension    | 5.4 | 5.5 | 5.6 | 7.0 | 7.1 | 7.2 |
-|--------------|-----|-----|-----|-----|-----|-----|
-| amqp         |     |     |     | *   | *   | *   |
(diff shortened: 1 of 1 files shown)


Crell:
Are we able to differentiate between built in and non?  If something is built in then it cannot be disabled.  We should mark those that cannot be disabled somehow.  (I'm open to suggestions on the best way to do that.)

Paul-Chabas:
I will do that next week. Not sure about the way to display that. Also, after a talk with Damien, it appeared that some extensions did not have the same name in the `.platform.app.yaml` and in php itself. For example `pecl-http` is just `http`. Should we mention that ? 
Please do not merge now, I need to update the doc to fit the latest changes.

Crell:
We should list only the name that users would put in their YAML file, I'd say. That's the relevant portion.  If it's really non-obvious then perhaps something like "http (pecl-http)"?  But I think the former is sufficient for the most part.

Crell:
@ralt Are there other extensions that should be named differently?  Paul never got to finish his work here.

Also, do we have a list of those extensions that are built in so non-disableable?

PR #912, opened 2018-08-30
Improve .environment when installing nvm manually

hanoii:


src/languages/nodejs-nvm.md
```diff
@@ -24,6 +24,10 @@ hooks:
 And in a `.environment` file in the root of your project:
  
 ``` 
+# This is necessary for nvm to work
+unset NPM_CONFIG_PREFIX
+# Disable npm update notifier, being a read only system it will probably annoy you.
```
Crell: Semi-colon, not comma. :smile:

Crell: Also, comment lines should end in a period.

Crell:
I think the build issue here is a transient Node issue.  I've seen it before.  Another commit to fix the comments as noted above will probably clear that up, too.

Thanks, @hanoii!  Does this resolve the issue you were seeing before?

hanoii:
@Crell you mean what I brought up in Slack  yesterday? No, not really. It improved it when `.environment` is sourced, but I still think it's not on build/deploy/command start hooks.

Btw as you brought this up here, just linking the ticket for cross reference.

hanoii:
I just realised that I was modifying a file that's not longer referenced

hanoii:
Slightly modified other small bits.

Crell:
Thanks, @hanoii!

PR #914, opened 2018-08-31
Add information about using the CLI

OriPekelman:


src/configuration/app/workers.md
```diff
@@ -30,7 +30,21 @@ Any number of workers may be defined with their own distinct name, subject to av
 
 Like with any other application container Platform.sh allows you to connect to the worker instance through SSH to inspect logs and interact with it.
 
-To do so simply add the name of the worker to the user name after the application name part of the SSH url preceded by a double dash (`--`).
+Like with any other application container Platform.sh allows you to connect to the worker instance through SSH to inspect logs and interact with it.
+
+As always, simplest is to use the CLI, simply add the worker name to the ssh command. For example:
```
Crell: This sentence's grammar is not comprehensible.

Suggestion: "Using the Platform CLI you would use the `--worker` switch, like so:"

src/configuration/app/workers.md
```diff
@@ -30,7 +30,21 @@ Any number of workers may be defined with their own distinct name, subject to av
...
+
+```
+platform ssh --worker=my_worker
+```
+
+To output the SSH command you can use:
+
+```
+platform ssh --worker=sidekiq-1 --pipe
```
Crell: Let's use the same worker name in both examples for consistency.

PR #917, opened 2018-09-04
Use $PLATFORM_APP_DIR instead of $HOME for NVM

hanoii:
$HOME is not a valid ENV var when .environment is being loaded early on, use $PLATFORM_APP_DIR instead


hanoii:
@Crell this actually ended up fixing that issue I was having (if you remember).

Follow up of #912.

And thanks @mrkschan for this. It might be worth noting somehwere that $HOME is not yet available on .environment. Actually there's not much documentation about that file either.

PR #919, opened 2018-09-11
Update example.md

T1l3:


Crell:
Hm.  The ending dot in a domain name is... technically correct but rarely used.  The intent here was a period on the end of a sentence. :smile:  I suppose adding an extra period is fine but we shouldn't remove the one at the end of the sentence, as that's just bad grammar then.

PR #920, opened 2018-09-12
Fix read repository scope name

scollado:
While I set up a new GitLab integration for a customer, I saw that the scope key for git repositories read access was wrong.

Double-checked on our GitLab instance and fixed the key in the documentation.


pjcdawkins:
Thanks @scollado !

PR #921, opened 2018-09-14
Add Python 3 build hook instructions to FAQ

NickAnderegg:
Added instructions to the FAQ on how to use Python 3 in the build hook of a non-Python container.

It includes the necessary caveats:

- It should only be done when needed to set up an application
- It should not replace multi-apps
- It's locked to the most recent version of Python 3 included with Debian

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -62,6 +62,7 @@
 * [Node.js](languages/nodejs.md)
   * [Alternative Node.js install](languages/nodejs/nvm.md)
 * [Python](languages/python.md)
+  * [Using Python 3 in build hooks](languages/python/python3.md)
 * [Ruby](languages/ruby.md)
 * [Go (beta)](languages/go.md)
 * [Java (alpha)](languages/java.md)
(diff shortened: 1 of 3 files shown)


Crell:
This shouldn't go in the FAQ.  Nothing should ever go in the FAQ.  The FAQ needs to be refactored out of existence.

This should go in a sub-page of /languages/python.md.  (/languages/python/virtenv.md or similar)

NickAnderegg:
Instructions have been moved to their own page.

I've also revised the instructions a bit to better follow best practices:

- `python3.sh` example script now only contains commands for creating a venv and installing pip. Dependency installation examples have been moved directly into the build hook.
- The example build hook now ends with the `deactivate` command to deactivate the virtual environment when it is no longer needed. (If it's not deactivated, ssh-ing into a container will cause `python` to continue pointing to Python 3.)
- The "Caution" block about only using this in build hooks has been moved to the top, because it seems easy to miss at the very bottom. Additionally, it's been made explicitly clear that this should not be used in deploy hooks or web configurations.

NickAnderegg:
Making `python3-pip` available to applications by default instead

PR #922, opened 2018-09-18
Fix links in GDPR Overview

jameswilson:
Fixes two broken links "what data we collect" and "data retention" on the [GDPR Overview page](https://docs.platform.sh/security/gdpr.html).


Crell:
Merci bien!

PR #936, opened 2018-10-03
WIP: Replace old class names with latest from README

### LINKED ISSUE #933: Documentation Memcached
I followed the documentation here https://docs.platform.sh/frameworks/drupal8/memcached.html and it did not work.

In the last section, for the configuration of the settings.platformsh.php , It seems there is an error:
Fatal error: Class 'Drupal\memcache\Drupal Memcached Config' not found

I then followed the documentation of the module and it got resolved.

royallthefourth:
Resolves #933


Crell:
Well, how nice of them...

pjcdawkins:
For reference: https://cgit.drupalcode.org/memcache/commit/?id=9b92d62

Crell:
Needed to make one or two small changes while we were at it.  But I tested this and verified that it works.

PR #937, opened 2018-10-04
Add route ID functionality.

Crell:


src/configuration/routes.md
```diff
@@ -115,6 +115,85 @@ https://blog.example.com.sprint-7onpvba-tvh56f275i3um.eu-2.platformsh.site/
 
 If your project involves only a single apex domain with one app or multiple apps under subdomains, it's generally best to use the `{default}` placeholder.  If you are running [multiple applications](/configuration/app/multi-app.md) on different apex domains then you will need to use a static domain for all but one of them.
 
+## Route identifiers
+
+All routes defined on an environment will be available in the application in the `PLATFORM_ROUTES` environment variable.  That value is a base64-encoded JSON string, which can be extracted in the language of your choice to look up the effective routes that were generated.
```
NickAnderegg: - "available in the application" -> "available to the application"
- "extracted in the language of your choice" -> "extracted with the language of your choice"
- "look up the effective routes that were generated" -> ???

NickAnderegg: I'm not sure what that last one actually means. I mean, it's explained in the next sentence, but the wording is really awkward.

src/configuration/routes.md
```diff
@@ -115,6 +115,85 @@ https://blog.example.com.sprint-7onpvba-tvh56f275i3um.eu-2.platformsh.site/
...
+
+"Effective route" here means that all placeholders will be replaced with the appropriate domain name, and depending on your configuration additional route entries may be generated as well (such as the automatic HTTP->HTTPS redirect).  That means the domains will vary from one environment to another.  To make it easier to locate specific routes in a standardized fashion you may also specify an `id` on each route that is stable across all environments.  You may also mark one, and only one route as `primary`.  The primary route may be special cased in the interface but has no impact on the runtime environment.
```
NickAnderegg: - "specify an `id` on each route that is stable" -> "specify an `id` on each route which will remain stable"

The original phrasing makes it sound like the user has to maintain stability.

NickAnderegg: > The primary route may be special cased in the interface but has no impact on the runtime environment.

It's entirely unclear the purpose of this sentence.

NickAnderegg: > That means the domains will vary from one environment to another.

Why does that mean that domains will vary between environments?

src/configuration/routes.md
```diff
@@ -115,6 +115,85 @@ https://blog.example.com.sprint-7onpvba-tvh56f275i3um.eu-2.platformsh.site/
...
+    type: upstream
+    upstream: 'site1:http'
+
+"https://site2.{default}/":
+    type: upstream
+    id: 'the-second'
+    upstream: 'site2:http'
+```
+
+This example defines two routes, on two separate subdomains, pointing at two separate app containers.  (They could just as easily be pointing at the same container for our purposes).  On a branch named `test`, the routes array in PHP would look like this:
```
NickAnderegg: If you're naming the array "routes," I'd format it as `routes`. If you're just generally referring to an array for routes, then I think "route array" would be a better phrasing.

src/configuration/routes.md
Crell: It is a but clunky.  Suggestions for better wording welcome.

src/configuration/routes.md
Crell: That's because right now the primary route means nothing, but will eventually be used to show one link by default in the UI, and stuff like that.  But for now it doesn't do anything.

Crell: Because the generated route will vary by branch.

src/configuration/routes.md
```diff
@@ -117,11 +117,11 @@ If your project involves only a single apex domain with one app or multiple apps
...
 
-"Effective route" here means that all placeholders will be replaced with the appropriate domain name, and depending on your configuration additional route entries may be generated as well (such as the automatic HTTP->HTTPS redirect).  That means the domains will vary from one environment to another.  To make it easier to locate specific routes in a standardized fashion you may also specify an `id` on each route that is stable across all environments.  You may also mark one, and only one route as `primary`.  The primary route may be special cased in the interface but has no impact on the runtime environment.
+When routes are generated, all placeholders will be replaced with appropriate domains names, and if your configuration specifies them, additional route entries may be generated (e.g. automatic HTTP to HTTPS redirects). To make it easier to locate routes in a standardized fashion, you may specify an `id` key on each route which remains stable across environments. You may also specify a single route as `primary`, which will cause it to be highlighted in the web interface but will have no impact on the runtime environment.
```
Crell: HTTP->HTTPS redirects are only generated automatically if they're NOT specified in the file.  The change in the first sentence is thus backwards.

PR #938, opened 2018-10-04
Use new Template repository links, not the old Example repos.

Crell:
Then we can close down the old repos and not have to maintain 2 of them anymore! :clinking_glasses:


pjcdawkins:
Hmm there are missing pieces in the D8 template such as https://github.com/platformsh/template-drupal8/issues/14

PR #949, opened 2018-10-30
create new backup and restore overview page discussing RPO and RTO

rinchen:
Created a new backup and restore overview page discussing RPO and RTO to handle frequent customer inquiries.

### DIFF
--- a/src/security/backups.md
+++ b/src/security/backups.md
@@ -2,7 +2,7 @@
 
 Backups vary by our offering and are governed by our [data retention](/security/data-retention.md) policies.
 
-Below are details on our Recovery Point Objective (RPO) and Recovery Time Objective (RTO) for Platform Standard and Platform Enterprise.
+Below are details on our Recovery Point Objective (RPO) and Recovery Time Objective (RTO) for our Platform Standard and Platform Enterprise offerings.
 
 ## Platform Standard
 
--- a/src/security/backups.md
+++ b/src/security/backups.md
@@ -1,0 +1,17 @@
+# Backup and restore
+
+Backups vary by our offering and are governed by our [data retention](/security/data-retention.md) policies.
+
+## Platform Standard
+
+With Platform Standard, we have enabled our clients to manage their own backup and restore functions. Please see the [snapshot and restore](/administration/snapshot-and-restore.md) page for details.
+
+**RPO**: User defined. The RPO is configured by our clients.
+
+**RTO**: Variable. Recovery time will depend upon the size of the data we are recovering
+
+## Enterprise
+
+**RPO**: 6 hours.  Platform.sh takes a backup of Platform.sh Enterprise environments every 6 hours.
+
+**RTO**: Variable. Recovery time will depend upon the size of the data we are recovering
--- a/src/security/backups.md
+++ b/src/security/backups.md
@@ -2,6 +2,8 @@
 
 Backups vary by our offering and are governed by our [data retention](/security/data-retention.md) policies.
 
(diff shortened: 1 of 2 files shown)


src/security/backups.md
```diff
@@ -0,0 +1,19 @@
...
+
+Below are details on our Recovery Point Objective (RPO) and Recovery Time Objective (RTO) for our Platform Standard and Platform Enterprise offerings.
+
+## Platform Standard
+
+With Platform Standard, we have enabled our clients to manage their own backup and restore functions. Please see the [snapshot and restore](/administration/snapshot-and-restore.md) page for details.
+
+**RPO**: User defined. The RPO is configured by our clients.
+
+**RTO**: Variable. Recovery time will depend upon the size of the data we are recovering
```
Crell: This is not accurate, AFAIK.  It is variable based on a whole bunch of factors like network and system load, but data size is the only thing it doesn't depend on.  (Engineering can confirm/dispute that.)

src/security/backups.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Platform Enterprise
+
+**RPO**: 6 hours.  Platform.sh takes a backup of Platform.sh Enterprise environments every 6 hours.
```
Crell: Is it worth noting here that we keep a sliding scale of backups, so this does not mean we have all old backups every 6 hours, just the recent ones?

Crell:
I am still not convinced this should be its own page.

src/security/backups.md
rinchen: Interesting question. The answer would go on the retention page. Since this page is meant to answer compliance questions, and this line item is for PE, it might be better to discuss that as part of on-boarding vs put in the public docs (although I do have that information...it's just not easily consumable without some guidance).

src/security/backups.md
rinchen: Are you referencing RTO? This is the engineering vetted language that we've (myself & Rob) been using. The amount of data we need to restore definitely impacts how long it will take to perform the restore.

src/security/backups.md
```diff
@@ -0,0 +1,19 @@
+# Backup and restore
+
+Backups vary by our offering and their retention is governed by our [data retention](/security/data-retention.md) policies.
+
+Below are details on our Recovery Point Objective (RPO) and Recovery Time Objective (RTO) for our Platform Standard and Platform Enterprise offerings.
+
+## Platform Standard
```
robertDouglass: Platform Professional is how we refer to this.

robertDouglass: Please change "Platform Standard" to "Platform Professional" everywhere, in all documents.

https://platform.sh/pricing

src/security/backups.md
robertDouglass: You can say "see retention policy" and link to it. Then I consider this question resolved.

robertDouglass: scratch that. You've linked to retention above.

src/security/backups.md
robertDouglass: For enterprise it's definitely also tied to data volume. If you have to recreate MySQL from a text based dump, the volume of data is linearly correlated to time.

src/security/backups.md
rinchen: Pushed updated wording

src/security/backups.md
Crell: For Enterprise, definitely it's size-dependent.

I just checked with Engineering, though, and it is size dependent for PS restore, too, due to the need to copy the backup from S3 first.  So I withdraw my objection to this verbage.

PR #953, opened 2018-11-02
Delete ui-ssl.png

mouhamadoupsh:
Outdated image


pjcdawkins:
Thanks for pointing out the image is out of date, but this image is still used in https://docs.platform.sh/golive/steps/tls.html (in your branch: [/src/golive/steps/tls.md](https://github.com/mouhamadoupsh/platformsh-docs/blob/patch-2/src/golive/steps/tls.md)) so that needs to be edited too.

PR #954, opened 2018-11-02
Update relationships.md

shiftedreality:
Small correction to relationships example.


Crell:
Those names were deliberate, since they're all consistent.  That said, the rest of the docs use `rediscache`, so it's probably wise to not have a one-off exception.

Thanks!

PR #957, opened 2018-11-06
Add note about disk space requirements and bump example

xtfer:


src/configuration/services/rabbitmq.md
```diff
@@ -98,3 +98,9 @@ dependencies:
 ```
 
 Then, when you SSH into your container, you can simply type any `amqp-` command available to manage your queues.
+
+## Troubleshooting
+
+### Messages not being recieved
+
+RabbitMQ probably doesn't have enough disk space allocated to it. On Platform.sh, the service writes all of its data to disk, and thus needs a disk value that is >= its RAM allocation. For most Master branches, the service needs at least 1300 MB of disk space to work properly.
```
cdchris12: We should probably make some note here about Dev containers; they typically need something like ~150MB in order to work.

src/configuration/services/rabbitmq.md
```diff
@@ -34,7 +34,7 @@ In your ``.platform/services.yaml``:
 ```yaml
 myrabbitmq:
     type: rabbitmq:3.5
```
cdchris12: We should also update this to default to RabbitMQ 3.7 in the example, not 3.5. @xtfer

src/configuration/services/rabbitmq.md
Crell: Setting a different disk size on dev and master isn't a thing, so I'm not sure that will be practical.

Where does the 1300 MB number come from?  I'm pretty sure RabbitMQ doesn't take 1.3 GB of RAM...  The number needs some context and explanation.

Also, let's spell out >=.

cdchris12: 1300MB is more than enough for the largest RMQ containers I've yet seen, which was ~1238MB or some such amount of RAM. Long story short, it's a very safe amount to allocate for the RMQ disk value, imho.

Crell: I'd like to have more guidance for users here than a magic number that we hope is big enough.  That's nearly a quarter of the default amount of disk on basic plans.  We need to justify dedicating that much space to it.

PR #960, opened 2018-11-07
Explain conflict resolution in routes.yaml

mrkschan:
Routes without the `{all}` placeholder take precedence over those using `{all}` placeholder in case of having conflict.


src/configuration/routes.md
```diff
@@ -115,6 +115,8 @@ https://blog.example.com.sprint-7onpvba-tvh56f275i3um.eu-2.platformsh.site/
 
 If your project involves only a single apex domain with one app or multiple apps under subdomains, it's generally best to use the `{default}` placeholder.  If you are running [multiple applications](/configuration/app/multi-app.md) on different apex domains then you will need to use a static domain for all but one of them.
 
+Please note that when there are two routes sharing the same HTTP scheme, domain, and path, where the first route is using the `{default}` placeholder and the other is using the `{all}` placeholder, the one without using `{all}` placeholder takes precedence.
```
Crell: Turn around the last part; "the route using `{default}` takes precedence.".  Fewer negations that way.  Otherwise looks good.

PR #961, opened 2018-11-07
Update Blackfire doc on using environment variables for server credentials

ChristopheDujarric:


src/administration/integrations/blackfire.md
```diff
@@ -53,7 +50,29 @@ git commit -m "Enable Blackfire."
...
+
+Blackfire enables to have a fine grained configuration of server credentials across branches and environments on Platform.sh.
+
+#### Configuring global server credentials
+
+Configuring server credentials on your default branch will enable you to make sure you can profile any other branch:
+
+```bash
+platform variable:create env:BLACKFIRE_SERVER_ID <insert your Server ID>
+platform variable:create env:BLACKFIRE_SERVER_TOKEN <insert your Server Token>
```
Crell: Since the line above says this is for the "default branch", these lines should include `-e master` to ensure they go to the right place regardless of what the user has checked out at the moment.

Also, there is no "default branch".  "master branch" is the more accurate term.

src/administration/integrations/blackfire.md
```diff
@@ -40,9 +40,6 @@ Paste the credentials above in your `.platform.app.yaml` as follows:
 runtime:
     extensions:
         - name: 'blackfire'
-          configuration:
-              server_id: '<insert your Server ID>'
-              server_token: '<Insert your Server Token'
```
Crell: Why the switch away from inline definition to env vars?  I'm sure both work, but why remove this version?

ChristopheDujarric: This is a rather old/not very flexible/powerful way to do it compared to env vars.

Best is to promote the most up to date way. And to promote only one way, in order to avoid confusion ;)

src/administration/integrations/blackfire.md
ChristopheDujarric: Thanks for the feedback! Fixed.

Crell:
Merci bien!

PR #966, opened 2018-11-14
WIP: Postgres 10 documentation.

OriPekelman:


src/configuration/services/postgresql.md
```diff
@@ -259,62 +256,11 @@ If you see this error: ``Fatal error: Uncaught exception 'PDOException' with mes
...
-5. Make a Git commit with these changes, and push it into the development environment that you're testing with.
-6. In the development environment, review `/var/log/deploy.log` to confirm that the migration was successful, inspect the database, and test your application as needed to confirm that the migration was successful and that the application works with the new PostgreSQL version.
-7. When you're ready to upgrade the database on your master environment, start off by triggering a fresh snapshot of the master environment's data. This will give you the ability to roll back the upgrade process in case something goes wrong.
-8. Push the same commit from step 5 into your project's master branch. This will carry out the migration in your production environment.
-9. Finally, push out another commit that removes the old database service and the deploy hook used for migration, which is no longer needed now that the upgrade process is finished.
+Simply change the version in your services.yaml file, Platform.sh will transparently migrate the data.
```
ralt: It only works for pg10. Upgrading from, say, 9.3 to 9.6 still doesn't work.

OriPekelman: but both 9.3 to 10 and 9.6 to 10.0 would work right ? if so let's just say upgrading from 9.3 to 9.6 is unsupported and be done with it.

Crell: I'm fine with just declaring 9.3->9.6 unsupported, but this comment isn't resolved until the text is updated to reflect that.

src/configuration/services/postgresql.md
```diff
@@ -259,62 +256,11 @@ If you see this error: ``Fatal error: Uncaught exception 'PDOException' with mes
...
-2. Create an up-to-date copy of your database by cloning a development environment from master.
-3. Rename the master environment's postgresql service, and switch the version to 9.6. Renaming ensures that the service's storage is reset, which is necessary to avoid exposing the 9.6 server to incompatible data from version 9.3.
-4. Open an SSH session in your master environment, with agent forwarding enabled (so you can SSH from master to your development environment).
-5. Now, use SSH to run `pg_dump` on the development environment, and pipe the data from SSH into a `pg_restore` command running in the master environment. You can use the [sample deploy hook](#example-deploy-hook) below as a starting point. You'll need to create a .pgpass file on the development environment, and also wrap the `pg_dump` invocation in a call to `ssh`.
+> Warning: Make sure you first test your migration on a separate branch
+> Warning: be sure to take a snapshot of your master environment **before** you merge this change.
 
-### Example deploy hook
+## Downgrading
```
Crell: Downgrading doesn't need its own section if it's just one line.  Just include it in the also-now-very-short upgrading section.  It's fine.

Crell:
If PG 11 is also coming any day now, let's include that as well.  Plus we should add it to the changelog.

ralt:
It's not `10.0` but `10`.

src/configuration/services/postgresql.md
```diff
@@ -182,7 +185,7 @@ Platform.sh supports a number of PostgreSQL extensions.  To enable them, list th
 
 ```yaml
 postgresql:
-    type: "postgresql:9.6"
+    type: "postgresql:10.0"
```
ralt: (‿|‿) 💨

Issue #852, opened 2018-06-15
Broken links in Table of Contents


alice-0-kim:
Hi, I noticed that when the titles contain double quotes, the links in table of contents do not map into the right content.

Example:
```
https://docs.platform.sh/development/troubleshoot.html#http-responses-%22502-bad-gateway%22-or-%22503-service-unavailable%22
```
does not match with
```
https://docs.platform.sh/development/troubleshoot.html#http-responses-502-bad-gateway-or-503-service-unavailable
```

Thanks!

Crell:
Hi Alice.  Thanks for the head sup.

Unfortunately those links are auto-generated and the Gitbook plugin we use isn't handling quotes in section titles properly.  That's an upstream bug with the TOC plugin, so until that gets fixed we're kind of stuck with it.  (Or trying to weasel around it by changing our section titles, which I've considered.)

Crell:
Related #653

Issue #653, opened 2017-09-28
The generated custom PHP.ini link on TOC is wrong


mrkschan:
On https://docs.platform.sh/languages/php.html, the TOC "`Custom php.ini" link is wrong.

That wrong link is generated as `https://docs.platform.sh/languages/php.html#custom-php.ini`, while the correct one should be `https://docs.platform.sh/languages/php.html#custom-phpini`. Note the `.`.

Crell:
Hrm.  That seems like an upstream bug caused by the period.  We can t

PR #967, opened 2018-11-14
Resolve TOC issues #653 and #852

OriPekelman:
the atoc plugin does not suffer from this bug. Please test nothing else broke.


Crell:
The PR appears broken as it's pulling in the entire user-widget codebase.  Please fix that before we can merge.

Crell:
Spot checking the build, though, it looks like it does fix both issues and I see nothing else that it's broken, so yay.  We can merge as soon as the stray code is removed from the PR.

Crell:
Resolves #653 
Resolves #852 

(I think that will trigger GitHub to close those automatically when this is merged.)

Crell:
Closing in favor of #970.

PR #971, opened 2018-11-19
Network storage as a service

Crell:
Not out of RC yet, but getting the ducks all lined up anyway.


src/configuration/services/network-storage.md
```diff
@@ -0,0 +1,170 @@
...
+The network storage service enables a new kind of `mount` that refers to a shared service rather than to a local directory.  Any application can use both `local` or `service` mounts, or either/or, or neither.
+
+
+## Define the service
+
+First, declare a new service in the `services.yaml` file like so:
+
+```yaml
+files:
+    type: network-storage:0.2
```
xtfer: Why does network storage have a version number? This seems confusing.

ralt: @xtfer is there any that does _not_ have it?

Crell: I don't know that the system could handle not having one. :smile: I'm more curious if we're going to call it 1.0 before tagging the final.

src/configuration/services/network-storage.md
```diff
@@ -0,0 +1,170 @@
...
+## Declare the mount
+
+Second, add the following entry to your mounts list:
+
+```yaml
+mounts:
+    'my/files':
+        source: service
+        service: files
+        source_path: network
```
ralt: maybe `network` is confusing there?

src/configuration/services/network-storage.md
ralt: We're probably going to keep 0.2 until it's been battle-tested a bit.

src/configuration/services/network-storage.md
Crell: Probably.  Changed it.

src/configuration/services/network-storage.md
```diff
@@ -0,0 +1,170 @@
+# Network Storage
+
+Platform.sh supports internal "storage as a service" to provide a file store that can be shared between different application containers.
+
+The network storage service enables a new kind of `mount` that refers to a shared service rather than to a local directory.  Any application can use both `local` or `service` mounts, or either/or, or neither.
```
derrix060: >  Any application can use both local or service mounts,or either/or, or neither.

I would suggest to rephrase to `Any application can use both local and/or service mounts, or neither.`. The original `or either/or, or neither` looks very confusing...

src/configuration/services/network-storage.md
derrix060: Maybe worth add a quick not what this version actually means and rather or not is required?

src/configuration/services/network-storage.md
```diff
@@ -0,0 +1,170 @@
...
+        source: service
+        service: files
+        source_path: uploads/done
+```
+
+In this example, `app1` will have access to the entire `uploads` directory by writing to `web/uploads`.  `app2`, by contrast, will have two mounts that it can write to: `process` and `done`.  The `process` mount will refer to the same directory as the `web/uploads/incoming` directory does on `app1`, and the `done` mount will refer to the same directory as the `web/uploads/done` directory on `app1`.
+
+## Worker instances
+
+When defining a [Worker](configuration/app/workers.md) instance it is important to keep in mind what mount behavior is desired.  Unless the `mounts` block is defined within the `web` and `workers` sections separately, a top level `mounts` block will apply to both instances.  However, `local` mounts will be separate for each instance while `service` mounts will refer to the same file system.  For example:
```
derrix060: wrong link to `Worker` here. Shouldn't be `(/configuration/app/workers.md)`??

src/configuration/services/network-storage.md
```diff
@@ -0,0 +1,170 @@
...
+        source: local
+        source_path: console
+
+# Crons run on the web container, so they have the same mounts as the web container.
+crons:
+    drupal:
+        spec: '*/20 * * * *'
+        cmd: 'cd web ; drush core-cron'
+
+# The worker defined here will also have 6 mounts; 2 of them will be shared with the web container, the other 4 will be local to the worker.
```
derrix060: Note that the lines aren't all rendered depending on the screen size. Consider limiting the amount of char for each line to a lower number (80, maybe?). 

How it's now, is very difficult to read the comments: I need to scroll down until the last line from this `yaml` file, then scroll right, then scroll up and continue reading where I've stopped.

src/SUMMARY.md
```diff
@@ -31,6 +31,7 @@
   * [PostgreSQL](configuration/services/postgresql.md)
   * [MongoDB](configuration/services/mongodb.md)
   * [InfluxDB](configuration/services/influxdb.md)
+  * [Network Storage](configuration/services/network-storage.md)
```
derrix060: Any reason why this new session be between InfluxDB and Redis? I would suggest ordering alphabetically the summary...

src/configuration/app/storage.md
```diff
@@ -36,13 +36,34 @@ The `source` specifies where the writable mount is.  `source_path` specifies the
...
+
+```yaml
+mounts:
+  'web/uploads':
+    source: service
+    service: files
+    source_path: uploads
+```
+
+This assumes that a `network-storage` service named `files` has already been defined.  See the [Network Storage](configuration/services/network-storage.md) page for more details and examples.
```
derrix060: Broken link. Add a `/`: `/configuration/services/network-storage.md`

src/SUMMARY.md
Crell: The menu is largely "clustered" right now than having any specific order.  It may make sense to reorder it but that should be its own PR.

Crell:
Various changes based on the feedback above.  Thanks, @derrix060!

Crell:
Rebased for the reordered sidebar.

ralt:
@Crell could you put `network-storage:1.0` now please?

src/configuration/services/network-storage.md
```diff
@@ -0,0 +1,187 @@
...
+mounts:
+    'my/files':
+        source: service
+        service: files
+        source_path: network
+```
+
+This block will declare a writeable mount on the application container at the path `my/files`, which will be provided by the `files` service defined above.  The `source_path` specifies the path within the network service that the mount points to.  It is often easiest to have it match the name of the mount point itself but that is not required.
+
+Note that you do *not* need to add a relationship to point to the `files` service.  That is handled automatically by the system.
```
ralt: Maybe "That is handled automatically by the system" should be "That is made implicit by the mount declaration" or something similar?

src/configuration/services/network-storage.md
```diff
@@ -0,0 +1,187 @@
...
+        source: service
+        service: files
+        source_path: uploads/done
+```
+
+In this example, `app1` will have access to the entire `uploads` directory by writing to `web/uploads`.  `app2`, by contrast, will have two mounts that it can write to: `process` and `done`.  The `process` mount will refer to the same directory as the `web/uploads/incoming` directory does on `app1`, and the `done` mount will refer to the same directory as the `web/uploads/done` directory on `app1`.
+
+## Worker instances
+
+When defining a [Worker](/configuration/app/workers.md) instance it is important to keep in mind what mount behavior is desired.  Unless the `mounts` block is defined within the `web` and `workers` sections separately, a top level `mounts` block will apply to both instances.  However, `local` mounts will be separate for each instance while `service` mounts will refer to the same file system.  For example:
```
ralt: > it is important to keep in mind what mount behavior is desired.

This reads weird, no? That feels like it should be "which mount behavior is desired"?

ralt: Also:

> However, `local` mounts will be separate for each instance

That feels like it's not explaining correctly. Something like "However, `local` mounts will be separate disks for each instance", or something along these lines? I'm not sure I like my way either.

SerheyDolgushev:
Maybe it would be a good idea to describe the migration process from the local mount to network storage?

src/configuration/services/network-storage.md
Crell: Is there a relationship added in $PLATFORM_RELATIONSHIPS?  If so, then yes it should be reworded.  If not, then implying that a relationship is added is misleading for the customer as it's not, from their POV.

src/configuration/services/network-storage.md
Crell: "independent for each instance"?  Like, what's the opposite of shared? :smile:

src/configuration/services/network-storage.md
ralt: hmmm no there isn't... yeah idk, I guess it's fine as is, that just sort of confused me first time I read it

Crell:
@SerheyDolgushev A valid point, yes.  I don't think we have one other than manual, though, so that's really all we'd be documenting.

src/configuration/services/network-storage.md
ralt: dunno, you're the native :P

SerheyDolgushev:
Having few words about manual files migration is much better than do not have it at all :)

src/configuration/services/network-storage.md
Crell: Looking at the context again, I think "what" is correct here.  "Which" implies an either/or type situation, whereas what we offer is a "mix and match" configuration.

PR #972, opened 2018-11-27
Add a note about snapshot retention to the snapshot-and-restore page

ajm348:
(copy paste from retention page)


src/administration/snapshot-and-restore.md
```diff
@@ -6,6 +6,8 @@ Snapshots are triggered directly via the Web Interface or via the CLI. The snaps
 
 You need to have the "admin" role in order to create a snapshot of an environment.
 
+Snapshots on Platform.sh Professional are retained for at least 7 days. They will be purged between 7 days and 6 months, at Platform.sh's discretion.
```
Crell: I'm fine with adding this line here, but it should link somewhere to the retention page for more complete details.  Maybe a "See the [data retention] page for more information" sentence at the end.

Issue #978, opened 2018-11-30
Remove old redirects from routes.yaml


derrix060:
I was investigating #977 and I've found some spelling issues in `routes.yaml`, them I've seen the comments saying that "should" be safe to remove at the end of this year.

I know that is not something urgent, but should be an easy job only deleting the `redirects.paths` section.

Crell:
Closing per #979

PR #979, opened 2018-11-30
#978 remove old redirects

derrix060:
Would be nice if someone that has access to the statistics for the website sees how many people visited these removed websites since some months ago. If the amount of people is big enough, I would say that just leave the code how it is...


Crell:
It looks like this has #977 in it by accident.

The last time I checked there was still a decent amount of traffic for the redirects, although that was quite a while ago.  It doesn't hurt anything to have them there, really, so I've not bothered going back to check in a while.

PR #981, opened 2018-12-05
Documented another possible case for "MySQL server has gone away"

olivermack:
In my case - a php worker environment - it seemed to work to restart the process when this issue popped up. The timeout-directive increase is an assumption though.

### DIFF
--- a/src/configuration/services/mysql.md
+++ b/src/configuration/services/mysql.md
@@ -277,3 +277,6 @@ Errors such as "PDO Exception 'MySQL server has gone away'" are usually simply t
 The current disk usage can be checked using the CLI command `platform db:size`. Because of issues with the way InnoDB reports its size, this can out by up to 20%. As table space can grow rapidly, *it is usually advisable to make your database mount size twice the size reported by the `db:size` command*.
 
 You are encouraged to add a [low-disk warning notification](/administration/integrations/notifications.html#low-disk-warning) to proactively warn of low disk space before it becomes an issue.
+
+Another issue might be in context of long running worker processes that try to re-use a timed out connection. In such cases it can happen to see the same error popping up in your logs. If possible try to establish a new connection in your application or - in case of workers - ensure a shorter execution/idle time or enforce some sort of restart by exiting the current command. 
+In PHP environments you can also try to increase the `mysql.connect_timeout` and/or `default_socket_timeout` in your [php.ini](/languages/php/ini.html).


src/configuration/services/mysql.md
```diff
@@ -277,3 +277,6 @@ Errors such as "PDO Exception 'MySQL server has gone away'" are usually simply t
 The current disk usage can be checked using the CLI command `platform db:size`. Because of issues with the way InnoDB reports its size, this can out by up to 20%. As table space can grow rapidly, *it is usually advisable to make your database mount size twice the size reported by the `db:size` command*.
 
 You are encouraged to add a [low-disk warning notification](/administration/integrations/notifications.html#low-disk-warning) to proactively warn of low disk space before it becomes an issue.
+
+Another issue might be in context of long running worker processes that try to re-use a timed out connection. In such cases it can happen to see the same error popping up in your logs. If possible try to establish a new connection in your application or - in case of workers - ensure a shorter execution/idle time or enforce some sort of restart by exiting the current command. 
+In PHP environments you can also try to increase the `mysql.connect_timeout` and/or `default_socket_timeout` in your [php.ini](/languages/php/ini.html).
```
ralt: You are likely hitting the mysql 10 minutes timeout, which means if anything, you want to _lower_ the timeout on the client side, to make sure it's trying to ping the connection again sooner.

olivermack: As far as I can see the default php-ini's `default_socket_timeout` sits at 60 seconds. Shouldn't this do the trick then?

ralt: I'm guessing this timeout is a TCP timeout for the initial connection. You likely want a "MySQL idle timeout", just maintaining the TCP connection is not enough.

olivermack: Yeah, I see. Thanks for the feedback. I was just wondering that this kind of incident is not documented anywhere. It would have helped me while troubleshooting.
Shall I remove the tip for the timeout-adjustments? I cannot really provide a tip for configuring this properly right now - except for wrapping the failing code and try to reconnect.

ralt: I... honestly don't know. I'm bad at documentation...

Crell:
I think it's not documented because you're the first person to hit this issue that I'm aware of.  Thanks for reporting it!

Based on the discussion thread above, I agree we should document it but decided we should take a different approach.  Rather than try to rewrite through PR comments I decided to make a fresh PR here: #983.

Let's continue over there.  @olivermack, I'd definitely appreciate your feedback on whether it's clear or not.  Thanks.

PR #984, opened 2018-12-07
Note that bash history is now persistent.

Crell:


src/changelog/2018.md
```diff
@@ -1,5 +1,11 @@
 # December 2018
 
+## Improved Bash support
+
+Bash history on application containers now persists between logins.
```
ralt: "usually" persists, given that history is stored on `/tmp`, which gets wiped out every now and then... not sure how to say it :)

Crell: We can either throw the word "most" in there and invite questions about what that means or ignore that detail.  I'm not sure that there's a better option unless we want a big paragraph in the main docs somewhere.

PR #987, opened 2018-12-12
Added a redirect http code example

RobertoNovelo:


Crell:
Makes sense, but then the description text below it should also be updated to say which redirect code each one will use.  (Also, it should probably use a 3087, not a 301.  See: https://evertpot.com/http/308-permanent-redirect)

demosdemon:
from the API:

```
    code = schema.SchemaNode(
        schema.Integer(),
        title="The redirect code to use.",
        validator=schema.OneOf((301, 302, 307, 308)),
        missing=302,
    )
```

PR #997, opened 2019-01-03
use prettier to unify formatting

demosdemon:
use `npm install && npm run format` to run prettier locally on all the files in the repository.

added `npm test` to the build, if prettier detects that a file should be reformatted, it will fail the build and should mark the pull request as a failure


Crell:
Um, no.  There's no reason to switch from * to - (I prefer * frankly), and we've already debated sentence spacing at length.  This will also break every other PR currently in the queue.

PR #1002, opened 2019-01-06
remove HHVM

robertDouglass:
as per https://platform.sh/blog/hhvm-deploys-off-into-the-sunset


Crell:
Damn, missed one when we removed it.  Thanks.

PR #1007, opened 2019-01-09
Container updates

Crell:
* New RabbitMQ
* New Solr


ludoge:
Solr 7.6 and RabbutMQ 3.7 are now live, merging.

PR #1009, opened 2019-01-10
Clarifying env:PLATFORMSH_CLI_TOKEN when using CLI on app container

ollie-platform:
The docs here: https://docs.platform.sh/gettingstarted/cli/api-tokens.html

Are unclear that the api token environment variable must be set as `env:PLATFORMSH_CLI_TOKEN`, and not `PLATFORMSH_CLI_TOKEN`, when setting up the Platform CLI on a platform app container. A client made this mistake and I made the same mistake when trying to debug the issue the client was having.


Crell:
style note: The "note" should be done as a block, like so:

```
> **note**
> Blah blah blah.
```

Otherwise, :+1:

PR #1014, opened 2019-01-17
Update snapshot-and-restore.md

alphon-sh:


Crell:
This totally warrants more than just a note.  I also don't quite grok the --branch-from description.  What does that even mean?

This needs a proper bulleted list and some examples.   It's quite a big deal. :smile:

PR #1017, opened 2019-01-19
update path in the example

gauthiergarnier:
The path on my side is in this format: `/etc/php/7.2-zts/fpm`


gauthiergarnier:
As discussed with @ralt , on PHP5.x containers, the existing path is correct.

Crell:
Given that the docs say step 1 for tuning is "get on PHP 7 or get out", we should probably make it a PHP 7-only command and then note that it's PHP7-only. :smile: Is the new command PHP 7-targeted and tested?

We should add a note to that effect, if so.  (This branch is from a forked repo so I can't update it directly.)

PR #1024, opened 2019-01-31
Cc edits

chadwcarlson:
A few simple typo edits to start out.

### DIFF
--- a/src/administration/web.md
+++ b/src/administration/web.md
@@ -1,6 +1,6 @@
 # Web Interface
 
-[Platform.sh](https://platform.sh/) provides a responsive Web Interface which allows you to interact with your projects and manage your environments. Everything you can do with this interface you can also achieve with the  [CLI (Command Line Interface)](/gettingstarted/cli.md) and vice versa.
+Platform.sh provides a responsive Web Interface which allows you to interact with your projects and manage your environments. Everything you can do with this interface you can also achieve with the  [CLI (Command Line Interface)](/gettingstarted/cli.md) and vice versa.
 
 ![Web Interface](/images/platform-ui.png)
 
--- a/src/administration/web/environments.md
+++ b/src/administration/web/environments.md
@@ -1,6 +1,6 @@
 # Platform.sh Environments
 
-[Platform.sh](https://platform.sh) helps a coder with the development workflow by making it easy to manage multiple environments, including the Master environment which runs the production website. It's precisely like a "development" or a "staging" server.. except they are created on the fly, and they are absolutely identical copies of their parent environments.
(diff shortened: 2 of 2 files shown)


src/administration/web.md
```diff
@@ -1,6 +1,6 @@
 # Web Interface
 
-[Platform.sh](https://platform.sh/) provides a responsive Web Interface which allows you to interact with your projects and manage your environments. Everything you can do with this interface you can also achieve with the  [CLI (Command Line Interface)](/gettingstarted/cli.md) and vice versa.
+Platform.sh provides a responsive Web Interface which allows you to interact with your projects and manage your environments. Everything you can do with this interface you can also achieve with the  [CLI (Command Line Interface)](/gettingstarted/cli.md) and vice versa.
```
Crell: Why remove the link here?

src/administration/web/environments.md
```diff
@@ -1,6 +1,6 @@
 # Platform.sh Environments
 
-[Platform.sh](https://platform.sh) helps a coder with the development workflow by making it easy to manage multiple environments, including the Master environment which runs the production website. It's precisely like a "development" or a "staging" server.. except they are created on the fly, and they are absolutely identical copies of their parent environments.
+[Platform.sh](https://platform.sh) helps a coder with the development workflow by making it easy to manage multiple environments, including the Master environment which runs the production website. It's precisely like a "development" or a "staging" server... except they are created on the fly, and they are absolutely identical copies of their parent environments.
```
Crell: An ellipsis is 3 periods.  Two doesn't make sense.  We could potentially swap it out for an ellipsis entity? :smile:

src/administration/web.md
chadwcarlson: My edit was based on point 3 in the section **Language usage** in the Contributing Guidelines. 

> The name of the company is always written as "Platform.sh", not simply "Platform". It should not be linked anywhere.

I noticed after my commit Platform.sh linked on other pages, so I wonder if this is a new standard that should be taken out of the Contributing Guidelines?

chadwcarlson: Or perhaps replaced with:

> The name of the company is always written as "Platform.sh", not simply "Platform", It should not be linked anywhere, unless it is the first mention on a page.

chadwcarlson: > The name of the company is always written as "Platform.sh", not simply "Platform". It should not be linked anywhere, unless it is the first mention on a page.

Crell: Oh!  Hm.  I did write that, didn't I... :smile: 

It's more likely that we just didn't clear up a lot of the old links lying around.  We should probably not link the name, although I MIGHT be convinced that having the first mention on the very first page be a link is a worthwhile exception (as I apparently thought it was, wrongly), since otherwise there's no link to our actual website from the docs.

Which now that I think about it is also wrong, because we have the header bar attached anyway and that does have a link back to the main site.  So yeah, kill all the links!

src/administration/web/environments.md
Crell: I'm good either way.

PR #1027, opened 2019-02-05
kill all the links

chadwcarlson:
[Platform.sh](https://platform.sh/) links removed as per Contributing Guidelines: 

> 'The name of the company is always written as "Platform.sh", not simply "Platform". It should not be linked anywhere.'

### DIFF
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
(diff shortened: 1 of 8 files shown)


.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="$PROJECT_DIR$" vcs="Git" />
+  </component>
+</project>
```
Crell: You need to set your global .gitignore :smile: Rebase without this please.

PR #1033, opened 2019-02-13
Add note about variable

dpraburaj:
A user suggested that it would be helpful if there was a note saying that `PLATFORM_RELATIONSHIPS` will not be available in the build phase.

### DIFF
--- a/src/configuration/app/build.md
+++ b/src/configuration/app/build.md
@@ -93,7 +93,7 @@ Hooks are executed using the dash shell, not the bash shell used by normal SSH l
 
 ### Build hook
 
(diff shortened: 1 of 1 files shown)


Crell:
Thanks!

There's a whole bunch of variables that are not available in the build phase.  Instead of calling out just the one, let's mention that some environment variables are not available and link to https://docs.platform.sh/development/variables.html#platformsh-provided-variables, which lists out all the variables and when they exist.

dpraburaj:
@Crell I've expanded the note. I tried but couldn't come up with better phrasing, please feel free to edit as required.

Crell:
I'm not able to push to your branch directly.  Try this for the ending of the sentence:

"he full list of build time and runtime variables is available on the [variables page](link here)"

dpraburaj:
I've updated the last sentence, thanks.

PR #1034, opened 2019-02-15
Update access.md

tristanblattmansh:
Replaced `listed` with contributor. Listed no longer exists

### DIFF
--- a/src/configuration/app/access.md
+++ b/src/configuration/app/access.md
@@ -1,6 +1,6 @@
 # Access
 
-The `access` key in `.platform.app.yaml` defines the user roles who can log in via SSH to the environments they have permission to access.  The contributor role is a minimum; anyone with an access level of the contributor role or higher can access the container via SSH.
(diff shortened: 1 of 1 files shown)


pjcdawkins:
"listed" doesn't refer to a role name, it's just slightly ambiguous phrasing for "the role specified as the value of `access`".

`contributor` is not the minimum

src/configuration/app/access.md
```diff
@@ -1,6 +1,6 @@
 # Access
 
-The `access` key in `.platform.app.yaml` defines the user roles who can log in via SSH to the environments they have permission to access.  The listed role is a minimum; anyone with an access level of the listed role or higher can access the container via SSH.
+The `access` key in `.platform.app.yaml` defines the user roles who can log in via SSH to the environments they have permission to access.  The contributor role is a minimum; anyone with an access level of the contributor role or higher can access the container via SSH.
```
pjcdawkins:
```suggestion
The `access` key in `.platform.app.yaml` defines the user roles who can log in via SSH to the environments they have permission to access.  The specified role is a minimum; anyone with an access level of this role or higher can access the container via SSH.
```

PR #1037, opened 2019-02-28
php: Add docs for request_terminate_timeout

mrkschan:


Crell:
We should include some guidance here on why and when you'd want to do that.  This doesn't have any context right now.

src/languages/php.md
```diff
@@ -69,6 +69,17 @@ Then when you are connected to your project over SSH you can run:
 $ php-fpm-status --socket=unix://$SOCKET --path=/-/status --full
 ```
 
+## Runtime configuration
+
+It is possible to change the PHP-FPM runtime configuration via the `runtime` block on your `.platform.app.yaml`. Below PHP-FPM options are configurable:
```
Crell: Try "The PHP-FPM options below are configurable".

PR #1043, opened 2019-03-08
Add Ruby version 2.6 to the Changelog

GuGuss:


Crell:
The example a bit further down should be updated to reference 2.6 now, too.

I'm fairly sure the fail here is just GitBook goofing again.  A new commit/build usually fixes it.

PR #1046, opened 2019-03-14
Snapshot restoration to different env in older regions

MacWeber:
This feature is not available in all regions, we should make it clear in our documentation.

### DIFF
--- a/src/administration/snapshot-and-restore.md
+++ b/src/administration/snapshot-and-restore.md
@@ -25,7 +25,7 @@ Please be aware that triggering a snapshot will cause a momentary pause in site 
 
 You will see the snapshot in the activity feed of your environment in the Platform.sh Web Interface. You can trigger the restore by clicking on the `restore` link. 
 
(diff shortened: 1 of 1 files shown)


src/administration/snapshot-and-restore.md
```diff
@@ -23,7 +23,9 @@ Please be aware that triggering a snapshot will cause a momentary pause in site
 ## Restore
 
 
-You will see the snapshot in the activity feed of your environment in the Platform.sh Web Interface. You can trigger the restore by clicking on the `restore` link. You can also restore the snapshot to a different environment using the CLI.
+You will see the snapshot in the activity feed of your environment in the Platform.sh Web Interface. You can trigger the restore by clicking on the `restore` link. 
+
+Except for a few regions, you can also restore the snapshot to a different environment using the CLI. If your region does not have this feature, feel free to open a support ticket requesting to restore the snapshot to an environment of your choice.
```
Crell: We should also include a link to the instructions for users to migrate regions.  We want users to migrate off of us-1 and us-2.  In fact, let's name-drop those specifically, and encourage users to migrate.

Crell:
Closing in favor of #1088 .

PR #1047, opened 2019-03-15
Add External Authentication module requirement

ban-dog:


Crell:
That... looks wrong.  The drupal/simplesaml_auth module should already be depending on other packages it requires, no?  In fact, the fact that we even specify the simplesamlphp/simplesamlphp library is a bug, because both of those should be listed in the module's composer.json file.

This sounds like an upstream issue to file with the simplesaml_auth library to add both composer dependencies.  If they refuse then the update on our side would just be to add the additional library to the existing `composer install` command in the next section.

Crell:
Weber filed an upstream patch here: https://www.drupal.org/project/simplesamlphp_auth/issues/3041553

No immediate movement on it, so I'm going to go ahead and merge this for now.  We can change it if upstream ever fixes its dependencies.

PR #1051, opened 2019-03-19
On eZ Platform, varnish needs to be disabled if using Fastly

vidarl:
Subject says it all

### DIFF
--- a/src/frameworks/ez/fastly.md
+++ b/src/frameworks/ez/fastly.md
@@ -2,6 +2,21 @@
 
 eZ Platform Enterprise is a "commercial extended" version of ez Platform that includes, among other things, support for push-based purging on the Fastly CDN.
 
+## Remove Varnish configuration
+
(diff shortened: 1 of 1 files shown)


src/frameworks/ez/fastly.md
```diff
@@ -2,6 +2,21 @@
 
 eZ Platform Enterprise is a "commercial extended" version of ez Platform that includes, among other things, support for push-based purging on the Fastly CDN.
 
+## Remove Varnish configuration
+
+As of eZ Platform 1.13.5, 2.4.3 and 2.5.0, varnish is enabled by default when deploying on Platform.sh. In order to use Fastly, Varnish must be disabled:
```
DominikaK:
```suggestion
As of eZ Platform 1.13.5, 2.4.3 and 2.5.0, Varnish is enabled by default when deploying on Platform.sh. In order to use Fastly, Varnish must be disabled:
```

Crell:
Legit.  Thanks.

PR #1056, opened 2019-03-24
Security overview page

xtfer:
Adds a top-level Security Overview page to allow easier linking for Sales et al.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -117,14 +117,14 @@
 ## Security and compliance
 
 * [Overview](security/index.md)
-* [GDPR](security/gdpr.md)
+* [Backup and restore](security/backups.md)
+* [Compliance guidance](security/compliance-guidance.md)
 * [Data collection](security/data-collection.md)
 * [Data retention](security/data-retention.md)
 * [Data deletion](security/data-deletion.md)
-* [Backup and restore](security/backups.md)
 * [Encryption](security/encryption.md)
+* [GDPR](security/gdpr.md)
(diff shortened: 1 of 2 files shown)


rinchen:
@xtfer I like the clarification of the menu but the way you've written it we'll have to update two indexes, one in summary.md and the other in index.md.  We've been using summary.md as index.md. See if you can get it down to one index please so we don't have to update multiple places.

xtfer:
@rinchen Yes, we will. I think thats the price to pay for having documentation that is actually navigable and useful.

Crell:
I don't see how this makes it "navigable and useful".  The list is already visible in the sidebar, with the exact same words.  Adding yet another page that links to them has no value.

Reordering the pages in the sidebar is fine; that's Joey's call and I'm good either way.  But the extra duplicate page doesn't add any value that I see.

Crell:
Is there anything to do here, or shall we close?

xtfer:
> I don't see how this makes it "navigable and useful". 

Sales is frequently asked for information on our Security practices and features during the presales process. We have no location to direct people to to cover that. Sending them to one page and suggesting they then explore the sidebar for the other links is clumsy. Sending all the links is worse.

This solves that problem, thus making it both navigable and useful.

> I'm unhappy with having duplicate indexes requiring changes when we add or modify entries.

This is a limitation of the way we decided to build our docs. It doesn't mitigate my point above.

Crell:
The Hugo migration looks like it will include auto-generated landing index pages, so this issue will be resolved by that.  Closing for now on that expectation.

PR #1064, opened 2019-04-01
Add some cors

OriPekelman:


Crell:
What's the value in explicitly documenting how to setup such an insecure configuration?  There may be one but it should be far better explained than this.

OriPekelman:
People are specifically searching for this. I am OK with expanding. Also I don't believe there are any important security concerns for CORS for static assets which is what we are talking about here. POST and DELETE are not going to do much harm. Nor GET for that matter.

Crell:
Mozilla explicitly says this is a bad idea:
https://twitter.com/gadgetopia/status/1119237178771607552

OriPekelman:
We are talking here about static content. When you are serving static content you probably want CORS wide open because it makes your life easier (otherwise you either have to hard code the domain, and have only one...) or you _must_ serve it through the app. And people are looking for this answer. I don't think it is mega-useful, but I can expand the section to explain that they _should not_ use the same for their CORS headers on the app .. ?

Crell:
@OriPekelman Are you going to continue this?  I am highly skeptical about a copy-paste example that is documented as insecure, and I'm not convinced that even for static files it's entirely secure.

alustenberg:
+1 on @Crell 's viewpoint;  while having ACAC: true while ACAO: * is set is not explicitly called out by spec, no modern browser will actually allow the request.

https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSNotSupportingCredentials

https://github.com/whatwg/fetch/issues/251

i'd just remove `Access-Control-Allow-Credentials: true` and go from there;  if it's purely for the case of static content, then 'credentials' should not come into play anyway.

PR #1066, opened 2019-04-01
docs: Correct typo in simplesamlphp documentation

mattpaz:


Crell:
Ugh, I always do that...  Good catch, thanks.

PR #1068, opened 2019-04-04
DO NOT MERGE: Legacy UI Documentation

chadwcarlson:
In the transition period where customers are still able to toggle back to the previous user interface, this acts as a snapshot of the previous public docs during that time.


Crell:
I got confirmation from Thomas L that this isn't linked from anywhere, so we can take it down.

PR #1072, opened 2019-04-08
We support only TLSv1.2 now. Updated the documentation to reflect that.

rinchen:
For SEC-54.  We support only TLSv1.2 now. Updated the configuration documentation to reflect that.

### DIFF
--- a/src/configuration/routes/https.md
+++ b/src/configuration/routes/https.md
@@ -65,7 +65,7 @@ https://{default}/:
 ### `min_version`
 
 > **Note**
-> This directive was put into place when Platform.sh still supported older versions of TLS for customers.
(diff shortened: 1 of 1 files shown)


src/configuration/routes/https.md
```diff
@@ -64,14 +64,18 @@ https://{default}/:
 
 ### `min_version`
 
-Setting a minimum version of TLS will cause the server to automatically reject any connections using an older version of TLS.  While the vast majority of modern browsers will default to TLS 1.2, there are some older browsers that still use insecure older versions of TLS.  Rejecting older versions with known security vulnerabilities is necessary for some security compliance processes.
+> **Note**
+> This directive was put into place when Platform.sh still supported older version of TLS for customers.
```
chadwcarlson: older versions instead of version?

PR #1076, opened 2019-04-16
Document Source Operations

GuGuss:


src/configuration/app/source-operations.md
```diff
@@ -0,0 +1,64 @@
...
+
+The command has two parameters:
+
+* `operation` (string): the name of the operation
+* `variables` (optional, object): additional variables (in the style of variables defined in the `.platform.app.yaml`) to inject in the environment of the source operation
+
+When this operation is triggered:
+
+* A clean Git checkout of the current environment HEAD commit is created; this checkout doesn't have any remotes, has all the tags defined in the project, but only has the current environment branch.
+* Sequentially (_the sequence is in lexicographic order of the path of the `.platform.app.yaml` file_), for each application that has defined this operation, the operation command is launched in container image of the application, with the same type of environment you will find in a build, with the addition of all the runtime environment variables of the current environment, optionally overridden by the variables specified in the operation payload.
```
pjcdawkins:
```suggestion
* Sequentially (_the sequence is in lexicographic order of the path of the `.platform.app.yaml` file_), for each application that has defined this operation, the operation command is launched in the container image of the application, with the same type of environment you will find in a build, with the addition of all the runtime environment variables of the current environment, optionally overridden by the variables specified in the operation payload.
```

src/configuration/app/source-operations.md
```diff
@@ -0,0 +1,64 @@
...
+* At the end of the process, if any commit was created, the environment branch is updated to this commit, and the normal build process of the environment is triggered.
+
+Note that this operation runs in an isolated environment, it is not part of the runtime cluster of the environment, and doesn't require the environment to be running.
```
pjcdawkins:
```suggestion
Note that this operation runs in an isolated container; it is not part of the runtime cluster of the environment, and doesn't require the environment to be running.
```

src/configuration/app/source-operations.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Automated Source Operations using cron
+
+You can use cron to automatically run your source operations.
+
+> **note**
+>
+> Automated source operations using cron requires to [get an API token and install the CLI in your application container](/gettingstarted/cli/api-tokens.md).
+
+Once the CLI is installed in your application container and an API token configured you can add a cron task to run once a day and run your source operations. We recommend to not trigger those source operations in your `master` production environment, but on a dedicated environment which you can use for testing before deployment.
```
pjcdawkins:
```suggestion
Once the CLI is installed in your application container and an API token configured you can add a cron task to run once a day and run your source operations. We do not recommend triggering source operations on your `master` production environment, but rather on a dedicated environment which you can use for testing before deployment.
```

src/configuration/app/source-operations.md
```diff
@@ -0,0 +1,64 @@
...
+
+On the example below, we synchronize the `update-dependencies` environment with its parent before running the `update` source operation:
+```yaml
+crons:
+    update:
+        # Run the 'update' source operation every day.
```
pjcdawkins:
```suggestion
        # Run the 'update' source operation every day at midnight.
```

src/configuration/app/source-operations.md
```diff
@@ -0,0 +1,64 @@
...
+        spec: '0 0 * * *'
+        cmd: |
+            if [ "$PLATFORM_BRANCH" = update-dependencies ]; then
+                platform environment:sync code data --no-wait
```
pjcdawkins:
```suggestion
                platform environment:sync code data --no-wait --yes
```

src/configuration/app/source-operations.md
```diff
@@ -0,0 +1,64 @@
...
+                platform source-operation:run update
```
pjcdawkins:
```suggestion
                platform source-operation:run update --yes
```

src/configuration/app/source-operations.md
```diff
@@ -0,0 +1,64 @@
...
+
+The command has two parameters:
+
+* `operation` (string): the name of the operation
+* `variables` (optional, object): additional variables (in the style of variables defined in the `.platform.app.yaml`) to inject in the environment of the source operation
+
+When this operation is triggered:
+
+* A clean Git checkout of the current environment HEAD commit is created; this checkout doesn't have any remotes, has all the tags defined in the project, but only has the current environment branch.
+* Sequentially (_the sequence is in lexicographic order of the path of the `.platform.app.yaml` file_), for each application that has defined this operation, the operation command is launched in the container image of the application, with the same type of environment you will find in a build, with the addition of all the runtime environment variables of the current environment, optionally overridden by the variables specified in the operation payload.
```
ralt: maybe mention not `$PLATFORM_RELATIONSHIPS`?

Crell: Reworded that bullet to not be as clumsy, and listed only 2 extra env vars that are available.

src/configuration/app/source-operations.md
```diff
@@ -0,0 +1,68 @@
...
+An application can define a number of operations that apply to its source code and that can be automated.
+
+A basic, common source operation could be to automatically update Composer dependencies like this:
+
+```yaml
+source:
+    operations:
+        update:
+            command: |
+                composer update
```
ralt: You probably want to add a `set -e` there, and only run the `git add`/`git commit` when there is a change.

src/configuration/app/source-operations.md
```diff
@@ -0,0 +1,68 @@
...
+
+The `update` key is the name of the operation. It is arbitrary, and multiple source operations can be defined.
+
+The environment resource gets a new `source-operation` action which can be triggered by the CLI:
+
+```
+platform source-operation:run update
+```
+
+The command has two parameters:
```
ralt: It feels like you're talking about the CLI command there, but then you mention `variables`, so you actually mean the YAML? Maybe you should say "The `command` key"?

src/configuration/app/source-operations.md
```diff
@@ -0,0 +1,68 @@
...
+You can use cron to automatically run your source operations.
+
+> **note**
+>
+> Automated source operations using cron requires to [get an API token and install the CLI in your application container](/gettingstarted/cli/api-tokens.md).
+
+Once the CLI is installed in your application container and an API token configured you can add a cron task to run once a day and run your source operations. We do not recommend triggering source operations on your `master` production environment, but rather on a dedicated environment which you can use for testing before deployment.
+
+On the example below, we synchronize the `update-dependencies` environment with its parent before running the `update` source operation:
+```yaml
```
ralt: Missing newline before this?

src/configuration/app/source-operations.md
```diff
@@ -0,0 +1,74 @@
...
+            command: |
+                set -e
+                composer update
+                git add composer.lock
+                git commit -m "Update Composer dependencies."
+```
+
+The `update` key is the name of the operation. It is arbitrary, and multiple source operations can be defined.
+
+(You may with to include more robust error handling than this example.)
```
robertDouglass: You may wish

src/configuration/app/source-operations.md
```diff
@@ -0,0 +1,74 @@
+---
+search:
+    keywords: ['.platform.app.yaml', 'automated', 'code', 'source', 'update', 'operation', 'dependencies', 'auto']
+---
+
+# Source operations
+
+> **note**
+>
+> Source Operations are currently in beta.  While the syntax is not expected to change some behavior might in the future.
```
robertDouglass: expected to change, some behavior

src/configuration/app/source-operations.md
```diff
@@ -0,0 +1,74 @@
...
+
+(You may with to include more robust error handling than this example.)
+
+The environment resource gets a new `source-operation` action which can be triggered by the CLI:
+
+```bash
+platform source-operation:run update
+```
+
+The `source-operation:run` command takes the command name to run plus, optionally, additional variables to inject in the environment of the source operation.  They will be interpreted the same way as any other [variable](/development/variables.md) set through the UI or CLI, which means you need an `env:` prefix to expose them as a Unix environment variable.  They can then be referenced by the source operation like any other variable.
```
robertDouglass: command takes name of the command name to run. Additional variables can be added to inject into the environment of the source operation.

src/configuration/app/source-operations.md
```diff
@@ -0,0 +1,74 @@
...
+```
+
+When this operation is triggered:
+
+* A clean Git checkout of the current environment HEAD commit is created; this checkout doesn't have any remotes, has all the tags defined in the project, but only has the current environment branch.
+* Sequentially, for each application that has defined this operation, the operation command is launched in the container image of the application.  The environment will have all of the variables available during the build phase, plus `PLATFORM_BRANCH` and `PLATFORM_ENVIRONMENT`, optionally overridden by the variables specified in the operation payload.
+* At the end of the process, if any commits were created, the new commits are pushed to the repository and the normal build process of the environment is triggered.
+
+Note that these operations run in an isolated container: it is not part of the runtime cluster of the environment, and doesn't require the environment to be running.  Also be aware that if multiple applications in a single project both result in a new commit, that will appear as two distinct commits in the Git history but only a single new build/deploy cycle will occur.
```
robertDouglass: running. If multiple applications in a single project result in new commits, they will appear as distinct commits in the Git history but only a single new build/deploy cycle will occur.

src/configuration/app/source-operations.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Automated Source Operations using cron
+
+You can use cron to automatically run your source operations.
```
robertDouglass: Damien thought that this should not be documented.

robertDouglass: He said that doing that was a bad idea, and as a result, Augustin took that part out of the How-to guide.

robertDouglass: But double check; it might be that the objection was solely based on having the CLI in the environment as a privileged user, and not because of the pattern of running source operations from cron. In that case, as long as we remind the user to have a service account for installing the CLI, then we're OK.

Crell: Ugh.  I thought it was included here because it was safe.  We already tell people outright to add the CLI to all projects for production, so not doing it here seems dumb.

GuGuss: It is safe to use cron to trigger source operations. What's not is to use your main API token to install the CLI on your app container.

PR #1080, opened 2019-04-22
Remove 301redirect as it no longer exists.

verbeeksteven:
301redirect shut down December 2018.


verbeeksteven:
There is a paid service I am using NakedSSL. Not sure if paid services are applicable here.

PR #1093, opened 2019-05-06
Add docs for dotnet

galister:
I made these docs some time ago but they didn't make it to the site for some reason. They're now freshened up for .NET Core 2.2.

There's also an example project for forking: https://github.com/galister/platformsh-example-aspnetcore

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -67,6 +67,7 @@
 * [Python](languages/python.md)
 * [Ruby](languages/ruby.md)
 * [Go](languages/go.md)
+* [C# / .NET Core](languages/dotnet.md)
 * [Java (alpha)](languages/java.md)
 
 ## Development
--- a/src/languages/dotnet.md
+++ b/src/languages/dotnet.md
@@ -1,0 +1,81 @@
+# .NET Core
+
+
+Platform.sh supports deploying .NET Core applications.
+
+## Supported versions
+
+* 2.0
+* 2.1
+* 2.2
+
+To specify a dotnet container, use the `type` property in your `.platform.app.yaml`.
+
+```yaml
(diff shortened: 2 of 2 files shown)


src/SUMMARY.md
```diff
@@ -67,6 +67,7 @@
 * [Python](languages/python.md)
 * [Ruby](languages/ruby.md)
 * [Go](languages/go.md)
+* [C# / .NET Core](languages/dotnet.md)
```
chadwcarlson: In the main doc you only reference this runtime as `.NET Core`, and this is the only place where you use `C# /` before. I'm not very familiar, but should this sidebar entry also then just be `.NET Core`?

galister: because the menu item is called `Languages` but .NET is not a language but a framework.

under the frameworks menu we have exclusively php frameworks, such as drupal, wordpress or symfony, so it would not fit there.

src/languages/dotnet.md
```diff
@@ -0,0 +1,83 @@
...
+
+To specify a dotnet container, use the `type` property in your `.platform.app.yaml`.
+
+```yaml
+type: 'dotnet:2.2'
+```
+
+## Building the application
+
+Dotnet containers do not come with a pre-defined build process that you may have seen on other container types. Instead, the process is defined by you, using the runtime and output directory variables passed to the build environment.
```
NickAnderegg: What is this supposed to mean? As far as I'm aware, only PHP has a defined build process with native composer support, but I don't think it warrants this paragraph. Unless this is just referred to the `dependencies` block, in which case, I'd say something like "does not have automatic dependency installation" or something to that effect.

ralt: Only this container has a separate source and output folder.

PR #1095, opened 2019-05-15
Custom environment variables in the build hook.

matthiaz:
Adding information on how to use custom environment variables in the build hook.

### DIFF
--- a/src/configuration/app/build.md
+++ b/src/configuration/app/build.md
@@ -163,3 +163,32 @@ hooks:
         fi
         drush -y updatedb
 ```
+
+## How can I use my custom environment variables?
+
+Custom environment variables can be defined in the console. They will be available through the $PLATFORM_VARIABLES but to extract them, you need a couple of extra steps. 
+$PLATFORM_VARIABLES contains a base64 string. If you unpack that BASE64 string, you will get a JSON string of all your variables. You could then use a tool like jq to extract values from that JSON string.
+
+Concrete build hook example:
+
+```yaml
+hooks:
+    build: |
+        set -e
+        echo "Download jq (to be able to parse JSON) and make it executable."
+        curl -L -o $HOME/jq https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64
+        chmod +x $HOME/jq
+        echo ""
+        echo "Decode variables using base64 decode"
+        echo $PLATFORM_VARIABLES | base64 --decode
+        echo ""
+        echo "Extract the specific key you need"
+        echo ""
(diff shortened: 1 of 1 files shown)


Crell:
This is way too complicated for people to do.  If we want people to use jq, we should just pre-install it.  (We've discussed that before.)  If they want to access an individual variable directly, they should prefix it with `env:` so it's available as a bare variable and remove the need to do any decoding at all.

Frankly PLATFORM_VARIABLES is kind of vestigial at this point.

src/configuration/app/build.md
```diff
@@ -163,3 +163,32 @@ hooks:
         fi
         drush -y updatedb
 ```
+
+## How can I use my custom environment variables?
+
+Custom environment variables can be defined in the console. They will be available through the $PLATFORM_VARIABLES but to extract them, you need a couple of extra steps.
```
chadwcarlson: Custom environment variables can be defined using the [management console](/administration/web/configure-project.md#variables) or with the [Platform.sh CLI](/development/variables.md). They will be available within your during build as the base64-encoded JSON `PLATFORM_VARIABLES`. 

As an example, you can unpack `PLATFORM_VARIABLES` with a tool like `jq` in your build hook to access them:

```yaml
```

matthiaz:
> This is way too complicated for people to do. If we want people to use jq, we should just pre-install it. (We've discussed that before.) If they want to access an individual variable directly, they should prefix it with `env:` so it's available as a bare variable and remove the need to do any decoding at all.
> 
> Frankly PLATFORM_VARIABLES is kind of vestigial at this point.

I agree, jq should be added. But before we do, we still have to give the customer something to work with. (This documentation comes straight out of communication to certain customers)

Crell:
The answer for the 99.9999% case is "use an env: prefix on your variable and none of this is needed".  That's the answer we should be pointing people to, especially on the shell.

src/configuration/app/build.md
matthiaz: Err... I like your text better than mine. How do I accept it?

Crell:
It's a totally different approach so probably just a new PR.  Or we do nothing, because the env: prefix is already documented.

src/configuration/app/build.md
chadwcarlson: Bad reviewing here on my part. Second sentence should be:

> They will be available during build as the base64-encoded JSON `PLATFORM_VARIABLES`.

chadwcarlson: As far as I know use what you like of my edit, push it, then I approve the changes. Not sure if there's an "accept changes".

matthiaz:
@Crell is right. it is already mentioned in the documentation to use the prefix. So ignore this PR, I'll do a better job at RTFM :-)

PR #1097, opened 2019-05-16
Solr image updates.

chadwcarlson:
Add new Solr images, update in `services.yaml` example to 8.0


Crell:
There are several more mentions of 7.6 further down the file.  We should update those, too.

PR #1098, opened 2019-05-16
Link Node.js build info to language page.

chadwcarlson:
Discussion here: https://platformsh.slack.com/archives/CDCCS5ZUN/p1557969315416300

### DIFF
--- a/src/languages/nodejs.md
+++ b/src/languages/nodejs.md
@@ -90,15 +90,16 @@ To use Platform.sh and Node.js together, configure the ``.platform.app.yaml`` fi
      type: upstream
      upstream: "app:http"
    ```
-  
-7. (Optional) If Platform.sh detects a `package.json` file in your repository, it will automatically include a default [`build` process](/configuration/app/build.md#build) that will run `npm prune --userconfig .npmrc && npm install --userconfig .npmrc`. You can modify that process to run a custom Composer command or to use an alternate Node.js package manager 
+   
+7. (Optional) If Platform.sh detects a `package.json` file in your repository, it will automatically include a `default` [`build` flavor](/configuration/app/build.md#build), that will run `npm prune --userconfig .npmrc && npm install --userconfig .npmrc`. You can modify that process to use an alternative package manager by including the following in your `.platform.app.yaml` file:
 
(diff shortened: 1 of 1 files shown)


src/languages/nodejs.md
```diff
@@ -83,13 +83,22 @@ To use Platform.sh and Node.js together, configure the ``.platform.app.yaml`` fi
...
-    type: upstream
-    upstream: "app:http"
-  ```
+   ```yaml
+   "https://{default}/":
+     type: upstream
+     upstream: "app:http"
+   ```
+  
+7. (Optional) If Platform.sh detects a `package.json` file in your repository, it will automatically include a default [`build` process](/configuration/app/build.md#build) that will run `npm prune --userconfig .npmrc && npm install --userconfig .npmrc`. You can modify that process to run a custom Composer command or to use an alternate Node.js package manager
```
Crell: Composer isn't even available on the Node images, I think.  The only other option on Node is `none`.

chadwcarlson: That's the line used [here](https://docs.platform.sh/configuration/app/build.html#nodejs-default-by-default), do we need to update that as well?

Crell: Hm.  That's misleading, because the example is for the syntax of the whole block, not a part of the node description.  Node's only options are default and none.  Maybe we should move the example block up to above the PHP subsection?  (That's separate from this update.)

chadwcarlson: Okay that's good, we should modify it up then because that definitely looked like it belonged to Node. 

Proposal:

7. (Optional) If Platform.sh detects a `package.json` file in your repository, it will automatically include a `default` [`build` flavor](/configuration/app/build.md#build), that will run `npm prune --userconfig .npmrc && npm install --userconfig .npmrc`. You can modify that process to use an alternative package manager by including the following in your `.platform.app.yaml` file:

```yaml
build:
     flavor: none
```
Consult the documentation specific to [Node.js builds](https://docs.platform.sh/configuration/app/build.html#nodejs-default-by-default) for more information.

Crell: Endorse.

PR #1104, opened 2019-05-22
add a third case to the mysql has gone away section for more_mysql

alexmadon:
this is from ticket https://platformsh.zendesk.com/agent/tickets/57797

### DIFF
--- a/src/configuration/services/mysql.md
+++ b/src/configuration/services/mysql.md
@@ -224,7 +224,7 @@ SELECT
 
 #### Disk space issues
 
-Errors such as `PDO Exception 'MySQL server has gone away'` are usually simply the result of exhausting your existing diskspace. Be sure you have sufficient space allocated to the service in [.platform/services.yaml](/configuration/services.md).
+Errors such as "PDO Exception 'MySQL server has gone away'" are usually simply the result of exhausting your existing diskspace. Be sure you have sufficient space allocated to the service in [.platform/services.yaml](/configuration/services.md).
 
 The current disk usage can be checked using the CLI command `platform db:size`. Because of issues with the way InnoDB reports its size, this can out by up to 20%. As table space can grow rapidly, *it is usually advisable to make your database mount size twice the size reported by the `db:size` command*.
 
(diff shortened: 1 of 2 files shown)


src/configuration/services/mysql.md
```diff
@@ -222,16 +222,22 @@ SELECT
...
 
 Another possible cause of "MySQL server has gone away" errors is a server timeout.  MySQL has a built-in timeout for idle connections, which defaults to 10 minutes.  Most typical web connections end long before that is ever approached, but it's possible that a long-running worker may idle and not need the database for longer than the timeout.  In that case the same "server has gone away" message may appear.
 
 If that's the case, the best way to handle it is to wrap your connection logic in code that detects a "server has gone away" exception and tries to re-establish the connection.
 
 Alternatively, if your worker is idle for too long it can self-terminate.  Platform.sh will automatically restart the worker process, and the new process can establish its own new database connection.
+
+#### Packet size limitations
+
+Another cause of the 'MySQL server has gone away' errors can be the size of the database packets. If that is the case the logs may show warnings like  'Error while sending QUERY packet' before the error. One way to resolve the issue if to use the `max_allowed_packet` parameter described above.
```
otaviojava: May suggest a comma here?
If that is the case, ...

chadwcarlson: > "One way to resolve the issue ~if~ is to use ...

src/configuration/services/mysql.md
```diff
@@ -222,16 +222,22 @@ SELECT
...
+Another possible cause of `'MySQL server has gone away'` errors is a server timeout.  MySQL has a built-in timeout for idle connections, which defaults to 10 minutes.  Most typical web connections end long before that is ever approached, but it's possible that a long-running worker may idle and not need the database for longer than the timeout.  In that case the same "server has gone away" message may appear.
 
 If that's the case, the best way to handle it is to wrap your connection logic in code that detects a "server has gone away" exception and tries to re-establish the connection.
 
 Alternatively, if your worker is idle for too long it can self-terminate.  Platform.sh will automatically restart the worker process, and the new process can establish its own new database connection.
+
+#### Packet size limitations
+
+Another cause of the `'MySQL server has gone away'` errors can be the size of the database packets. If that is the case, the logs may show warnings like  `'Error while sending QUERY packet'` before the error. One way to resolve the issue is to use the `max_allowed_packet` parameter described above.
```
Crell: For all of these changes, don't mix backtick and single-quote.  We don't need both.  I'd actually say since these are text strings, not code strings, they should be in double-quotes.

src/development/troubleshoot.md
```diff
@@ -95,6 +95,41 @@ To find active background processes, run `ps aufx` on your application container
...
+
+```
+CREATE OR REPLACE SQL SECURITY INVOKER 
+VIEW `view_name` AS 
+SELECT 
+```
+
+## MySQL server has gone away
+
+#### Disk space issues
```
Crell: 3 #, not 4.  Same below.

Issue #1077, opened 2019-04-17
Issues with Using Solr with Drupal 8.x page


chrisolof:
I recently set this up and found three issues with the current "Using Solr with Drupal 8.x" doc page:

1. The Platform docs suggest running solr 6.x, yet the module suggests running 7.x.

   >For better performance and more features, 7.x should be used!

   I went with 7.x and I'd like to confirm Platform.sh's 7.6 solr appears to work perfectly with this module.

1. The Platform docs say to directly copy the solr config folder from the module folder to your .platform folder.  This was the way in D7, but no longer in D8.  From the search api solr docs:

   > Before starting the Solr server you will have to make sure it uses the proper configuration files. They aren't always static but vary on your Drupal setup. But the Search API Solr Search module will create the correct configs for you!
   > 
   > 1. Create a Search API Server according to the search_api documentation using "Solr" or "Multilingual Solr" as Backend and the "Standard" or "Basic Auth" Connector.
   > 1. Download the config.zip from the server's details page or by using `drush solr-gsc`
   > 1. Extract the config.zip to the conf directory of your new core.

   I can confirm that the generated files are not the same as those you'll get if you directly copy out of the module's config folder.  They are tailored to the site.

1. The generated/copied solrcore.properties config file contains a `solr.install.dir` line.  On Platform.sh, you must remove this line or your solr core won't be accessible to Drupal.  The old D7 `solr.install.dir` config file didn't contain this line and, as such, did not need adjustment to work on Platform.sh.  The module says this on the subject:

   > There's file called `solrcore.properties` within the set of generated config files. In most cases you have to adjust the `solr.install.dir` property to match your distribution specifics path, for example
   > 
   > ```
   > solr.install.dir=/opt/solr
   > ```

   I found that simply deleting the line was sufficient.  I'm guessing allows the server default to be used, which contains the correct path.

PR #1109, opened 2019-05-30
Remove the include/archive syntax as it's buggy

Crell:


src/configuration/yaml.md
```diff
@@ -167,3 +141,18 @@ web:
                 '^/sitemap\.xml$':
                     allow: true
 ```
+
+
+### `!archive`
+
+Another custom tag available is `!archive`, which specifies a value is a reference to a directory on disk, relative to the location of the YAML file.  Essentially it defines the value of key as "this entire directory".  Consider this `services.yaml` fragment:
```
otaviojava: suggestion: "Essentially it defines the importance of key as "this entire directory".  Consider this `services.yaml` fragment:"

src/configuration/yaml.md
```diff
@@ -167,3 +141,18 @@ web:
...
+
+```yaml
+mysearch:
+    type: solr:6.6
```
otaviojava: Maybe, update here to solr8 too

src/frameworks/drupal8/solr.md
```diff
@@ -40,9 +40,7 @@ solrsearch:
     configuration:
         cores:
             maincore:
-                core_config: !include
-                    type: archive
-                    path: "solr-conf/6.x"
+                conf_dir: !archive "solr-conf/6.x"
```
otaviojava: "solr-conf/8.x" ?

src/configuration/yaml.md
Crell: Mm, that doesn't fit in English.  Translation issue I think.

src/frameworks/drupal8/solr.md
Crell: I'm not sure about changing the Drupal page yet, as there's some weirdness with the newer Solr bridge I've not sorted out yet.  See #1077.  I'd rather not mess with the version examples there until we know what else needs to change.

PR #1110, opened 2019-05-30
Expand MySQL page to refer to both MariaDB and MySQL.

Crell:
Do not merge until Oracle-MySQL is tested and confirmed available on all regions.

| Region        | 5.7          | 8.0  |
| :------------- |:-------------:| :-----:|
| bc      |  |  |
| au      |      |    |
| fr-1 |     |    |
| us-2 |     |    |
| eu-2 |     |    |
| ca-1 |     |    |
| de-2 |     |    |
| eu-3 |     |    |

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -32,7 +32,7 @@
   * [Kafka](configuration/services/kafka.md)
   * [Memcached](configuration/services/memcached.md)
   * [MongoDB](configuration/services/mongodb.md)
-  * [MySQL](configuration/services/mysql.md)
+  * [MySQL/MariaDB](configuration/services/mysql.md)
   * [Network Storage](configuration/services/network-storage.md)
   * [PostgreSQL](configuration/services/postgresql.md)
   * [Redis](configuration/services/redis.md)
--- a/src/configuration/services/mysql.md
+++ b/src/configuration/services/mysql.md
@@ -1,24 +1,31 @@
 # MariaDB/MySQL (Database service)
 
-MariaDB is a MySQL-compatible relational database system. Its XtraDB storage engine is equivalent to MySQL with InnoDB.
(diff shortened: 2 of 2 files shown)


src/SUMMARY.md
```diff
@@ -32,7 +32,7 @@
   * [Kafka](configuration/services/kafka.md)
   * [Memcached](configuration/services/memcached.md)
   * [MongoDB](configuration/services/mongodb.md)
-  * [MySQL](configuration/services/mysql.md)
+  * [MySQL/MariaDB](configuration/services/mysql.md)
```
otaviojava: MariaDB is different from MySQL. Maybe put MariaDB then explain MariaDB is a fork of MySQL.

src/configuration/services/mysql.md
```diff
@@ -1,24 +1,31 @@
...
+
+* oracle-mysql:5.7
+* oracle-mysql:8.0
 
 > **note**
 >
-> Downgrades of MariaDB are not supported. MariaDB will update its own datafiles to a new version automatically but cannot downgrade them. If you want to experiment with a later version without committing to it use a non-master environment.
+> Downgrades of MySQL or MariaDB are not supported. Both will update their own datafiles to a new version automatically but cannot downgrade them. If you want to experiment with a later version without committing to it use a non-master environment.
+
+Other than the type, MySQL and MariaDB are otherwise identical and the rest of this page refers to both equally.
```
chadwcarlson: I'm good with this, but I think that this sentence should be above the version lists, something like:

> The service type `mysql` refers to MariaDB.  The service type `oracle-mysql` refers to MySQL as released by Oracle, Inc. Other than the type, MySQL and MariaDB are otherwise identical and the rest of this page refers to both equally.

Otherwise it gets buried to me.

src/SUMMARY.md
Crell: From our POV as a hosting company they're 99% identical.  The differences are all application level, hence the proposal to combine the doc pages that would otherwise be 99% identical.

PR #1115, opened 2019-06-06
Fix Django template links at Python language docs

jonatasbaldin:
Hey there!

Just removing a broken link from the docs and adding the two Django templates that I found in the GH org.

Cheers 🚀


src/languages/python.md
```diff
@@ -174,4 +174,5 @@ A number of project templates for Python applications are available on GitHub.
 * [Starter kit Python 3 minimal example](https://github.com/platformsh/platformsh-example-python-3)
 * [Starter kit Python with uwsgi](https://github.com/platformsh/platformsh-example-python-uwsgi)
 * [Starter kit Python and Elastic Search](https://github.com/platformsh/platformsh-example-python-elasticsearch)
-* [Starter kit Django project](https://github.com/platformsh/platformsh-example-django)
+* [Starter kit Django 1.11 project](https://github.com/platformsh/template-django1)
```
Crell: This line should be just Django 1, not 1.11 specifically.  It's just one more thing to have to update come 1.12.

Otherwise, :+1:

jonatasbaldin: Fixed!

PR #1117, opened 2019-06-09
make env:PLATFORMSH_CLI_TOKEN clearer

robertDouglass:
Is there any use-case where it wouldn't be `env:PLATFORMSH_CLI_TOKEN` ?

### DIFF
--- a/src/gettingstarted/cli/api-tokens.md
+++ b/src/gettingstarted/cli/api-tokens.md
@@ -16,7 +16,7 @@ You may be asked to reverify your password, then enter a unique application name
 
 After creating the token it will be displayed once at the top of the page in a green banner.  You may also view it later by clicking the "view" link next to the token name.  You will be asked to reverify your password as well when viewing the token.
 
-Now set that token to an environment variable named `PLATFORMSH_CLI_TOKEN` on your system where the CLI will run.  Consult the documentation for your CI system to see how to do that.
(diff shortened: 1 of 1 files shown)


chadwcarlson:
Is this the right syntax? (i.e. Circle CI doesn't require the `env:` prefix as far as I can tell. https://circleci.com/docs/2.0/env-vars/#example-configuration-of-environment-variables)

It seems correct if you are using the CLI within a Platform environment and configuring it through the management console, but the paragraph above doesn't seem to me to be talking about that, the next section does.

Crell:
There's a very subtle and annoying difference here.

The paragraph is talking about automating the CLI on *any* environment, in which case you need the Unix Environment Variable `PLATFORMSH_CLI_TOKEN` to be set.

On Platform.sh specifically, and only there, you set the Unix Environment Variable by creating a Platform.sh Variable (either Environment or Project level) for `env:PLATFORMSH_CLI_TOKEN`, which will result in a Unix Environment Variable named `PLATFORMSH_CLI_TOKEN` being created.

If you are using Circle CI, Travis, Jenkins, etc., then the `env:` prefix will make it not work.

This is a problem caused by our overloading of the terminology.  Line 39, a little further down, already describes this problem and how to address it.  If anything, we should not be making the docs Psh-specific but highlighting the Psh-specific bits earlier so people are more likely to notice them.

But any documentation solution is a bandaid only.

PR #1132, opened 2019-06-27
Improve documentation to show Java with full support at Platform.sh

otaviojava:
It requires: https://github.com/platformsh/language-examples/pull/26

### DIFF
--- a/src/configuration/services/influxdb.md
+++ b/src/configuration/services/influxdb.md
@@ -35,10 +35,6 @@ relationships:
 
 You can then use the service in a configuration file of your application with something like:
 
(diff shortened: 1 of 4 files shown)


src/configuration/services/influxdb.md
```diff
@@ -35,6 +35,10 @@ relationships:
 
 You can then use the service in a configuration file of your application with something like:
 
+{% codetabs name="Java", type="java", url="https://examples.docs.platform.sh/java/influxdb" -%}
```
chadwcarlson: Put PHP inside of another codetabs so they are both together.

PR #1133, opened 2019-07-01
WIP: Added a bit of documentation on how long snapshots take

matthiaz:


Crell:
According to @ralt this is wrong.  Should we just close it?

PR #1139, opened 2019-07-10
Update with some extra position info.

michielkenis:


Crell:
I have confirmed with the engineering team that this is not in fact true.  I am not sure why you were seeing this behavior but the YAML file is unordered.  (Putting those first is a good idea, just from a maintenance standpoint, but shouldn't impact behavior at all.)

PR #1142, opened 2019-07-10
Elasticsearch 7.2

Crell:
Pending confirmation from Engineering that it's fully tested.

### DIFF
--- a/src/changelog.md
+++ b/src/changelog.md
@@ -3,6 +3,9 @@
 ## 2019
 
 * **June 2019**
+  * Elasticsearch 7.2: We now support [Elasticsearch 7.2](/configuration/services/elasticsearc.md).
+
+* **July 2019**
   * Java: We support and documented the use of [Java](/languages/java.md) runtimes 8, 11, and 12, that includes examples that use the [Java Config Reader](https://github.com/platformsh/config-reader-java/).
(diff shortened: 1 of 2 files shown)


src/changelog.md
```diff
@@ -3,6 +3,9 @@
 ## 2019
 
 * **June 2019**
+  * Elasticsearch 7.2: We now support [Elasticsearch 7.2](/configuration/services/elasticsearc.md).
```
chadwcarlson: bad link - elasticsearch.md

src/changelog.md
```diff
@@ -3,6 +3,9 @@
 ## 2019
 
 * **June 2019**
+  * Elasticsearch 7.2: We now support [Elasticsearch 7.2](/configuration/services/elasticsearch.md).
```
chadwcarlson: If 5.2 and 5.4 are now deprecated, there should probably be an entry for them in the changelog too, like "[Elasticsearch](/configuration/services/elasticsearch.md) 5.2 and 5.4 support is now deprecated.

chadwcarlson: Plus I think June and July got switched, and this should be these should be the most recent additions in **July 2019**.

otaviojava: Fixed!

PR #1143, opened 2019-07-10
Note support for configurable Redis eviction policies.

Crell:
I debated putting the new section as a subsection of Ephemeral Redis, but that would have meant either no header for a 4th level header, which I want to avoid.  Open to discussion on that.

### DIFF
--- a/src/configuration/services/redis.md
+++ b/src/configuration/services/redis.md
@@ -33,7 +33,7 @@ rediscache:
     type: redis:5.0
 ```
 
-Data in an Ephemeral Redis instance is stored only in memory, and thus requires no disk space.  When the service hits its memory limit it will automatically evict old cache items to make room for new ones.
+Data in an Ephemeral Redis instance is stored only in memory, and thus requires no disk space.  When the service hits its memory limit it will automatically evict old cache items according to the [configured eviction rule](#eviction-policy) to make room for new ones.
 
 ### Persistent Redis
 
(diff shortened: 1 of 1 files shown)


src/configuration/services/redis.md
```diff
@@ -24,7 +24,7 @@ The following versions are available but are not receiving security updates from
 
 ### Ephemeral Redis
 
-The `redis` service type is configured to serve as a LRU cache with the eviction policy `allkeys-lru` - its storage is not persistent.  It is not suitable for use except as a disposable cache.
+The `redis` service type is configured to serve as a LRU cache; its storage is not persistent.  It is not suitable for use except as a disposable cache.
 
 To add an Ephemeral Redis service, specify it in your `.platform/services.yaml` file like so:
```
chadwcarlson: Since these two sections are separated, readers might not immediately know this is something that could be configured. Maybe we should link to that section in the next paragraph?

"Data in an Ephemeral Redis instance is stored only in memory, and thus requires no disk space. When the service hits its memory limit it will automatically evict old cache items by default to make room for new ones, but it is also possible to [configure a different eviction policy](/configuration/services/redis.md#setting-the-eviction-policy)."

chadwcarlson: As per your comment, I don't think it's necessary to add this as a subsection to Ephemeral Redis, so long as you link to it like I suggested.

PR #1146, opened 2019-07-12
Defines alias to mariaDB

otaviojava:


Crell:
Let's just call it mariadb only.  Then we can also note in an extra paragraph that `mysql` refers to mariadb for BC reasons.

otaviojava:
Thank you @Crell what do you think now?

src/configuration/services/mysql.md
```diff
@@ -6,13 +6,16 @@ See the [MariaDB documentation](https://mariadb.org/learn/) or [MySQL documentat
 
 ## Supported versions
 
-The service type `mysql` refers to MariaDB.  The service type `oracle-mysql` refers to MySQL as released by Oracle, Inc.  Other than the `type`, MySQL and MariaDB are otherwise identical and the rest of this page refers to both equally.
+The service type `mariadb` refers to MariaDB such as `mysql` that for compatibility reasons it refers to MariaDB as well.  The service type `oracle-mysql` refers to MySQL.  Other than the `type`, MySQL and MariaDB are otherwise identical and the rest of this page refers to both equally.
```
chadwcarlson: This paragraph seems a little jumbled to me. How about:

"The service types `mariadb` and `mysql` both refer to MariaDB for compatibility reasons. The service type `oracle-mysql` refers to MySQL as released by Oracle, Inc. Other than the `type`, MySQL and MariaDB are otherwise identical and the rest of this page refers to both equally."

otaviojava: Thank you, fixed

otaviojava:
moves it to a this branch: https://github.com/platformsh/platformsh-docs/pull/1152

PR #1147, opened 2019-07-15
Update nodejs available versions

ludoge:
Switch documented nodejs versions to major versions only.
Don't provide a (non-exhaustive) list of deprecated available versions.

### DIFF
--- a/src/languages/nodejs.md
+++ b/src/languages/nodejs.md
@@ -4,22 +4,15 @@ Node.js is a popular JavaScript runtime built on Chrome's V8 JavaScript engine. 
 
 ## Supported versions
 
-* 6.11
-* 8.9
+* 6
+* 8
 * 10
 
 If you need other versions, take a look at our [options for installing them with NVM](/languages/nodejs/nvm.html).
 
 ## Deprecated versions
 
-The following versions are available but are not receiving security updates from upstream, so their use is not recommended. They will be removed at some point in the future.
-
-* 0.12
-* 4.7
-* 4.8
-* 6.9
-* 6.10
-* 8.2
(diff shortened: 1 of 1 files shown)


src/languages/nodejs.md
```diff
@@ -4,22 +4,15 @@ Node.js is a popular JavaScript runtime built on Chrome's V8 JavaScript engine.
...
 
-The following versions are available but are not receiving security updates from upstream, so their use is not recommended. They will be removed at some point in the future.
-
-* 0.12
-* 4.7
-* 4.8
-* 6.9
-* 6.10
-* 8.2
+Some versions with a minor (such as 8.9) are available but are not receiving security updates from upstream, so their use is not recommended. They will be removed at some point in the future.
```
Crell: IMO we should still list the numbers as long as the containers themselves exist.

ludoge: That's a fair point, but the list there was nowhere near exhaustive. There are 20 available nodejs versions, I don't think we should list them all? A good compromise would be listing the latest deprecated minor (e.g. 8.11 but no 8.10 or 8.9), IMO

ralt: The list depends on the region. Do you really want to go down that rabbit hole? :)

Crell: Wait, we have 20 available node versions, or there are 20 node versions?  If we offered, why was that never documented?

If the former, then yeah, not listing all of them makes sense.

ralt: we have 20 available node versions

Crell: Dear lord...

PR #1151, opened 2019-07-19
updates kafka and Java version

otaviojava:


src/configuration/services/kafka.md
```diff
@@ -7,6 +7,7 @@ See the [Kafka documentation](https://kafka.apache.org/documentation/) for more
 ## Supported versions
 
 * 2.1
+* 2.2
 
 ## Relationship
```
chadwcarlson: The service.yaml example should have this most recent version in it as well, like you did here https://github.com/platformsh/platformsh-docs/pull/1142/files

src/languages/java.md
```diff
@@ -10,6 +10,7 @@ Java is a general-purpose programming language, and one of the most popular in t
 * 8
 * 11
 * 12
+* 13
```
chadwcarlson: Same as above, the example app.yaml should use 13 now

PR #1154, opened 2019-07-22
Alphabetize languages and services everywhere.

chadwcarlson:
Ordering for codetabs and sidebar alphabetized everywhere.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -58,17 +58,18 @@
 
 ## Languages
 
+
+* [Go](languages/go.md)
+* [Java](languages/java.md)
+* [Node.js](languages/nodejs.md)
+  * [Alternative Node.js install](languages/nodejs/nvm.md)
 * [PHP](languages/php.md)
   * [Extensions](languages/php/extensions.md)
   * [Performance tuning](languages/php/tuning.md)
   * [PHP.ini settings](languages/php/ini.md)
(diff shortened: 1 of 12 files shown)


Crell:
Alphabetizing the sidebar: :+1:.

Alphabetizing the tabs, I'm torn on.  On the one hand, it makes complete sense and is consistent.  On the other, PHP is still the vast majority of our customer base AFAIK, so making all of them pick another tab first seems like extra work for our most common users.

chadwcarlson:
@Crell I figured this would be a push back point for PHP, and it makes sense. It made sense to me at the time to have the order consistent everywhere, but if we still want PHP to be first, how should codetabs be ordered?

Should we alphabetize the sidebar and then rank codetabs based on language popularity with customers throughout? Do we know how each of these rank?

src/SUMMARY.md
```diff
@@ -58,17 +58,18 @@
 
 ## Languages
 
+
+* [Go](languages/go.md)
+* [Java](languages/java.md)
+* [Node.js](languages/nodejs.md)
+  * [Alternative Node.js install](languages/nodejs/nvm.md)
 * [PHP](languages/php.md)
   * [Extensions](languages/php/extensions.md)
```
otaviojava: Nice!!!
I would like to do something similar to Java soon.

Crell:
Rob has decreed(tm) that we go alphabetical, so alphabetical it is.

PR #1157, opened 2019-07-23
Recommend using machine users for automated tasks

GuGuss:


src/administration/snapshot-and-restore.md
```diff
@@ -80,11 +80,11 @@ Snapshots may be triggered by calling the CLI from an automated system such as J
 
 > **note**
 >
-> Snapshots using cron requires you to [install the CLI in your container and set up an API token](/gettingstarted/cli/api-tokens.md).
+> Automated snapshots using cron requires to [get an API token and install the CLI in your application container](/gettingstarted/cli/api-tokens.md).
```
pjcdawkins: Either restore "requires you to", or reword this ("requires to" doesn't work)

src/configuration/routes/https.md
```diff
@@ -135,13 +135,15 @@ tls:
             -----END CERTIFICATE-----
 ```
 
-## Automatic SSL certificate renewal
+### Automated SSL certificate renewal using Cron
 
-It is possible to set a variable from a cron task, which in turn will cause the site to redeploy.  If the Let's Encrypt certificate is due to expire in less than one month it will automatically renew at that time.  That makes it feasible to set up auto-renewal of the Let's Encrypt certificate.  The caveat is that, like any deploy, there is a very brief downtime (a few seconds, usually) so it's best to do during off-hours.
+If the Let's Encrypt certificate is due to expire in less than one month it will automatically renew at that time.  That makes it feasible to set up auto-renewal of the Let's Encrypt certificate.  The caveat is that, like any deploy, there is a very brief downtime (a few seconds, usually) so it's best to do during off-hours.
```
pjcdawkins:
```suggestion
If the Let's Encrypt certificate is due to expire in less than one month then it will be renewed automatically during a deployment.  That makes it feasible to set up regular auto-renewal of the Let's Encrypt certificate.  The caveat is that, like any deploy, there is a very brief downtime (a few seconds, usually) so it's best to do during off-hours.
```

src/configuration/routes/https.md
```diff
@@ -135,13 +135,15 @@ tls:
...
 
-You will first need to install the CLI in your application container.  See the section on [API tokens](/gettingstarted/cli/api-tokens.md) for instructions on how to do so.
+> **note**
+>
+> Automated SSL certificate renewal using cron requires to [get an API token and install the CLI in your application container](/gettingstarted/cli/api-tokens.md).
```
pjcdawkins: "requires to get" doesn't work, it needs to be "requires you to get" or "requires getting [and installing]"

src/gettingstarted/cli/api-tokens.md
```diff
@@ -19,28 +19,30 @@ After creating the token it will be displayed once at the top of the page in a g
...
  
 Consult the [Users](/administration/users.md) documentation for more information about the differences between access levels.
 
-## Automating the CLI on a Platform.sh environment
+## Install the CLI on a Platform.sh environment
 
 A common use case for an API token is to allow the Platform.sh CLI to be run on an app container, often via a cron hook.  An API token is necessary for authentication, but the CLI will be able to auto-detect the current project and environment.
 
-First, obtain an API token as above.  Set it as the [top-level](https://docs.platform.sh/development/variables.html#top-level-environment-variables) environment variable `env:PLATFORMSH_CLI_TOKEN` either through the management console or via the CLI, like so:
+First, create a `machine user` (see above) that you invite to your project. Then, login as that `machine user` to obtain an API token. Set this token as the [top-level](https://docs.platform.sh/development/variables.html#top-level-environment-variables) environment variable `env:PLATFORMSH_CLI_TOKEN` either through the management console or via the CLI, like so:
```
pjcdawkins:
```suggestion
First, create a `machine user` (see above) that you invite to your project. Then, log in as that `machine user` to obtain an API token. Set this token as the [top-level](https://docs.platform.sh/development/variables.html#top-level-environment-variables) environment variable `env:PLATFORMSH_CLI_TOKEN` either through the management console or via the CLI, like so:
```

src/gettingstarted/cli/api-tokens.md
```diff
@@ -19,28 +19,30 @@ After creating the token it will be displayed once at the top of the page in a g
...
 
-## Automation users
+## Machine users
 
-For projects with more than one user it is often a good idea to create a dedicated "automation user" to run automation tasks such as backups. Like human users, every user account needs its own unique email address.
- 
-The automation user can be given a very restrictive set of permissions limited to just its needed tasks. Backups, for instance, require `Admin` access but no SSH key, while checking out code from a CI server to run tests on it would require an SSH key but only `Reader` access.
- 
-It will also show up in logs and activity streams as a separate entry from human users. We recommend having a separate automation user for each project to be automated.
+For security reasons we recommend to create a dedicated `machine user` to run automation tasks such as taking backups, renewing SSL certificates or triggering source operations. We also recommend to create a unique `machine user` for each project to be automated.
```
Crell: Machine user is not a code string here, so it shouldn't be using code offsets.  Normal quotes are fine.

src/gettingstarted/cli/api-tokens.md
```diff
@@ -67,5 +69,4 @@ crons:
 ```
 
 > **note**
->
```
Crell: Stylistically we have the extra line break in most places.  Don't mess with it here.  If we want to standardize that docs-wide that would be a separate PR task for just that.

src/administration/snapshot-and-restore.md
```diff
@@ -80,11 +81,11 @@ Snapshots may be triggered by calling the CLI from an automated system such as J
 
 > **note**
 >
-> Snapshots using cron requires you to [install the CLI in your container and set up an API token](/gettingstarted/cli/api-tokens.md).
+> Automated snapshots using cron requires you to [get an API token and install the CLI in your application container](/gettingstarted/cli/api-tokens.md).
```
chadwcarlson: This file has moved to `/development/cli/api-tokens.md`. 

The redirect has been set-up, but since you'll have to resolve the new location conflict, you may want to update this link too.

Crell:
Deprecated in favor of #1159

PR #1162, opened 2019-07-30
Document the GlueScript integration

GuGuss:


src/administration/integrations/gluescripts.md
```diff
@@ -0,0 +1,94 @@
...
+```bash
+platform integration:add --type script --file .platform/scripts/SCRIPT.js
+```
+
+where:
+* `SCRIPT` is the name of your Javascript file
+
+> **note**
+>
+> The GlueScript integration can only be managed by the project owner.
```
ralt: project admins*

src/administration/integrations/gluescripts.md
```diff
@@ -0,0 +1,94 @@
...
+
+where:
+* `SLACK_PAYLOAD_URL` is the URL provided by Slack for your custom webhook
+
+### 2. Commit the script
+
+Commit the script to Git:
+
+```bash
+git add .platform/scripts/SCRIPT.js
```
ralt: It's unclear: this file has this?

```javascript
{
    "type": "script",
    "script": "var message = activity.text;
var color;
// rest of the code..."
}
```

?

In that case, it's rather a JSON file than a JS file, no?

GuGuss: JS is fine based on the syntax color scheme of my code editor. When the file is saved as JSON, it doesn't recognize the syntax :)

pjcdawkins: Now that you're using the CLI, the file contents should be plain JS:

```
var message = activity.text;
var color;
// rest of the code...
```

src/SUMMARY.md
```diff
@@ -148,6 +148,7 @@
   * [HipChat](administration/integrations/hipchat.md)
   * [Tideways](administration/integrations/tideways.md)
   * [Webhooks](administration/integrations/webhooks.md)
+  * [GlueScripts](administration/integrations/gluescripts.md)
```
chadwcarlson: GlueScript

src/administration/integrations/gluescripts.md
```diff
@@ -0,0 +1,94 @@
...
+  "script": "<CODE>"
+}
+```
+
+where:
+* `CODE` is the Javascript code of the integration
+
+#### Custom Slack integration
+
+For example, the GlueScript can perform a custom Slack integration as follow:
```
chadwcarlson: For example, GlueScript can perform a custom Slack integration as follows:

src/administration/integrations/gluescripts.md
Crell: Wait, so is the file a Javascript embedded in JSON, or is it now just straight up Javascript?

I hope the latter, because the former is borderline abusive in terms of UX. :smile:

Crell:
The original internal report on GlueScript said it was "basically Javascript", implying not really or had some caveats. What are those?  What ES version is supported?  "Javascript" is a very moving target so we need to be a *lot* more specific here about what is and is not supported.

And if it is just plain Javascript, then... can we just remove the name "GlueScript" entirely and call it Javascript and not confuse people with another language name they need to remember?

src/administration/integrations/gluescripts.md
OriPekelman: JS and resolved in the "Other PR". Closing this one because all discussions were resolved there.

OriPekelman:
Closing this one. Everything else is happening on #1163

PR #1164, opened 2019-07-30
Update and rename gluescripts.md to gluescript.md

robertDouglass:


robertDouglass:
This needs a LOT of work. First, the different types of documentation are all mashed up together and it's really hard for the reader to pull it apart.

An intro is needed that explains:

- What is this new thing and who is it for? What problems does it solve? Who chould use it? Who shouldn't use it?
- What gets run? When does it run? 
- Real-life use case examples

Then, a separate document needs to provide REFERENCE material for:
- The full lifecycle of GlueScript
- The full environment that scripts run in 
- All the objects we provide
- All the APIs there are
- Reference to the actual JS package we're using in case someone needs to see the nitty-gritty details of the language

robertDouglass:
I tried to rename the file gluescript.md

PR #1165, opened 2019-07-30
New Theming

ashplatform:


Crell:
The next/prev buttons that GitBook adds are kinda weird.  They sit at the bottom of the page on their own with odd spacing, and centered, so if there's only one shown it just sorta... sits there randomly.

Crell:
Can we push sidebar items a bit closer together?  They space out so far right now that even on my huge screen some of the sections are large enough that I don't see more than one section.  A full em between them is too much.  Maybe a half em?

ashplatform:
@Crell Can you add a screenshot.

ashplatform:
I think we the whole site needs a rebuild, the CSS code is weird to say the least. I think we should address these structural issues with a rebuild using Hugo. There is no mobile view for instance.

Crell:
As per previous comment image: It's not visually obvious which items are main level and which are a sub-page.  They're indented, but using the same font means they visually feel like they're part of the same list.  Reducing the line spacing may help there; I'm not sure.

Crell:
image

The logo block on the left feels like it's the highlighted or selected version of the title block to the right.  It subjectively feels very awkward, like there's something I am missing or should be noticing or something.  (Entirely emotional response, I know, but it's throwing me off.)

src/scripts/user-widget/README.md
```diff
@@ -0,0 +1,38 @@
+# User widget
```
Crell: The user widget is supposed to be pulled in as a git submodule, no?  This looks like a Git goof.

src/styles/styles.css
```diff
@@ -1,18 +1,13 @@
...
-  --brand-color-neural-blue: #ededef;
-  --brand-color-primary-blue: #145cc6;
-  --brand-color-primary-pink: #ffaeae;
+  --brand-color-primary-blue: #145CC6;
   --brand-color-light-blue: #d0def4;
-  --brand-color-light-pink: #ffeeee;
-  --brand-color-accent-red: #fb3728;
+  --brand-font-color-black: #49475F;
+  --brand-font-color-dark: #3E3D54;
+  --sidebar-bg-color: #EDEDEF;
```
Crell: I included these as a complete list of the brand-defined colors.  It's the same list used in all of our slide decks now.  Please leave them.  The intent is to then reference them from another variable definition for a particular stylesheet that can then be used there.  Eg, `--highlight-color: var(--brand-color-neutral-gray);`.

Crell: (And yeah, I didn't bring you up to speed on that before you dug in here.  That's my fault for forgetting about it until now.)

src/styles/styles.css
```diff
@@ -27,12 +22,15 @@ body {
 body,
 .book,
 .book-summary {
-  font-family: "Gotham SSm A", "Gotham SSm B", "Open Sans", "Helvetica Neue", Helvetica, Arial, sans-serif;
+  font-family: "Overpass", "Helvetica Neue", Helvetica, Arial, sans-serif;
```
Crell: This can use the CSS var reference.

src/styles/styles.css
```diff
@@ -27,12 +22,15 @@ body {
...
 }
+
+/*****************
+TOP BAR
+**********************/
 #platform-bar {
-  background: #000;
+  background: #302F45;
   color: #fff;
```
Crell: Can this also be a var reference, even if it's not one of the brand colors proper?

Crell:
As seen in previous screenshot, the search bar seems unnecessarily tall.  That's too much padding space around the text itself.

ashplatform:
@Crell I just pushed an update for the sidebar navigation, i think it's a little easier to differentiate now and I reduced the padding.

Regarding the next / back arrows, they seem to collapse to the bottom of the page based on the width of the viewable content.

ashplatform:
@Crell Are you happy to go live with what we currently have? Chad's post is meant to drop tomorrow .

src/styles/styles.css
```diff
@@ -502,34 +505,39 @@ section.examples-lists > div {
...
   background-color: var(--button-link-background-color);
-  border: 1px solid #302F45;
+  border: none;
   border-radius: 10px;
   font-family: 'Open Sans', sans-serif;
   text-decoration: none !important;
   margin: 10px 5px;
   padding: 24px 24px;
   display: inline-flex;
+  font-weight: 600;
```
chadwcarlson: So I think that this definition takes away the differences between the title/text lines in button-link.descriptive. The way it was previously was the title (first line) was bold, and the following line was plain text. This is probably a consequence some poorly written inheritance on my part.

src/styles/styles.css
```diff
@@ -540,6 +548,12 @@ a.button-link.next {
 
 a.button-link.descriptive {
   flex-direction: column;
+  color: #1A182A;
```
chadwcarlson: Maybe it's just me, but these buttons being the only ones that go grey->grey seems a little odd to me. What do you think about them being the same as button-link.next?

src/styles/styles.css
```diff
@@ -502,34 +505,39 @@ section.examples-lists > div {
...
-  font-weight: 600;
   align-items: center;
   justify-content: center;
 }
 
 a.button-link.prev {
-  font-weight: 600;
   align-items: center;
   justify-content: center;
+  color: #ffffff;
```
chadwcarlson: I can't see this one very well, especially with the white text. Is there a big reason we need to have back be different than next? I do like the next style change a lot.

PR #1166, opened 2019-08-02
Code integrations are not supported with Source Operations

GuGuss:


Crell:
Hrm.  This warning doesn't go far enough.  There's no might.  Source Ops are actually incompatible with an external integration, now that I think about it.  Which is... sucky.

The warning needs to be more definitive and higher in the page.

GuGuss:
> Hrm. This warning doesn't go far enough. There's no might. Source Ops are actually incompatible with an external integration, now that I think about it. Which is... sucky.
> 
> The warning needs to be more definitive and higher in the page.

@Crell Thx for the feedback. I proposed a new way of displaying this note (which could be a warning IMHO instead of a note). Let me know what you think.

src/configuration/app/source-operations.md
```diff
@@ -5,9 +5,10 @@ search:
 
 # Source operations
 
-> **note**
+> **warning**
 >
-> Source Operations are currently in beta.  While the syntax is not expected to change, some behavior might in the future.
+> Source Operations are currently in Beta. While the syntax is not expected to change, some behavior might in the future.
+> For example, external code integrations like GitHub, Gitlab or Bitbucket, is not yet compatible with source operations yet.
```
Crell: "not yet compatible with yet".  Pick one yet. :smile: 

That said, I don't think they're going to be compatible any time soon.  Not until we have bidirectional commit to a linked repo, which is a blocker for a lot of things and I've seen no movement on it.

Honestly I'd consider this two separate warnings.  One that it's in beta (note), and one that it doesn't work with integrations (warning).

src/configuration/app/source-operations.md
```diff
@@ -7,7 +7,9 @@ search:
 
 > **note**
 >
-> Source Operations are currently in beta.  While the syntax is not expected to change, some behavior might in the future.
+> Source Operations are currently in Beta.  While the syntax is not expected to change, some behavior might in the future.
+>
+> Also be aware that Source Operations are not compatible with [3rd party repository integrations](/adminsitration/integrations.md) at this time. If you use a 3rd party repository then any changes from Source Operations will be overwritten by the integration.
```
chadwcarlson:
```suggestion
> Also be aware that Source Operations are not compatible with [3rd party repository integrations](/administration/integrations.md) at this time. If you use a 3rd party repository then any changes from Source Operations will be overwritten by the integration.
```

PR #1120, opened 2019-06-12
Getting Started, Developing on Platform.sh, Going Live Guides

chadwcarlson:
## General
* Allows for the use of `.cast` and `.mp4` files in the documentation.
* Moves existing "Getting Started" sections to "Development", updating redirect routes for old links.
* Renames and moves "Your first project" to a static "Project templates" page for all languages, moved to "Development" section as well.

## Getting Started Guides
* Start with a template
   Leads the user through 5 steps to open a free account and initialize a template project through the management console. The guide ends with instructions for how to install the CLI and a "Next steps" landing page that links to other guides. 
* Import your own code 
* Developing on Platform.sh > Local development
* Developing on Platform.sh > Development environments
* Next Steps > External integrations 
   Landing page that links to the most popular Git integrations. I'm fine with leaving those sections where they already exist in the docs, but each of those pages could probably be improved at least with asciinema videos of running the `platform integration:add` command for each.
* Next Steps > Going live
   Follows the existing Going live docs closely, with lots of cross-links. If we have a domain I could use, I would like to include an ascii video of actually adding the domain to a project.

PR #1168, opened 2019-08-06
Fixes link and cursor on buttons

chadwcarlson:
Addresses:

DEVREL-219
DEVREL-223


Crell:
What the heck is going on with the user widget files?

chadwcarlson:
@Crell looks like it got included during https://github.com/platformsh/platformsh-docs/pull/1120, so I removed it - generated during `platform build`.

PR #1169, opened 2019-08-06
Manual registry and example config yamls generated during build

chadwcarlson:
Includes a new subdirectory (`/src/registry/`) that contains:

```
src/
    registry/
        images/               # Registry information for service/runtime images
           registry.json      # Manually edited (for now) JSON for each image.
```

The Nodes.js module that generates example configuration yaml files during build can be found here: devrel/docs/tools/pshregistry-parser/.


chadwcarlson:
Image versions are a little goofy right now, mostly because I didn't know how it was going to be received from the gitlab api. From the tests I've done so far, it's pretty easy to retrieve the versions for `/images/elasticsearch` like this

```
[ '0.90', '1.4', '1.7', '2.4', '5.2', '5.4', '6.5', '7.2' ]
```

so that property will likely change closer to that.

src/registry/images/registry.json
```diff
@@ -0,0 +1,751 @@
...
+      "service_name": "headless",
+      "url": "\/configuration\/services\/headless-chrome.html"
+    },
+    "endpoint": "http",
+    "name": "Headless Chrome",
+    "runtime": false,
+    "supported": [
+      {
+        "major": 73,
+        "minor": null
```
Crell: Are major and minor split in the Manifest files and thus in what we're likely to get in the future from an automated system?  Otherwise this doesn't seem that helpful to split, since the user doesn't care about that distinction.

src/registry/templates/registry.json
```diff
@@ -0,0 +1,627 @@
+{
```
Crell: I can't really follow the structure of the schema here.  Sometimes a service is the top level key, sometimes a language.  What's the structural intent here?

I would think a flatter list of `template name => info` would be easier, which can then have tags for what services/languages it includes.  That's then simple enough to filter on if we want just the templates that include mariadb or just the templates that use node, for example.

Crell: Also... should this file live rather in template-builder rather than docs?  It seems like we could auto-generate it off of template-builder itself just by parsing the YAML files.  (Famous last words, I know...)

Crell:
This PR seems to be missing any of the code to produce the snippets off of the registry file.  git add goof? :smile:

src/registry/templates/registry.json
Kazanir: The use of "templates" here is a namespace conflict with the actual templating system; they are not necessarily connected. I believe everything is relocated under the `images/` path now, which seems better to me.

These should not live in template-builder, which is _our implementation_ of the catalog and templating system. This registry is intended to be a system-wide registry of all available apps and services (this is why some of the languages are keys in the file, those are just types of application containers) and is not formally connected to template-builder or the templating process.

Kazanir: Ah, I see, the registry.json in `templates/` is different and serves as an index of which templates consume which service types.

chadwcarlson: Yes, this will be removed from this PR for now.

Kazanir:
I don't understand what happened here: https://github.com/platformsh/platformsh-docs/pull/1169/commits/923bd5e46ab7fc93ab08562864ae845abfbb3edf

The example files seem to no longer be present in either the source repository or the built site artifact after deployment. Where have they gone?

PR #1170, opened 2019-08-06
Updates description of mariadb/oracle mysql endpoints.

chadwcarlson:
Addresses:

DEVREL-207


src/configuration/services/mysql.md
```diff
@@ -12,10 +12,12 @@ The service types `mariadb` and `mysql` both refer to MariaDB for compatibility
 * mariadb:10.1
 * mariadb:10.2
 
+
```
Crell: Why the extra line break here?  Seems unnecessary.  One should be sufficient.

Crell:
What's with all the user widget files?

src/configuration/services/mysql.md
Crell: For posterity: Because GitBook is breaking the list formatting otherwise and making it like 5 different lists, because reasons.

PR #1171, opened 2019-08-07
Lisp

OriPekelman:
Still requires a bit of precision around handling .asd files, and would want some review.

### DIFF
--- a/src/languages/lisp.md
+++ b/src/languages/lisp.md
@@ -1,0 +1,56 @@
+# Lisp
+
+Platform.sh supports building and deploying applications written in Lisp using Common Lisp (the SBCL version) with ASDF and Quick Lisp support.  They are compiled during the Build hook phase, and support both committed dependencies and download-on-demand.
+
+## Supported versions
+
+* 1.5
+
+To specify a Lisp container, use the `type` property in your `.platform.app.yaml`.
+
+```yaml
+type: 'lisp:1.5'
+```
+
+## ASDF dependencies
+
+The recommended way to handle Lisp dependencies on Platform.sh is using ASDF. commit an `.asd` file in your repository and the system will:
+{to be completed}
+
+## Platform.sh variables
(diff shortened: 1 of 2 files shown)


src/languages/lisp.md
```diff
@@ -0,0 +1,168 @@
...
+```yaml
+relationships:
+  pg: postgresql:postgresql
+```
+
+The following would access that relationship, and provide your Lisp program the credentials to connect to a PostgreSQL instance. Add this to your `.asd` file:
+
+```lisp
+:depends-on (:postmodern)
+````
```
chadwcarlson:
```suggestion
```
```

File: src/languages/lisp.md [hunk 2]
```diff
@@ -0,0 +1,168 @@
...
+
+# Accessing Services
+
+The services configuration is available in the environment variable `PLATFORM_RELATIONSHIPS`. 
+
+To parse them, add the dependencies to your `.asd` file:
+
+```lisp
+:depends-on (:jsown :babel :s-base64)
+````
```
chadwcarlson:
```suggestion
```
```

chadwcarlson: Alright, tried to just suggest a change on these two points, but suggestions use Markdown ticks too, so it doesn't show up correctly. 

So don't commit my suggestion, since it looks like it will remove everything if you do.

For here and below, change the four ticks ("````") to three ticks ("```") so that they match the number at the beginning of the code block. Otherwise, the fourth shows up in docs on an extra line (https://pr-1171-gs7ujbi-652soceglkw4u.eu-3.platformsh.site/languages/lisp.html#accessing-services)

src/languages/lisp.md
chadwcarlson: Same as above. Don't accept the suggestion (unfortunately) - just remove one back-tick from the end of the code block.

PR #1176, opened 2019-08-13
Updating header on mobile 

ashplatform:
Hide user widget - it does hide the free trial link so this may not work
Also hides the title "User Documentation"

### DIFF
--- a/src/_layouts/website/page.html
+++ b/src/_layouts/website/page.html
@@ -25,7 +25,8 @@
     </div>
     <div class="site-name">
       <a href="/">
-        <span class="fa fa-home"></span>User Documentation
+        <span class="fa fa-home"></span>
+        <span class="title">User Documentation</span>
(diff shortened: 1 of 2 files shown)


Crell:
When I shrink my browser window super-thin, the header is still taking up a lot of space.  The hamburger is also not visible although it is at slightly larger widths.  The sidebar is also not collapsing so that's all I see.

image

ashplatform:
@Crell 

CSS updates to fix the collapsing issue on small viewports.

I think the menu will collapse on a mobile device, there is some device detection happening I think. The site really needs a complete rebuild, that would be my advice.

Crell:
*Much* nicer, and yes when I tested it on my phone the sidebar starts collapsed.

I don't disagree that it needs an overhaul at some point, but sadly that point is not today. :frowning_face:

PR #1180, opened 2019-08-14
Remove extra line so codeblock formats correctly.

chadwcarlson:
Gitbook is weird. Just because their was an extra line between the codeblock and the previous paragraph, the codeblock didn't generate properly.

Also adds yaml syntax call to earlier codeblock.


Crell:
Silly GitBook...

PR #1187, opened 2019-08-22
Document the new Services tab.

chadwcarlson:
Resolves DEVREL-275 and its child issues.


src/configuration/app-containers.md
```diff
@@ -123,3 +123,27 @@ web:
 ## Upgrading from previous versions of the configuration file.
 
 Although we make an effort to always maintain backward compatibility in the `.platform.app.yaml` format, we do from time to time [upgrade the file](/configuration/app/upgrading.md) and encourage you to upgrade as well.
+
+## Console access
+
+You can view a list of applications for an environment in the management console by selecting that environment from the dropdown at the top of the page, and then clicking the "Services" tab that appears above that environment's overview information.
+
+Each application container has a tab on the left, so select the app you are interested in.
```
Crell: This sentence reads rather clunky.

Crell:
```suggestion
Select the application container on the left to show more detailed information for it on the right.
```

src/configuration/app-containers.md
```diff
@@ -123,3 +123,27 @@ web:
 ## Upgrading from previous versions of the configuration file.
```
Crell: I don't think this is the right page for this information.  It should be with the other UI-tour pages.  The /configuration/ section is all about editing YAML files.

src/configuration/services.md
```diff
@@ -63,6 +63,24 @@ For example, the current default storage amount per project is 5GB (meaning 5120
 
 By default, Platform.sh will allocate CPU and memory resources to each container automatically.  Some services are optimized for high CPU load, some for high memory load.  By default, Platform.sh will try to allocate the largest "fair" size possible to all services, given the available resources on the plan.  That is not always optimal, however, and you can customize that behavior on any service or on any application container.  See the [application sizing](/configuration/app/size.md) page for more details.
 
+## Console access
```
Crell: Same comment as above.  This all belongs in "Administration", not Configuration.

src/configuration/routes.md
```diff
@@ -206,6 +206,12 @@ You can get a list of the configured routes of an environment by running `platfo
 
 ![Platform Routes CLI](/images/cli/platform-routes-cli.png)
 
+## Console access
```
Crell: Ibid.

Crell:
The images that are being replaced are twice the size of the old ones.  That's surprising.  And probably not good for page load time.

PR #1190, opened 2019-08-23
Build limitations

chadwcarlson:
There is a disk limit during the build phase that should be more visible.


src/development/troubleshoot.md
```diff
@@ -65,6 +65,17 @@ For a MariaDB database, the command `platform db:size` will give approximate dis
...
 
+## No space left on device
+
+During the build hook, you may run into the following error depending on the size of your application:
+
+```
+W: [Errno 28] No space left on device: ...
+```
+
+The cause of this issue has to do with the amount of disk provided to the build container before it is deployed. Build containers are restricted to 4 GB during build, no matter how much disk has been set aside for the deployed application.
```
Crell: Subtle wording nitpick here.  It's unclear from the phrasing here (and above) if this is referring to the disk space of the application itself, or the writeable mount.  More precisely, the build env is limited to 4 GB regardless of what the mounted disk size is, and thus by implication the application image is limited to 4 GB, too.  Can we rephrase "for the deployed application" a bit to make that distinction?

src/development/troubleshoot.md
```diff
@@ -65,6 +65,17 @@ For a MariaDB database, the command `platform db:size` will give approximate dis
...
+
+If for some reason your application requires more than 4 GB during build, you can open a support ticket to have this limit increased, however in practice the most amount of disk given during build still caps off at 8 GB in these cases.
```
Crell:
```suggestion
If for some reason your application requires more than 4 GB during build, you can open a support ticket to have this limit increased.  The most disk space available during build still caps off at 8 GB in these cases.
```

src/development/troubleshoot.md
```diff
@@ -65,6 +65,17 @@ For a MariaDB database, the command `platform db:size` will give approximate dis
...
 
+## No space left on device
+
+During the build hook, you may run into the following error depending on the size of your application:
+
+```
+W: [Errno 28] No space left on device: ...
+```
+
+The cause of this issue has to do with the amount of disk provided to the build container before it is deployed. Application images are restricted to 4 GB during build, no matter how much disk has been set aside for the deployed application.
```
Crell:
```suggestion
The cause of this issue has to do with the amount of disk provided to the build container before it is deployed. Application images are restricted to 4 GB during build, no matter how much writable disk has been set aside for the deployed application.
```

PR #1195, opened 2019-08-27
Rename "snapshot" to "backup" everywhere

pjcdawkins:
Depends on https://github.com/platformsh/platformsh-cli/pull/840


pjcdawkins:
The upstream change is now released (CLI 3.48.0), and gradually people will be updating (`platform self:stats` to see download stats per version). Maybe wait a day.

PR #1196, opened 2019-08-27
Java frameworks

otaviojava:
This documentations show how to use Platform.sh with the most popular Java framework.

* Spring
* Jakarta EE
* Eclipse MicroProfile
* Hibernate


Crell:
To be consistent with the other frameworks we mention directly, these should be under Featured Frameworks and have a page-per-service.  Putting them under the language feels awkward and inconsistent to me.

That said, can any of these be eliminated with code in the templates to begin with that only activate when a given relationship is available, or commented out and ready to be uncommented, or other "inline" mechanisms?  That would render many of these pages unnecessary.

otaviojava:
@Crell 
What do you think to move the other to its respective language?
I mean, Drupal related to PHP and Python with its related frameworks and so on.
Usually, When I look on related Java framework on a click in Java language.

PR #1198, opened 2019-08-30
WIP: Initial Elixir Docs

OriPekelman:


src/languages/elixir.md
```diff
@@ -0,0 +1,135 @@
+# Elixir
+
+Platform.sh supports building and deploying applications written in Elixir.  They are automatically compiled during the Build phase, and support both committed dependencies and download-on-demand. The underlying Erlang version is 
+22.0.7.
+
+## Supported versions
+
+* 1.9
```
Crell: Didn't the previous paragraph just say 22.0.7?

src/languages/elixir.md
```diff
@@ -0,0 +1,135 @@
...
+- Your `mix.exs` file is at the root of your application.
+
+Platform.sh will then run `mix do deps.get, deps.compile, compile` on your application to build a binary.
+
+If you don't want these assumptions, you can disable this behavior by specifying in your `.platform.app.yaml`:
+
+```yaml
+build:
+    flavor: none
+```
```
Crell: Wait, why are we using a flavor here?  We have moved away from it for everything since Node, because it's too much magic.  Why is this back?  That's very inconsistent.

src/languages/elixir.md
```diff
@@ -0,0 +1,135 @@
...
+    [
+	  {:platformshconfig, "~> 0.1.0"}
+    ]
+  end
+```
+
+
+## Platform.sh variables
+
+Platform.sh exposes relationships and other configuration as [environment variables](/development/variables.md). Most notably, it allows a program to determine at runtime what HTTP port it should listen on and what the credentials are to access [other services](/configuration/services.md).
```
Crell: This is redundant with the rest of the docs.

src/languages/elixir.md
```diff
@@ -0,0 +1,135 @@
...
+
+To get the `PORT` environment variable (the port on which your web application is supposed to listen) you would:
+
+```elixir
+String.to_integer(System.get_env("PORT") || "8888")
+```
+
+Some of the environment variables are in Json format and are base64 encoded. You would need to import a Json parsing library such as Jason or Poison to read those.
```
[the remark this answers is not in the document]
Crell: Although again, this is redundant with the rest of the docs.

src/languages/elixir.md
```diff
@@ -0,0 +1,135 @@
...
+disk: 512
+```
+
+Note that there will still be a proxy server in front of your application.  If desired, certain paths may be served directly by our router without hitting your application (for static files, primarily) or you may route all requests to the Elixir application unconditionally, as in the example above.
+
+# Accessing Services
+
+The simplest possible way to go around this is to use the [Platform.sh Config Reader](https://hex.pm/packages/platformshconfig) library from hex.
+
+If you are building a Phoenix app for example it would suffice to add a database to `.platform/services.yaml` and a relationship in `.platform.app.yaml`  put the lib in you `deps` and assuming you renamed the proc.secrets.exs to releases.exs per the[Phoenix guide](https://hexdocs.pm/phoenix/releases.html) change:
```
Crell:
```suggestion
If you are building a Phoenix app for example it would suffice to add a database to `.platform/services.yaml` and a relationship in `.platform.app.yaml`  put the lib in you `deps` and assuming you renamed the `proc.secrets.exs` to `releases.exs` per the[Phoenix guide](https://hexdocs.pm/phoenix/releases.html) change:
```

src/languages/elixir.md
```diff
@@ -0,0 +1,135 @@
...
+
+# Accessing Services Manually
+
+The services configuration is available in the environment variable `PLATFORM_RELATIONSHIPS`. 
+
+Given a relationship defined in `.platform.app.yaml`:
+
+```yaml
+relationships:
+  postgresql: postgresql:postgresql
```
Crell:
```suggestion
  database: db:postgresql
```

The newly-standard naming convention.

src/languages/elixir.md
```diff
@@ -0,0 +1,135 @@
...
+      {:poison, "~> 3.0"}
+    ]
+  end
+```
+
+And assuming you use `ecto` you could put in `config/config.exs`:
+
+```elixir
+relationships = Poison.decode!(Base.decode64!(System.get_env("PLATFORM_RELATIONSHIPS")))
+[postgresql_config | _tail] = relationships["postgresql"]
```
Crell:
```suggestion
[postgresql_config | _tail] = relationships["database"]
```

src/languages/elixir.md
ludoge: 22.0.7 is the version of Erlang; 1.9 is the version of Elixir, which runs on the Erlang VM so has a dependency on it.

src/languages/elixir.md
OriPekelman: Florian wanted a flavour; I was not in favor .. @ralt ... even more so that I need to make it quite magical (ie either two flavours or I need   a big if/else block there).
But I am good either way.

src/languages/elixir.md
OriPekelman: OK We can just keep the note about when stuff is evaluated..

src/languages/elixir.md
OriPekelman: Agreed.

OriPekelman:
I am going to review the whole thing because my lib behaves just a bit differently now.. Also we need to decide about the build flavor.

src/languages/elixir.md
```diff
@@ -0,0 +1,127 @@
...
+web:
+    commands:
+        start: mix run --no-halt
+    locations:
+        /:
+            allow: false
+            passthru: true
+variables:
+    env:
+        MIX_ENV: 'prod'
```
OriPekelman: This is probably no longer necessary, I think we are exporting it now by default

src/languages/elixir.md
```diff
@@ -0,0 +1,127 @@
...
+
+{% codesnippet "/registry/images/examples/full/elixir.app.yaml", language="yaml" %}{% endcodesnippet %}
+
+## Assumptions
+
+Platform.sh is making assumptions about your application to provide a more streamlined experience. These assumptions are the following:
+
+- Your `mix.exs` file is at the root of your application.
+
+Platform.sh will then run `mix do deps.get, deps.compile, compile` on your application to build a binary.
```
robertDouglass: Is it important to note that it installs rebar? I guess by calling `mix local.rebar`
https://lab.plat.farm/images/elixir/-/blob/1.9/files/etc/platform/flavor.d/default.py#L43

chadwcarlson: I added a note about the default build

PR #1199, opened 2019-08-30
Adds pshregistry-parser to generate config files on build.

chadwcarlson:
Includes a new subdirectory (`/src/registry/`) that contains:

```
src/
    registry/
        images/               # Registry information for service/runtime images
           registry.json      # Manually edited (for now) JSON for each image.
           registry.yaml      # Requested yaml version of the same file.
```

After build, the registry will be available at `docs.platform.sh/registry/images/registry.json`, but you should not rely on it staying there for your use case. This file will eventually be generated elsewhere, requested during build, and ultimately removed from the repository and deployed docs.

Example configuration files will sit in `registry/images/examples` and have three types: `commented`, `full`, and `snippet`. 

A heavily commented example `.platform/services.yaml` file for Varnish would be available at `docs.platform.sh/registry/images/examples/commented/varnish.services.yaml`, for example.

To find out which files are in each subdirectory (i.e. `commented/`), there is a `content.json` file in each that lists its contents (`docs.platform.sh/registry/images/examples/commented/content.json`).

The Nodes.js module that generates example configuration yaml files during build can be found here: `devrel/docs/tools/pshregistry-parser/`.


src/scripts/updateExampleConfigFiles.js
```diff
@@ -0,0 +1,7 @@
+const psh = require("pshregistry-parser");
+
+
+let registrySource = "_book/registry/images/registry.json";
+let registry = new psh.RegistryParser(registrySource);
```
Crell: Nit: Both of these can be `const`.

Crell:
With the latest tweak to the parser library, is this ready to roll?

PR #1201, opened 2019-09-04
Addition: How to access the rabbitmq management plugin via browser

benjaminhirsch:


src/configuration/services/rabbitmq.md
```diff
@@ -67,6 +67,15 @@ If your service is running on a different port, you can re-open your SSH session
...
 Finally, while the session is open, you can launch a RabbitMQ client of your choice from your local workstation, configured to connect to `localhost:5672` using the username and password you found in the relationship variable.
 
+## Access the management plugin  (Web UI)
+In case you want to access the browser-based UI, you have to use an SSH tunnel. To open a tunnel, log into your application container like usual, but with an extra flag to enable local port forwarding:
+
+```bash
+ssh -L 15672:mq.internal:15672 <projectid>-<branch_ID>@ssh.eu.platform.sh
+```
+
+After you successfully established a connection, you should be able to open http://localhost:15672 in your browser. You'll find the credentials like mentioned above.
```
Crell: Why not just direct people to `platform tunnel:open`, which opens tunnels for everything?

benjaminhirsch: Good point, should I'll add that as an extra note/hint or replace it? I'm asking because using platform tunnel:open  would also open the tunnel for the AMQP Port which would make the whole ssh part under "_From your local development environment_" obsolete, or I'm wrong?

Crell: :thinking: Hrm.  The admin port is a separate port from the access port, so no, it wouldn't be exposed by `platform tunnel`.  That said, you can only SSH to something that has a defined relationship, and last I checked we don't have an endpoint type for the admin port.

Have you tested that this works on Psh in the first place?

benjaminhirsch: I'll test it later this day and will give you feedback. :)

benjaminhirsch: OK I'll tested a bit arround with the following config:

**services.yaml:**  
```
rediscache:  
    type: redis:5.0  
database:  
    type: mysql:10.2  
    disk: 512  
rabbitmq:  
    type: rabbitmq:3.7  
    disk: 1024
```

**.platform.app.yaml**  
```
relationships:  
    applicationcache: "rediscache:redis"  
    database: "database:mysql"  
    rabbitmq: "rabbitmq:rabbitmq"
```

If I use platform tunnel:open, it opens me tunnels for redis and mysql, but not for rabbitmq. I guess this is not working as intended, right? Maybe I should then open an ticket @ platformsh-cli?

OriPekelman: The 15672 port is not exposed in the relationships, so the CLI does not open a tunnel to it. It is available and accessible.  So for the moment this documentation proposition is correct.

Crell: Ah, I didn't realize the port was actually open.  Good to know.  I'm good with this PR then.

PR #1203, opened 2019-09-05
Update data-retention.md

tristanblattmansh:
Update document truncate to 100MB


Crell:
So far you're winning the trivial PR of the month award. :smile:

PR #1208, opened 2019-09-06
Pull generated configuration files into docs

chadwcarlson:
I have left in the original example snippets for the time being, so that we can compare and make adjustments to pshregistry-parser if necessary. 

There have been no changes to the service/relationship name conventions at this point - will be in separate PR.


.platform.app.yaml
```diff
@@ -26,14 +26,19 @@ hooks:
...
       gitbook build
 
       # Gitbook doesn't allow for a custom favicon except through a theme plugin. So Plan B.
       cp src/images/favicon.ico _book/gitbook/images/favicon.ico
 
-      # Build example configuration files from the Registry.
-      node _book/scripts/updateExampleConfigFiles.js
+      # Move example configuration file API points to _book.
+      mv src/registry/images/examples _book/registry/images
+      mv src/registry/images/registry.yaml _book/registry/images/
```
Crell: Wouldn't they get pulled over automatically if they're built in src before the book is built?

.gitignore
```diff
@@ -6,3 +6,5 @@ _books/*
 node_modules/*
 .idea/*
 .platform/local/*
+src/registry/images/examples/*
```
chadwcarlson: @Crell They will so long as these two lines are removed. I included them as a sort of `platform build` failsafe and worked around this with the move commands. I don't care what we do here, but I still will occasionally forget and commit user-widget on PRs. I figured, .gitignore these, otherwise that situation is user-widget + 120 example files.

.platform.app.yaml
Crell: The git commit stuff is a red herring.  That's all irrelevant to this move command (which should maybe be a recursive copy?)  The issue here is that the files are built above, but not copied over by GitBook automatically so we have to do it manually.  That the directory is gitignored is beside the point.

PR #1210, opened 2019-09-10
Re-add dotnet docs.

chadwcarlson:
Reopens https://github.com/platformsh/platformsh-docs/pull/1093 without conflicts and small changes.

### DIFF
--- a/src/languages/dotnet.md
+++ b/src/languages/dotnet.md
@@ -75,9 +75,3 @@ web:
   commands:
     start: ./WebApplication1
 ```
-
-## Project templates
-
-Platform.sh offers a project template for ASP.NET Core applications using the structure described above. It can be used as a starting point or reference for building your own application.
-
-[ASP.NET Core](https://github.com/platformsh/template-aspnet-core)
--- a/src/languages/dotnet.md
+++ b/src/languages/dotnet.md
@@ -23,7 +23,7 @@ hooks:
         dotnet publish --self-contained --runtime "$DOTNET_DEFAULT_RUNTIME" --output "$PLATFORM_OUTPUT_DIR" -p:UseRazorBuildServer=false -p:UseSharedCompilation=false
 ```
 
-those environment variables are defined as:
+Those environment variables are defined as:
 
 * `DOTNET_DEFAULT_RUNTIME`: specifies the target runtime used by application containers on Platform.sh (Debian).
 * `PLATFORM_OUTPUT_DIR`: the output directory for compiled languages available at build time.
--- a/src/languages/dotnet.md
+++ b/src/languages/dotnet.md
@@ -23,9 +23,14 @@ hooks:
         dotnet publish --self-contained --runtime "$DOTNET_DEFAULT_RUNTIME" --output "$PLATFORM_OUTPUT_DIR" -p:UseRazorBuildServer=false -p:UseSharedCompilation=false
 ```
 
+those environment variables are defined as:
+
+* `DOTNET_DEFAULT_RUNTIME`: specifies the target runtime used by application containers on Platform.sh (Debian).
+* `PLATFORM_OUTPUT_DIR`: the output directory for compiled languages available at build time.
+
 Typically .NET builds will start a collection of build servers, which are helpful for repeated builds. On Platform.sh, however, if this process is not disabled, the build process will not finish until the idle timeout is reached.
 
-Because of this it is recommended to include `-p` toggles that disable the Razor compiler for dynamic cshtml pages (`UseRazorBuildServer`) and the .NET msbuild compiler (`UseSharedCompilation`).
+As a result, it is recommended to include `-p` toggles that disable the Razor compiler for dynamic cshtml pages (`UseRazorBuildServer`) and the .NET msbuild compiler (`UseSharedCompilation`).
 
 If making multiple builds is desired for your application, make sure to call `dotnet build-server shutdown` at the end of your build hook.
 
--- a/src/languages/dotnet.md
+++ b/src/languages/dotnet.md
@@ -23,7 +23,11 @@ hooks:
         dotnet publish --self-contained --runtime "$DOTNET_DEFAULT_RUNTIME" --output "$PLATFORM_OUTPUT_DIR" -p:UseRazorBuildServer=false -p:UseSharedCompilation=false
 ```
 
-When using a custom build hook, make sure to copy the `-p` toggles over, or builds will take much longer. Alternatively, when making multiple builds within one build hook, you might want to leave the servers enabled and call `dotnet build-server shutdown` at the end of your build hook.
+Typically .NET builds will start a collection of build servers, which are helpful for repeated builds. On Platform.sh, however, if this process is not disabled, the build process will not finish until the idle timeout is reached.
+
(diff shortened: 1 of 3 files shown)


src/languages/dotnet.md
```diff
@@ -0,0 +1,78 @@
+# .NET
+
+Platform.sh supports deploying .NET applications by allowing developers to define a build process and pass its variables to the .NET build environment.
+
+## Supported versions
+
+* 2.0
+* 2.1
+* 2.2
```
Crell: Shouldn't this be using the new registry now? :smile:

src/languages/dotnet.md
```diff
@@ -0,0 +1,78 @@
...
+
+To specify a .NET container, use the `type` property in your `.platform.app.yaml`.
+
+{% codesnippet "/registry/images/examples/full/dotnet.app.yaml", language="yaml" %}{% endcodesnippet %}
+
+## Building the application
+
+.NET containers do not come with a pre-defined build process that you may have seen on other container types. Instead, the process is defined by you, using the runtime and output directory variables passed to the build environment.
```
Crell: I'm not sure this is relevant.  Only two container types, PHP and node, include "flavors" as we've been moving away them.  We don't have this comment on any other language AFAIK.

src/languages/dotnet.md
chadwcarlson: Absolutely, haven't gotten to pulling in the supportedHTML element yet on any of them.

src/languages/dotnet.md
```diff
@@ -0,0 +1,78 @@
...
+
+For simple applications, we recommend using `dotnet publish --self-contained`, while specifying the target runtime and output folders via environment variables:
```
Crell: Avoid the use of first-person.  This can be reworded.  "For simple applications, `blah` is sufficient and allows you to specify..."  Or something like that.

src/languages/dotnet.md
```diff
@@ -0,0 +1,78 @@
...
+
+Incoming requests are passed to the application using either a TCP (default) or UNIX socket. The application must use the [appropriate environment variable](/configuration/app/web.html#socket-family) to determine the URI to listen on. In case of a TCP socket ([recommended](https://go.microsoft.com/fwlink/?linkid=874850)), the application must listen on `http://127.0.0.1`, using port defined by the `PORT` environment variable.
+
+
+There will be an Nginx server sitting in front of your application. We recommend serving static content via Nginx, as this allows easy control of headers (including cache headers), and also has marginal performance benefits. Note that HTTPS is also terminated at the Ngnix proxy, so the `app.UseHttpsRedirection();` line (which is in `Startup.cs` by default) should be removed. To force HTTPS-only, please refer to the [routes documentation](/configuration/routes/https.html#https).
```
Crell: The parenthetical here should not be parenthetical.  The sentence doesn't really make sense without it.  "the `blah` line in `Startup.cs` should be removed"

src/languages/dotnet.md
```diff
@@ -0,0 +1,78 @@
...
+      passthru: true
+      rules:
+        # Serve these common asset types with customs cache headers.
+        \.(jpe?g|png|gif|svgz?|css|js|map|ico|bmp|eot|woff2?|otf|ttf)$:
+          allow: true
+          expires: 300s
+
+  commands:
+    start: |
+      ./WebApplication1
```
Crell: No need to make this a multi-line block when it's a single line.

How does one control the output file name?

src/languages/dotnet.md
```diff
@@ -0,0 +1,78 @@
...
+web:
+  locations:
+    "/":
+      allow: false
+      passthru: true
+
+  commands:
+    start: |
+    ./WebApplication1
+```
```
Crell: Do we explain how to do this on any other language?

(Honestly we should revisit the language pages sometime soon.)

src/languages/dotnet.md
Crell: Oh!  OK, I thought we'd done that part.

src/languages/dotnet.md
chadwcarlson: Removing.

src/languages/dotnet.md
chadwcarlson: Updated. From what I can tell it's taken from the name you give this file and how you define within it https://github.com/platformsh/template-builder/blob/aspnet/templates/aspnet-core/files/PlatformshAspNetCore.sln

Crell: Neither "Web" nor "Application" appear in that file. So, um, huh?

chadwcarlson: Oh I'm just saying that that file would be called `WebApplication1.sln` with that name throughout this line 
```
Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "PlatformshAspNetCore", "PlatformshAspNetCore.csproj", "{6150B0DB-45B5-4429-9410-2BA4E68874CC}"
```
so then that defines https://github.com/platformsh/template-builder/blob/446ccc0925747318576f765163a02b33190edf24/templates/aspnet-core/files/.platform.app.yaml#L28

src/languages/dotnet.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Building the application
+
+For simple applications, using `dotnet publish --self-contained` is sufficient for building applications in .NET containers, and it allows you to specify the target runtime and output folders via environment variables:
+
+```yaml
+hooks:
+    build: |
+        set -xe
+        dotnet publish --self-contained --runtime "$DOTNET_DEFAULT_RUNTIME" --output "$PLATFORM_OUTPUT_DIR" -p:UseRazorBuildServer=false -p:UseSharedCompilation=false
```
Crell: These aren't already set, are they?  Are users expected to set these?  If so, that should be documented very explicitly right here.  (I'm... not sure why you would want do do that via custom env vars frankly but that's another story.)

src/languages/dotnet.md
```diff
@@ -0,0 +1,78 @@
...
+```
+
+Typically .NET builds will start a collection of build servers, which are helpful for repeated builds. On Platform.sh, however, if this process is not disabled, the build process will not finish until the idle timeout is reached.
+
+Because of this it is recommended to include `-p` toggles that disable the Razor compiler for dynamic cshtml pages (`UseRazorBuildServer`) and the .NET msbuild compiler (`UseSharedCompilation`).
```
Crell:
```suggestion
As a result, it is recommended to include `-p` toggles that disable the Razor compiler for dynamic cshtml pages (`UseRazorBuildServer`) and the .NET msbuild compiler (`UseSharedCompilation`).
```

src/languages/dotnet.md
Crell: Ah, ok.

src/languages/dotnet.md
```diff
@@ -0,0 +1,83 @@
...
+
+  commands:
+    start: ./WebApplication1
+```
+
+## Project templates
+
+Platform.sh offers a project template for ASP.NET Core applications using the structure described above. It can be used as a starting point or reference for building your own application.
+
+[ASP.NET Core](https://github.com/platformsh/template-aspnet-core)
```
Crell: We should probably not merge this until that repo is populated.  Or remove this for now and re-add it when that template is resolved.

I'd favor the latter, I think.

PR #1213, opened 2019-09-16
Overhaul the Drupal 8 Solr page for the latest changes to the module.

Crell:
This is a mess, but it seems to work in testing.

### DIFF
--- a/src/frameworks/drupal8/solr.md
+++ b/src/frameworks/drupal8/solr.md
@@ -1,71 +1,62 @@
 # Using Solr with Drupal 8.x
 
-Platform.sh recommends using Solr version 6.6, yet we support [other versions](../../configuration/services/solr.md#supported-versions). The following instructions are for Solr 6, but they also work for the previous versions if you change the version numbers accordingly.
+The Drupal [Search API Solr](https://www.drupal.org/project/search_api_solr) module has a somewhat involved setup process, as it requires live access to the Solr server in order to generate the configuration files for it.  The following procedure is therefore necessary to ensure each step is able to proceed.
+
+Search API Solr stores its configuration in the Drupal Configuration API.  However, that system does not easily support environment-aware information.  The setup process therefore depends on config-overrides in `settings.platformsh.php`, which may need to be modified slightly depending on  your Solr configuration.
+
+Search API Solr requires Solr 6.6 or higher, and recommends Solr 8 or higher.
 
 Advanced Solr service configuration and implementation in other frameworks other than Drupal can be found at the [Solr services page](../../configuration/services/solr.md)
 
-## Requirements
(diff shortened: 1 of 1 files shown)


Crell:
The diff isn't all that helpful, so here's the rendered page in question:

https://pr-1213-jjusj5a-652soceglkw4u.eu-3.platformsh.site/frameworks/drupal8/solr.html

src/frameworks/drupal8/solr.md
```diff
@@ -1,71 +1,62 @@
...
 relationships:
-    solr: 'solrsearch:main'
+    solrsearch: 'search:main'
 ```
 
-That is, the application's environment would include a `solr` relationship that connects to the `main` endpoint, which is the `maincore` core.
+That is, the application's environment would include a `solrsearch` relationship that connects to the `main` endpoint, which is the `maincore` core.
 
-The key (left side) is the name that will be exposed to the application in the `PLATFORM_RELATIONSHIPS` [variable](/development/variables.md).  The right hand side is the name of the service we specified above (`solrsearch`) and the endpoint (`main`).  If you named the service something different above, change `solrsearch` to that. The same rule is valid for the value of the endpoint named `main`.
+The key (left side) is the name that will be exposed to the application in the `PLATFORM_RELATIONSHIPS` [variable](/development/variables.md).  The right hand side is the name of the service specified above (`search`) and the endpoint (`main`).  If you named the service or endpoint something different above, change those values accordingly.
```
chadwcarlson:
```suggestion
The key (left side) is the name that will be exposed to the application in the `PLATFORM_RELATIONSHIPS` [variable](/development/variables.md).  The right hand side is the name of the service specified above (`search`) and the endpoint (`main`).  If you named the service or endpoint something than different above, change those values accordingly.
```

src/frameworks/drupal8/solr.md
```diff
@@ -1,71 +1,62 @@
...
 
-## Configuration
+### 4. Add auto-configuration code to `settings.platformsh.php`
 
-The configuration can be managed from `settings.platformsh.php` by adding the following code snippet.
+The configuration can be managed from `settings.platformsh.php` by adding the following code snippet.  It will override the environment-specific parts of the configuration object with the correct values to connect to the Platform.sh Solr instance.
 
 > **note**
 >
```
chadwcarlson: Can't make a suggestion here from this view of the diff, but maybe it would be a good idea to include a link here https://github.com/platformsh/config-reader-php for someone who's gotten this far but doesn't know about these yet

Crell: I think the composer install command is sufficient.  This is largely copy-pasta instructions, so it's just enough to support the pasta.

PR #1214, opened 2019-09-18
Adding cloudns to ALIAS list

matthiaz:
Adding cloudns, they support ALIAS as well.


Crell:
Good to know, thanks!

PR #1215, opened 2019-09-18
add MySQL and MariaDB 10.3 and 10.4 to supported versions

ludoge:


src/configuration/services/mysql.md
```diff
@@ -11,11 +11,15 @@ The service types `mariadb` and `mysql` both refer to MariaDB for compatibility
...
 * mariadb:10.2
+* mariadb:10.3
+* mariadb:10.4
 
 
 * mysql:10.0
 * mysql:10.1
 * mysql:10.2
+* mysql:10.3
+* mysql:10.4
```
Crell: Let's not include the `mysql` aliases for new ones.  People shouldn't be using them.  We want to transition to using just the accurate labels.

chadwcarlson:
We will need to add 10.3 & 10.4 to [this file](https://github.com/platformsh/platformsh-docs/blob/0ead91170f09acbb398fd71bc76695321be4aeee/src/registry/images/registry.json#L222) as well so that our configuration files reflect the most recent versions.

chadwcarlson:
Cleaned up in https://github.com/platformsh/platformsh-docs/pull/1225

PR #1216, opened 2019-09-18
Add SCA details.

chadwcarlson:
See DEVREL-314 for more information.


Crell:
I'm fine here, but I'd like a :+1: from @rinchen before we mark it approved.

PR #1218, opened 2019-09-19
Use relationship/service naming conventions for snippets.

chadwcarlson:
**Don't merge until Registry-Parser MR !6 is**

DEVREL-321

* All service/relationship names are updated to match the new convention in the registry, and in any other manually written snippets.
* Because the same "database/db" convention is used for mysql, mariadb, and oracle-mysql types, that page is cleaned up a bit so that it's not overly redundant.

## TODO

* Language-examples use relationship names that should be updated to the convention
    * This brings up a question. Language examples is a live site that uses most of the services, including multiple databases that the convention would say should all be the primary "database/db". That won't work on LE, since multiple services can't have the same relationship name. But if registry uses "database/db" for each of these potential primary databases, the code snippets and service usage examples won't match. This is also an issue if console team wants to eventually use snippet files to append/generate  a full relationships block/services.yaml.
* Persistent Redis: The relationship/service name for `redis-persistent` is defined in registry-parser, so there will need to be a separate PR there in order to update the snippet here https://pr-1218-fi5ym5q-652soceglkw4u.eu-3.platformsh.site/configuration/services/redis.html#persistent-redis. I'm thinking "data" for the service name.
    * Registry Parser MR # 6
    * Need to update `pshregistry-parser` to 1.0.4

### DIFF
--- a/src/configuration/services/influxdb.md
+++ b/src/configuration/services/influxdb.md
@@ -35,8 +35,8 @@ if (getenv('PLATFORM_RELATIONSHIPS')) {
 	$relationships = json_decode(base64_decode($relationships), TRUE);
 
 	// For a relationship named 'timedb' referring to one endpoint.
-	if (!empty($relationships['timedb'])) {
-		foreach ($relationships['timedb'] as $endpoint) {
+	if (!empty($relationships['database'])) {
+		foreach ($relationships['database'] as $endpoint) {
 			$settings['influxdb_host'] = $endpoint['host'];
 			$settings['influxdb_port'] = $endpoint['port'];
 			break;
--- a/src/registry/images/registry.json
+++ b/src/registry/images/registry.json
@@ -3,8 +3,8 @@
     "description": "",
     "disk": false,
     "docs": {
-      "relationship_name": "headless",
-      "service_name": "headless",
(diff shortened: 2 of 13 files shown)


src/configuration/services/influxdb.md
```diff
@@ -35,8 +35,8 @@ if (getenv('PLATFORM_RELATIONSHIPS')) {
 	$relationships = json_decode(base64_decode($relationships), TRUE);
 
 	// For a relationship named 'timedb' referring to one endpoint.
-	if (!empty($relationships['timedb'])) {
-		foreach ($relationships['timedb'] as $endpoint) {
+	if (!empty($relationships['database'])) {
+		foreach ($relationships['database'] as $endpoint) {
```
Crell: I can't imagine InfluxDB being the primary database on a project.  This seems like a bad use of the naming convention.

src/registry/images/registry.json
```diff
@@ -107,8 +107,8 @@
     "description": "",
     "disk": true,
     "docs": {
-      "relationship_name": "timedb",
-      "service_name": "influx",
+      "relationship_name": "database",
+      "service_name": "db",
```
Crell: Ibid.

PR #1220, opened 2019-09-23
pricing: add info on developer account sharing

sunishigh:


sunishigh:
I believe mentioning that in the docs to be worthy. Let me know if that'd be okay and I can reword/rephrase. From the point of view of a customer, I'd prefer to find that out in an explicit way.

Crell:
We don't want to document things that users should not be doing.  That makes it seem like a tacit approval, which is the exact opposite of the case.

PR #1227, opened 2019-09-24
Avoid `nvm: not found` on PE

Plopix:
It sounds like this is required.

To avoid 
```    Deploying applications
      Starting deployment
      Preparing slugs
      Entering maintenance mode
      Running deploy hook
      W: /etc/platform/7q34uh34hl5oa_stg/auto_deploy.sh: 2: /etc/platform/7q34uh34hl5oa_stg/auto_deploy.sh: nvm: not found
```

I don't have the issue on PS though. (project id: `7q34uh34hl5oa`

### DIFF
--- a/src/languages/nodejs/nvm.md
+++ b/src/languages/nodejs/nvm.md
@@ -18,6 +18,8 @@ hooks:
         nvm install 9.5.0
     deploy: |
         unset NPM_CONFIG_PREFIX
+        export NVM_DIR="$PLATFORM_APP_DIR/.nvm"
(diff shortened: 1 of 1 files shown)


Crell:
Hm.  That... shouldn't be necessary.

Can you open a support ticket first to figure out what's going on?  Let's confirm if it's really the problem or if there's something else at work in your case.

Crell:
This doesn't seem to be the case, so closing.

Plopix:
https://support.platform.sh/hc/en-us/requests/89697?page=1

@Crell this is needed.

PR #1230, opened 2019-09-27
Update registry to handle between service collisions.

chadwcarlson:
Updates new documentation convention to handle collisions between services when snippet files are joined together in generated services/app.yamls.

### DIFF
--- a/src/registry/images/registry.json
+++ b/src/registry/images/registry.json
@@ -256,7 +256,7 @@
     "repo_name": "mongodb",
     "disk": true,
     "docs": {
-      "relationship_name": "database",
+      "relationship_name": "mongodatabase",
       "service_name": "dbmongo",
       "url": "\/configuration\/services\/mongodb.html"
     },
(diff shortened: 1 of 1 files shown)


chadwcarlson:
oracle-mysql relationship: 

> oracle should be mysqldatabase. Oracle is its own database that is very different. Just to keep life interesting.

src/registry/images/registry.json
```diff
@@ -336,7 +336,7 @@
     "disk": true,
     "docs": {
       "relationship_name": "database",
```
chadwcarlson:
```suggestion
      "relationship_name": "mysqldatabase",
```

src/registry/images/registry.json
```diff
@@ -388,7 +388,7 @@
...
```
chadwcarlson:
```suggestion
      "relationship_name": "postgresdatabase",
```

src/registry/images/registry.json
```diff
@@ -257,7 +257,7 @@
...
```
chadwcarlson:
```suggestion
      "relationship_name": "mongodatabase",
```

PR #1232, opened 2019-09-30
Make-etc-hosts-easier

matthiaz:
There is an add-on on chrome and firefox that allows you to do the same thing, but without having to fiddle with /etc/hosts file.
Makes it a lot easier when you set up websites often.

### DIFF
--- a/src/golive/steps.md
+++ b/src/golive/steps.md
@@ -55,6 +55,12 @@ In OS X and Linux you can add that IP  to your `/etc/hosts` file.  In Windows th
 
 ![Hosts File](/images/config-files/hosts-file.png)
 
+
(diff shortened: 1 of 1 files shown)


src/golive/steps.md
```diff
@@ -55,6 +55,12 @@ In OS X and Linux you can add that IP  to your `/etc/hosts` file.  In Windows th
 
 ![Hosts File](/images/config-files/hosts-file.png)
 
+
+Alternatively there is also an add-on for firefox and google chrome that allow you to dynamically switch DNS IP addresses. No need to fiddle with your hosts files.
```
Crell:
```suggestion
Alternatively there is also an add-on for Firefox and Google Chrome that allow you to dynamically switch DNS IP addresses without modifying your `hosts` file.
```

PR #1236, opened 2019-10-02
Document default PORT value.

diogomatsubara:
When an user is following the python getting started, the $PORT variable
is referenced in the documentation but nowhere in the text it says which
is the default. Fix that by being explicit what are the default in the
first occurance.

### DIFF
--- a/src/languages/python.md
+++ b/src/languages/python.md
@@ -41,7 +41,7 @@ In this example, we use Gunicorn to run our WSGI application.  Configure the `.p
        start: "gunicorn -b 0.0.0.0:$PORT project.wsgi:application"
    ```
 
-   This assumes the WSGI file is `project/wsgi.py` and the WSGI application object is named `application` in the WSGI file.
+   This assumes the WSGI file is `project/wsgi.py` and the WSGI application object is named `application` in the WSGI file. By default, `$PORT` is `8888`.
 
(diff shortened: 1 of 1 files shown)


Crell:
I'm not sure this is needed.  Currently, we tell people to just read from the port.  If they're doing so, why does it matter what the typical value is?  By design, you shouldn't care.

kotnik:
One needs to know the default port when running custom Python apps, for example.

Crell:
Custom Python apps can/should read from $PORT like everyone else...

kotnik:
Can't argue against that. Maybe that should be documented then?

Crell:
It is?  https://docs.platform.sh/configuration/app/web.html#socket-family

Or do you mean on the Python page specifically?

kotnik:
Right, I was referring Python page in particular.

chadwcarlson:
@Crell Documenting the default value somewhere is probably a good idea in my opinion. In general, yes, use `PORT`, but at least one of our templates needs to define it explicitly in the deploy hook https://github.com/platformsh-templates/microprofile-tomee/blob/66e1596674a8477bf9cfaf91d17692c8db545ec9/.platform.app.yaml#L16

Crell:
Closing in favor of #1272

PR #1250, opened 2019-10-22
Adding a note about memory usage to blackfire

xaviemirmon:


xaviemirmon:
Merging all changes into one PR

PR #1251, opened 2019-10-22
Adding a note about memory usage to tideways

xaviemirmon:


xaviemirmon:
Did all changes in one PR

PR #1252, opened 2019-10-22

xaviemirmon:


src/administration/integrations/blackfire.md
```diff
@@ -157,3 +157,10 @@ Once you are done, please disable logging with:
 
 * `platform variable:delete php:blackfire.log_file`
 * `platform variable:delete php:blackfire.log_level`
+
+
+## Notes
+
+### Memory usage
+
+This extension adds extra memory overhead to your application and each request will require more memory. So, depending on how aggressive your `sizing_hints` are, you may need to adjust them or upgrade your plan.
```
lolautruche: Actually overhead occurs only for profiling requests. If you don't trigger a profile, there is no overhead at all, which makes it safe for production.

xaviemirmon: Hi @lolautruche 

I've noticed a slight change when using Blackfire on my instance (~2-3mb).  Nowhere near the overhead from New Relic.  But was still enough to cause issues.   Is it possible that the PHP extension has to be loaded causing a bit of memory consumption?

romainneutron: What do you mean by overhead? CPU usage? Please detail this please

xaviemirmon: Hi @romainneutron 

I mean

> each request will require more memory

Any advice on how to clarify that would be greatly appreciated

Thanks!

Xavier

romainneutron: What are you talking about, which memory? The PHP memory allocated by the engine and that you can get using `memory_get_usage`, or the memory allocated to start a PHP process by FPM?

xaviemirmon: I'm talking about the PHP-fpm child processes memory as defined in the `sizing_hints` variable in the `.platform.app.yml` file.   

I noticed this because having Blackfire enabled seemed to lead to an increase of Signal 9 errors.  

I looked the usage using `tail -n5000 /var/log/php.access.log | awk '{print $6}' | sort -n | uniq -c`

tinefin: New Relic should not have a noticeable performance overhead, not even for the transaction traces. It is around 8 years ago that this could be the case. Unless a config variable for non-standard amount of tracing is enabled of course, or something like that.. 

Last I checked their documentation, it was very clear on which options could cause performance impact:
https://docs.newrelic.com/docs/agents/php-agent/php-agent-api

If it does look like New Relic has noticeable performance overhead, I would highly recommend lodging a ticket with New Relic support as then something is wrong.

Crell:
Given the discussion above I'm going to close this issue.  A good profiling extension has no more overhead than any other extension, of which there are many that could be enabled.

PR #1257, opened 2019-11-04
Update solr.md

michielkenis:
Something is definitely missing here (or I must be blind). 

Things we're missing is the active bug on the Search API Solr where Solarium is outdated (using the two steps on https://www.drupal.org/project/search_api_solr/issues/3085196#comment-13305381 fixes this), but requires adding the composer files again. 

Another thing required for our setup to work, is modify the `conf_dir` on `services.yaml`. After your step 6, we need to modify the `conf_dir` in order to get a working Index on this... 

I'm happy to discuss this here or on Slack, but I believe there is a missing step on this page.


Crell:
Closing in favor of #1258.  Thanks for the heads up!

PR #1264, opened 2019-11-06
Apply Dedicated Enterprise for contractual-like statements.

ralt:
I voluntarily did not go through all the docs; just the places that
were sensitive.


Crell:
I manually merged this into `dedicated-ent`, tracked in #1265.  Thanks!

PR #1270, opened 2019-11-19
Update Drupal settings.

michielkenis:


src/frameworks/drupal7/composer-manager.md
```diff
@@ -16,6 +16,8 @@ projects[composer_manager][patch][] = "https://www.drupal.org/files/issues/compo
 
 If you're checking the entire codebase into Git, do so with Composer Manager as well.  Then download and apply the two patches in the issues above and commit the result.
 
+Hereafter, it's important to uncheck the two "Automatically..." options at the config page on `admin/config/system/composer-manager/settings`. If checked, Drupal tries to update the composer folder. Since this isn't a writable mount, installation of composer based modules will fail due to these writing permissions.
```
Crell: Looks good, except "Hereafter" doesn't quite fit grammatically, even if it does sound cool. :smile: I think the sentence works fine without that word.
```suggestion
It's important to uncheck the two "Automatically..." options at the config page on `admin/config/system/composer-manager/settings`. If checked, Drupal tries to update the composer folder. Since this isn't a writable mount, installation of composer based modules will fail due to these writing permissions.
```

michielkenis: Valid suggestion :-)

PR #1278, opened 2019-11-29
Adding php 7.4 extensions column

matthiaz:
redis is no longer supported in php7.4 adding a note for it specifically as well.

### DIFF
--- a/src/languages/php/extensions.md
+++ b/src/languages/php/extensions.md
@@ -143,7 +143,7 @@ This is the complete list of extensions that can be enabled:
 | sysvmsg       |     |     |     | *   | *   | *   | *   | *   |
 | sysvsem       |     |     |     | *   | *   | *   | *   | *   |
 | sysvshm       |     |     |     | *   | *   | *   | *   | *   |
-| tideways|     |     |     | *   | *   | *   | *   | *   | *   |
+| tideways      |     |     | *   | *   | *   | *   | *   | *   |
 | tideways-xhprof  |     |     |     | *   | *   | *   | *   | *   |
 | tidy          | *   | *   | *   | *   | *   | *   | *   | *   |
 | tokenizer     |     |     |     | *   | *   | *   | *   | *   |
@@ -168,7 +168,7 @@ This is the complete list of extensions that can be enabled:
 > your environment. For PHP 7, use `ls /etc/php/*/mods-available`.
 
 > **note**
-> Starting with PHP 7.4, we're no longer including the redis PECL extension with the PHP application image. Unfortunately, that library has had some unfortunate API breaks recently that make it much harder for us to support reliably. Instead, we recommend users install the [predis](https://packagist.org/packages/predis/) package, which will work across all supported PHP versions.
+> Starting with PHP 7.4, we're no longer including the redis PECL extension with the PHP application image. Unfortunately, that library has had some unfortunate API breaks recently that make it much harder for us to support reliably. Instead, we recommend users install the [predis](https://packagist.org/packages/predis/predis) package, which will work across all supported PHP versions.
 
 ## Custom PHP extensions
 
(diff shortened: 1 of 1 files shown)


src/languages/php/extensions.md
```diff
@@ -57,116 +57,119 @@ You can disable those by adding them to the `disabled_extensions` list.
...
+| sodium        |     |     |     |     |     | *   | *   | *   |
+| sourceguardian|     |     |     | *   | *   |     |     |     |
+| spplus        | *   | *   |     |     |     |     |     |     |
+| sqlite3       | *   | *   | *   | *   | *   | *   | *   | *   |
+| sqlsrv        |     |     |     | *   | *   | *   |     |     |
+| ssh2          | *   | *   | *   | *   | *   | *   | *   | *   |
+| sysvmsg       |     |     |     | *   | *   | *   | *   | *   |
+| sysvsem       |     |     |     | *   | *   | *   | *   | *   |
+| sysvshm       |     |     |     | *   | *   | *   | *   | *   |
+| tideways|     |     |     | *   | *   | *   | *   | *   | *   |
```
pjcdawkins:
```suggestion
| tideways      |     |     | *   | *   | *   | *   | *   | *   |
```

src/languages/php/extensions.md
```diff
@@ -57,116 +57,119 @@ You can disable those by adding them to the `disabled_extensions` list.
...
+| zendopcache   | *   |     |     |     |     |     |     |     |
+| zip           |     |     |     | *   | *   | *   | *   | *   |
 
 > **note**
 > You can check out the output of `ls /etc/php5/mods-available` to
 > see the up-to-date complete list of extensions after you SSH into
 > your environment. For PHP 7, use `ls /etc/php/*/mods-available`.
 
+> **note**
+> Starting with PHP 7.4, we're no longer including the redis PECL extension with the PHP application image. Unfortunately, that library has had some unfortunate API breaks recently that make it much harder for us to support reliably. Instead, we recommend users install the [predis](https://packagist.org/packages/predis/) package, which will work across all supported PHP versions.
```
pjcdawkins:
```suggestion
> Starting with PHP 7.4, we're no longer including the redis PECL extension with the PHP application image. Unfortunately, that library has had some unfortunate API breaks recently that make it much harder for us to support reliably. Instead, we recommend users install the [predis](https://packagist.org/packages/predis/predis) package, which will work across all supported PHP versions.
```

fabpot:
Removing support for the redis extension is bad news. Even if the API breaks, I don't see why it cannot be included anymore. BC breaks should be taken care of by the developer using the extension, not psh. Or am I missing something?

You recommend using predis, which is unfortunately not maintained anymore. No commit since July 2017 and releases in the last 3 years. See https://github.com/nrk/predis/issues/587 for more info.

jww-sh:
> Removing support for the redis extension is bad news. Even if the API breaks, I don't see why it cannot be included anymore. BC breaks should be taken care of by the developer using the extension, not psh. Or am I missing something?
> 
> You recommend using predis, which is unfortunately not maintained anymore. No commit since July 2017 and releases in the last 3 years. See [nrk/predis#587](https://github.com/nrk/predis/issues/587) for more info.

you can still add it as a custom extension, the issue is that there are multiple versions of PHP-redis, 3, 4 and 5

fabpot:
@jww-sh Ok, understood. So, instead of recommending installing predis, what about documenting how to install one the versions of the Redis extension? Would it be possible?

jww-sh:
> @jww-sh Ok, understood. So, instead of recommending installing predis, what about documenting how to install one the versions of the Redis extension? Would it be possible?

We could also point users to 

https://docs.platform.sh/languages/php/extensions.html#custom-php-extensions

Crell:
We're discussing some possible workarounds right now; so far nothing to announce but we're discussing as I write this. :smile:

Crell:
Closing this thread, as the changes in it have already been merged elsewhere.

PR #1282, opened 2019-12-05
Document Route Attributes

OriPekelman:


src/configuration/routes.md
```diff
@@ -196,6 +196,38 @@ function get_domain_for(string $id) {
...
+        [original_url] => https://site1.{default}/
+        [attributes] => Array
+            (
+                [foo] => bar
+            )
+        // ...
+    )
+```
+
+These extra attributes may be used to "tag" routes in more comples scenarios that can then be read by your application.
```
ralt: complex*

chadwcarlson:
```suggestion
These extra attributes may be used to "tag" routes in more complex scenarios that can then be read by your application.
```

src/configuration/routes.md
```diff
@@ -196,6 +196,38 @@ function get_domain_for(string $id) {
...
+
+```
+"http://{default}/":
+    type: upstream
+    upstream: "app:http"
+    attributes:
+        "foo": "bar"
+```
+
+Will appear in the routes data in PHP like so:
```
chadwcarlson:
```suggestion
Attributes will appear in the routes data for PHP like so:
```

src/configuration/routes.md
```diff
@@ -196,6 +196,38 @@ function get_domain_for(string $id) {
...
+
+```text
```
ralt: Maybe you should add a snippet showing something like:

```
function route_attributes() {
    foreach ($_ENV["PLATFORM_ROUTES"] as $route_key => $route) {
        if ($route["original_route"] === $_REQUEST["HTTP_X_ORIGINAL_ROUTE"]) {
            return $route["attributes"];
        }
    }
}

$current_attributes = route_attributes();
```

??

Crell: It's actually a bit more complex than that; that logic may get you a redirect.  We have "how to mess with routes" documented elsewhere, mainly in config reader.

ralt: Then probably mention it? After reading this I'm interested in figuring out how I can mess with routes to get the current one, so a link would be handy.

PR #1284, opened 2019-12-06
Use the same service name as in routes.yaml below

achekulaev:
In the varnish config example service is called "proxy", while in `routes.yaml` example down on the same page it is addressed as "varnish:http".
Just copying and pasting code chunks will not work because of this. 

This change makes service name the same in both examples.


Crell:
Thanks!

PR #1286, opened 2019-12-11
`health.webhook` [no longer] seems to be a valid type

christianyates:
Following the original instructions, e.g. 

```bash
platform integration:add --type health.webhook --url=A-URL-THAT-CAN-RECEIVE-THE-POSTED-JSON
```

Generates the following error:
```
[InvalidValueException]
  health.webhook is not one of: bitbucket, bitbucket_server, github, gitlab, hipchat, webhook, health.email, health.pagerduty, health.slack
```

Using the parameter `webhook` seems to work as expected.

### DIFF
--- a/src/administration/integrations/notifications.md
+++ b/src/administration/integrations/notifications.md
@@ -66,15 +66,15 @@ A notification can trigger a message to be sent to a web endpoint.
 To do so, register a `health.webhook` integration as follows:
 
 ```bash
-platform integration:add --type health.webhook --url=A-URL-THAT-CAN-RECEIVE-THE-POSTED-JSON
+platform integration:add --type webhook --url=A-URL-THAT-CAN-RECEIVE-THE-POSTED-JSON
 ```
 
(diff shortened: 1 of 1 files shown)


src/administration/integrations/notifications.md
```diff
@@ -66,15 +66,15 @@ A notification can trigger a message to be sent to a web endpoint.
 To do so, register a `health.webhook` integration as follows:
```
christianyates: Missed one:
```suggestion
To do so, register a `webhook` integration as follows:
```

Crell:
Hm.  I'm not sure if the generic `webhook` (cf https://docs.platform.sh/administration/integrations/webhooks.html) is actually a substitute.  That reads all activities, not just health notifications.  When did `health.webhook` stop working and why?

pjcdawkins:
health.webhook is not yet in the CLI, that's the confusion that let to this PR I think

pjcdawkins:
.. it is in the CLI now.

PR #1288, opened 2019-12-12
Update relationship name in solr.md

georgezee:
I'm suggesting the replacement of `solrsearch` with `solr` for the relationship name in your example, so that it is compatible with the Enterprise environments that use the latter.


Crell:
Hm.  We actually have a consistent standard on the naming throughout the docs now, so I don't want to break that.  But you're right, Dedicated is probably lagging on this point.  I'll have to loop the Ops team in.

georgezee:
That's fine - I just wanted to point out the inconsistency as it tripped us up in our recent upgrade.

PR #1292, opened 2019-12-19
Database configuration for Lando

GuGuss:
Just importing the Platform.sh database in Lando when using a Platform.sh Drupal template  won't work as is.

Also, we could add a note on how to import files in Lando as well.

### DIFF
--- a/src/development/local/lando.md
+++ b/src/development/local/lando.md
@@ -40,3 +40,22 @@ lando db-import database.sql.gz
 ```
 
 Rsync can download user files easily and efficiently.  See the [exporting tutorial](/tutorials/exporting.md) for information on how to use rsync.
+
+Then you need to update your `sites/default/settings.local.php` to configure your codebase to connect to the local database that you just imported:
+
+```php
+/* Working in local with Lando */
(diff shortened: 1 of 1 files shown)


src/development/local/lando.md
```diff
@@ -40,3 +40,22 @@ lando db-import database.sql.gz
...
 Rsync can download user files easily and efficiently.  See the [exporting tutorial](/tutorials/exporting.md) for information on how to use rsync.
+
+Then you need to update your `sites/default/settings.local.php` to configure your codebase to connect to the local database that you just imported:
+
+```php
+/* Working in local with Lando */
+if (getenv('LANDO') === 'ON') {
+  $lando_info = json_decode(getenv('LANDO_INFO'), TRUE);
+  $settings['trusted_host_patterns'] = ['.*'];
+  $settings['hash_salt'] = md5(getenv('LANDO_HOST_IP'));
```
Crell: I know it's just generating a random string here, but this seems like a poor way to do so.  Some security linters will object if you even use md5() at all.  Let's just put a "change me" string in here.

PR #1299, opened 2020-01-17
Fix gitlab integration argument

ralt:
not sure where `--server-project` came from


matthiaz:
Discussion here to figure out where it came from https://platformsh.slack.com/archives/CF8T8GHFE/p1579249252000700

PR #1303, opened 2020-01-27
Add virtual host config.

chadwcarlson:
* Adds virtual host configuration for Rabbit MQ.
* Cleans up headings on that page.


Crell:
Does this addition have any implication for the relationships array, like it does for MySQL, or is it all on the client?

src/configuration/services/rabbitmq.md
```diff
@@ -83,3 +83,19 @@ dependencies:
 ```
 
 Then, when you SSH into your container, you can simply type any `amqp-` command available to manage your queues.
+
+## Configuration
+
+### Virtual hosts
+
+You can configure additional [virtual hosts](https://www.rabbitmq.com/vhosts.html) to a RabbitMQ service, which can be useful for separating resources. In your `.platform/services.yaml` file define the names of the virtual hosts under the `configuration.vhosts` attribute:
```
JGrubb: Can we flesh this out a tiny bit more?  "Useful for separating resources" but why would you want to do that?

ralt: It's like, yes, you can use prefix on your mysql tables to use a single database, but the correct abstraction is using several databases. Same idea for rabbitmq vhosts. Not sure we need to expand on this.

ludoge:
This feature has been deployed to all live regions.

ralt:
@Crell no implications there

PR #1314, opened 2020-02-13
Adding information about how to add a hidden folder

matthiaz:


src/configuration/app/storage.md
```diff
@@ -114,6 +114,17 @@ web:
 
 See the [web locations](/configuration/app/web.md) documentation for more details.
 
+## Unable to mount hidden folder?
+
+We ignore every YAML key that starts with a dot. This causes a mount like '.myhiddenfolder' to be ignored. If you want to mount a hidden folder, you'll have to prepend it with a `/`
```
chadwcarlson:
```suggestion
Platform.sh ignores YAML keys that start with a dot. This causes a mount like `.myhiddenfolder` to be ignored. If you want to mount a hidden folder, you'll have to prepend it with a `/`:
```

src/configuration/app/storage.md
```diff
@@ -114,6 +114,17 @@ web:
...
+
+```yaml
+mounts:
+  '/.myhiddenfolder':
+    source: local
+    source_path: '.myhiddenfolder'
```
chadwcarlson:
```suggestion
    source_path: 'myhiddenfolder'
```

chadwcarlson: This one's your call. In our templates we don't keep mount source paths hidden to keep it simple and visible, so this suggestion is more of keeping it consistent with that choice.

Crell: Dot files are annoying, so let's not use them in source paths.

src/configuration/app/storage.md
```diff
@@ -114,6 +114,17 @@ web:
 
 See the [web locations](/configuration/app/web.md) documentation for more details.
 
+## Unable to mount hidden folder?
```
Crell:
```suggestion
## Why can't I mount a hidden folder?
```

PR #1320, opened 2020-02-20
Fix missing ssh forwarding instructions

raupie:
https://docs.platform.sh/development/transfer-dedicated.html

ref: `First, set up <link>SSH forwarding</link> by default for Platform.sh domains.`
404: https://docs.platform.sh/dedicated/support/ssh-agent.md#forwarding-keys-by-default

Link is broken, pointing to old dedicated enterprise documentation.

Adding the `Forwarding keys by default` instructions to the `ssh.md` documentation and updating the link on `transfer-dedicated.md`.

### DIFF
--- a/src/development/ssh.md
+++ b/src/development/ssh.md
@@ -71,6 +71,20 @@ You have your SSH keys (if not, take a look at the section above), but you need 
 
 That's it! You're all set. Now you'll be able to use Git and command shells with any Platform.sh environment that your user account is authorized to work with.
 
+### Forwarding keys by default
+
+It may be helpful to set your SSH client to always forward keys to Platform.sh servers, which can simplify other SSH or Rsync commands.  To do so, include a block in your local `~/.ssh/config` file like so:
+
+```
+Host *.us.platform.sh
+       ForwardAgent yes
+
+Host *.eu.platform.sh
+       ForwardAgent yes
+```
+
+(You can include other configuration as desired.)
+
 ![Setting Up Your Project Add SSH Key Done](/images/management-console/account-ssh-keys.png)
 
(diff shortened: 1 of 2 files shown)


src/development/ssh.md
```diff
@@ -71,6 +71,20 @@ You have your SSH keys (if not, take a look at the section above), but you need
 
 That's it! You're all set. Now you'll be able to use Git and command shells with any Platform.sh environment that your user account is authorized to work with.
 
+### Forwarding keys by default
```
Crell: Is this still necessary with the integrated Dev (wings)?  I thought it wasn't anymore.  (We only document the integrated approach, not the old legacy one.)

raupie: Not sure I understand. Integrated approach? Using the CLI?

Crell: `platform ssh -e production` gets you to the prod environment these days on Dedicated.  So, when would someone still need to setup key forwarding?

raupie: I was assuming for rsync/scp

src/development/ssh.md
```diff
@@ -71,6 +71,20 @@ You have your SSH keys (if not, take a look at the section above), but you need
...
+
+```
+Host *.us.platform.sh
+       ForwardAgent yes
+
+Host *.eu.platform.sh
+       ForwardAgent yes
+```
+
+(You can include other configuration as desired.)
```
Crell:
```suggestion
Include one `Host` entry for each Platform.sh region you want to connect to, such as `us-2` or `eu-4`.  (You can include other configuration as desired.)
```

PR #1341, opened 2020-03-23
Revise sizing page to account for the new beta memory controls.

Crell:


src/configuration/app/size.md
```diff
@@ -1,27 +1,93 @@
...
+
+Application containers have memory allocated to them based on their container size as well, but the ratio is configurable.  Specifically, there are two properties that can be adjusted.  Their default values are shown below.
+
+```php
+resources:
+    base_memory: 64
+    memory_ratio: 256
 ```
+
+The `base_memory` is a fixed size (in megabytes) that the container will have, period.  The `memory_ratio` is a multiplier based on the CPU size.  That is, an `S` CPU size is 0.5 CPU slides, with the above configuration it would get 64 + 0.5 * 256 = 192 MB of RAM.  If the same application had its size increased to an XL, it would get 64 + 4 * 256 = 1088 MB of RAM.
```
chadwcarlson: These `*` characters don't render for some reason. Maybe switch to parenthesis or `x`?

Crell: Ugh.  It's probably trying to render them as formatting codes.  I hate Markdown.

otaviojava (dismissed):

chadwcarlson (approved):

CollierCZ:
This functionality is being redone and will require new docs when it's ready.

PR #1342, opened 2020-03-25
Added redirectssl to www redirection service

dubcanada:
Hello,

I am the owner/developer of the service, we have a guide on how to setup on https://redirectssl.com/guides/platformsh let me know if this is not something you guys allow.

If it is I would love to get redirectssl added to the list.

Thanks,


Crell:
That's great, thanks!

PR #1344, opened 2020-03-26
updated drupal-solr doc to reflect recent developments

yuriybabenko:
There are two changes here:

1. Changed the wording from `older than 8.9.0` to `older than 9.0`. This was done because the [issue in question](https://www.drupal.org/project/drupal/issues/2876675) was recently closed, and the changes did not make their way into 8.9, but **are** present in 9.0.

2. Changed event-dispatcher's version from `4.3.4 as 3.4.99` to `4.3.4 as 3.4.35`. With `99` in place, composer errors out, complaining about not being able to meet minimum stability level (same behaviour with `minimum-stability` set to dev, alpha, beta, RC).

### DIFF
--- a/src/frameworks/drupal8/solr.md
+++ b/src/frameworks/drupal8/solr.md
@@ -12,12 +12,12 @@ Advanced Solr service configuration and implementation in other frameworks other
 
 ### 0. Upgrade Symfony Event Dispatcher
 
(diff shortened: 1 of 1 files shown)


Crell:
Thanks for the heads up!

PR #1348, opened 2020-04-02
Document one way of doing a custom 404 page.

Crell:
I have tested this.  It actually works. :smile:


tucksaun:
But the status code becomes 200 if you use a static file.
One way I have used in the past is using a simple PHP script just to change to be able to call `http_response_code`

PR #1354, opened 2020-04-06
Clarify that Solr config is effectively required these days.

Crell:


src/configuration/services/solr.md
```diff
@@ -178,7 +176,7 @@ search:
                 core: collection1
 ```
 
-The Solr 6.x Drupal 8 configuration files are reasonably generic and should work in many other circumstances, but explicitly defining a core, configuration, and endpoint is generally recommended.
+The default configuration is based on an older version of the Drupal 8 Search API Solr module that is no longer in use.  While it may wor for generic cases defining your own custom configuration, core, and endpoint is strongly recommended.
```
chadwcarlson, line 179:
```suggestion
The default configuration is based on an older version of the Drupal 8 Search API Solr module that is no longer in use.  While it may work for generic cases defining your own custom configuration, core, and endpoint is strongly recommended.
```

PR #1355, opened 2020-04-07
Missing relationship configuration credentials

xtfer:
Adds missing ElasticSearch credentials

### DIFF
--- a/src/frameworks/drupal8/elasticsearch.md
+++ b/src/frameworks/drupal8/elasticsearch.md
@@ -57,9 +57,12 @@ Then, paste the following code snippet into your `settings.platformsh.php` file.
 ```php
 // Update these values to the relationship name (from .platform.app.yaml)
 // and the machine name of the server from your Drupal configuration.
-$relationship_name = 'elasticsearch';
+$relationship_name = 'RELATIONSHIP_NAME';
 $es_cluster_name = 'YOUR_CLUSTER_HERE';
 if ($platformsh->hasRelationship($relationship_name)) {
+
+  $creds = $platformsh->credentials($relationship_name)
+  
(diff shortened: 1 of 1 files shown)


src/frameworks/drupal8/elasticsearch.md
```diff
@@ -57,9 +57,12 @@ Then, paste the following code snippet into your `settings.platformsh.php` file.
 ```php
 // Update these values to the relationship name (from .platform.app.yaml)
 // and the machine name of the server from your Drupal configuration.
-$relationship_name = 'elasticsearch';
+$relationship_name = 'RELATIONSHIP_NAME';
 $es_cluster_name = 'YOUR_CLUSTER_HERE';
 if ($platformsh->hasRelationship($relationship_name)) {
+
+  $creds = $platformsh->credentials($relationship_name)
```
Crell, line 64: This isn't necessary; the `formattedCredentials()` method call below will call `credentials()` internally.

src/frameworks/drupal8/elasticsearch.md
```diff
@@ -57,9 +57,12 @@ Then, paste the following code snippet into your `settings.platformsh.php` file.
 ```php
 // Update these values to the relationship name (from .platform.app.yaml)
 // and the machine name of the server from your Drupal configuration.
-$relationship_name = 'elasticsearch';
+$relationship_name = 'RELATIONSHIP_NAME';
```
Crell, line 60: I'd prefer to leave this at the value that is most commonly used, so it doesn't need to be modified.

xtfer: Fair enough, however it doesn't match https://docs.platform.sh/configuration/services/elasticsearch.html#usage-example.

PR #1356, opened 2020-04-08
Add Xdebug instructions.

Crell:


src/languages/php/xdebug.md
```diff
@@ -0,0 +1,82 @@
...
+## Setting up Xdebug
+
+Xdebug is not enabled the same way as other extensions, as it should not be active on most requests.  Xdebug has a substantial impact on performance and should not be run in production.
+
+Instead, Xdebug can be enabled by adding the following configuration to the application's `.platform.app.yaml` file:
+
+```yaml
+runtime:
+    xdebug:
+        idekey: PHPSTORM
```
pjcdawkins, line 16: This should probably be a secret, more complex key, for production?

Crell: In practice, I'm not sure it means anything.  Security is handled through SSH, not the idekey.  The IDE key is not a password-like field.

PR #1359, opened 2020-04-10
Source Operations with external Git integration

GuGuss:
- [ ] Document that the `enable_codesource_integration_push` setting needs to be enabled for this to work.

### DIFF
--- a/src/configuration/app/source-operations.md
+++ b/src/configuration/app/source-operations.md
@@ -56,9 +56,7 @@ Git integration can be configured to send commits made to the Platform.sh Git re
 
 > **note**
 >
-> Source Operations will only be triggered on environment created by a branch, and not to environment created by a Pull Request on the external upstream (GitHub, Bitbucket, Gitlab).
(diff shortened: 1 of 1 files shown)


src/configuration/app/source-operations.md
```diff
@@ -50,6 +50,14 @@ When this operation is triggered:
 
 Note that these operations run in an isolated container: it is not part of the runtime cluster of the environment, and doesn't require the environment to be running.  Also be aware that if multiple applications in a single project both result in a new commit, that will appear as two distinct commits in the Git history but only a single new build/deploy cycle will occur.
 
+## Source Operations with an external Git integration
+
+Git integration can be configured to send commits made to the Platform.sh Git remote, to the upstream repository instead. This means that if a source operation did generate a new commit, the commit will be pushed to the upstream repository.
```
Crell, line 55: Can be configured? I thought this was automatic if an integration was active?

GuGuss: @Crell Currently, the feature is behind a feature flag called: `enable_codesource_integration_push` which default to `false`. I'll make this more explicit.

PR #1364, opened 2020-04-14
fix nvm not found in the deploy hook

dpraburaj:
When following the steps in https://docs.platform.sh/languages/nodejs/nvm.html,
`nvm` invocations inside the deploy hook appear to fail. I have tested this on `php:7.3` and `ruby:2.3`. 

The warning/error that occurs is:

```
  Executing deploy hook for application app
    W: /bin/dash: 3: nvm: not found
```

Adding

 ```
        [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
```

to the deploy hook appears to resolve the issue. Please let me know if you require further information or if there is an alternative to adding the above lines to the deploy hook.


Crell:
This is apparently just an issue for Dedicated.  cf: #1227

Plopix:
For future reference: 

`export NVM_DIR="$HOME/.nvm"` -> `export NVM_DIR="$PLATFORM_APP_DIR/.nvm"`

I had to replace `$HOME` to `$PLATFORM_APP_DIR` indeed it depends if the build is reused or not etc.. I don't know the internal mechanism but I guessed and it worked ;-)

PR #1366, opened 2020-04-14
Hugo migration.

chadwcarlson:
- ~add search~
- ~style search~ https://github.com/platformsh/platformsh-docs/pull/1381
- ~make multi-app~ https://github.com/platformsh/platformsh-docs/pull/1381
- ~template listing design feedback~
- ~link checks~
- ~main body content  margins~
- ~book links on mobile~
- ~Replace editpage button~


Crell:
The search box seems misaligned:

image

Crell:
I get a pair of 403 errors in the network tab when I try to use the search box.

.platform.app.yaml
```diff
@@ -1,80 +1,51 @@
...
+disk: 5120
 
-# Configure the web server to serve our static site.
+# The configuration of the application when it is exposed to the web.
 web:
     commands:
-        # Run a no-op process that uses no CPU resources, since this is a static site.
-        start: sleep infinity
+      # Run the Meilisearch server
+      start: "./meilisearch/meilisearch --http-addr localhost:${PORT}"
```
Crell, line 27: If the goal is for this to be the central search repository for multiple sites, I agree with Ori that it should be its own project if at all possible.  I don't know how much complexity that adds, though.  At the very least it should be its own app container.  (We aren't paying for it, so an extra app container is cheap. :smile: )

chadwcarlson: I'm testing that now, totally agree

PR #1373, opened 2020-04-16
NewRelic instructions are PHP-specific

dman-coders:
As revealed in ZD86807, Trying to follow the PHP instructions to profile a Java application leads to sadness.
The doc page didn't explicitly acknowledge that, and there are no explicit clues that
```
runtime:
    extensions:
        - newrelic
```
is a php-specific thing.

### DIFF
--- a/src/administration/integrations/new-relic.md
+++ b/src/administration/integrations/new-relic.md
@@ -1,6 +1,6 @@
 # New Relic
 
-Platform.sh supports [New Relic APM](https://newrelic.com/products/application-monitoring).
+Platform.sh supports [New Relic APM](https://newrelic.com/products/application-monitoring) for profiling PHP applications. These instructions do not apply to other languages.
 
 ## On a Grid plan
 


Crell:
Do we not support NR on non-PHP at all, or is it just that snippet that is PHP-specific?  If the latter then we should just mark that part.

dman-coders:
It's not supported for anything but PHP.
That is: there is no support or docs available for anything else. Only-PHP is silently assumed

Apparently it's not prohibited, to try it with java, but we are on our own to figure that out.
ALL instructions on this page are only for PHP.
If we had other instructions, there would be no cross-over, so it would be added as a new page or new section.
I'm trying to figure out what that instruction would look like.

otaviojava:
@Crell 
Each Language has your own way to configure New Relic. 
I wrote to Java: 
https://community.platform.sh/t/how-to-configure-java-with-new-relic-at-platform-sh/533

Crell:
Hrm.  In that case, let's rename the current page to "New Relic (PHP)", and add a new page "New Relic (Java)" that is Otavio's writeup from community.  It belongs in docs, not community.

otaviojava:
@Crell  I like the idea of keep this integration page.
And create more two pages: One to PHP and another with Java.

PR #1380, opened 2020-04-23
Document YAML anchors and aliases.

Crell:


src/configuration/yaml.md
```diff
@@ -156,3 +156,87 @@ mysearch:
...
+```
+
+Which is equivalent to:
+
+```yaml
+foo:
+    thing: stuff
+    many: {'stuff', 'here'}
+bar:
+    thing: other
```
chadwcarlson, line 211: Is this syntax meant to override `thing`? i.e now  is `thing: stuff`?

PR #1390, opened 2020-05-01
Update Bitbucket OAuth instructions for Workspaces

pjcdawkins:
OAuth consumers now live under the Workspace, not the user account

Also fixes a typo, and subtly adds a space at the beginning of the suggested command, which (if copied and pasted) helps to keep the secret out of shell history


docs/src/administration/integrations/bitbucket.md
```diff
@@ -7,16 +7,14 @@ description: |
 
 {{< description >}}
 
-## Setting up an OAuth consumer
+### Set up an OAuth consumer
```
chadwcarlson, line 10:
```suggestion
## Set up an OAuth consumer
```

PR #1391, opened 2020-05-01
Search updates

chadwcarlson:
- Fixes how docs page body text is parsed a bit, helping relevancy
- Adds subsections as a higher relevancy attribute
- Adds synonyms

### DIFF
--- a/docs/static/scripts/xss/src/containers/Search.js
+++ b/docs/static/scripts/xss/src/containers/Search.js
@@ -35,7 +35,7 @@ class Search extends Component {
       .catch(() => this.setState({ error: true }))
   }
 
-  clearInput = () => {
+  clearInputFunc = () => {
     this.setState({
       search: ''
     })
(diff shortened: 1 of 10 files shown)


docs/static/scripts/xss/src/containers/Search.js
```diff
@@ -35,7 +35,7 @@ class Search extends Component {
       .catch(() => this.setState({ error: true }))
   }
 
-  clearInput = () => {
+  clearInputFunc = () => {
```
Crell, line 38: Pourquoi?

chadwcarlson: I was getting an error that was ambiguous and hard to track down at first. Part of it was making me think that because the first input tag's id was also `clearinput` that React was being weird and confusing these two. It obviously def wasn't that, but I just didn't change it back.

PR #1394, opened 2020-05-01
Update contributing doc for hugo specifics.

chadwcarlson:


CONTRIBUTING.md
```diff
@@ -4,9 +4,9 @@ Thank you for helping make our documentation better!
...
 In order to maintain a consistent style and voice throughout our documentation please try to follow these standards and conventions when filing Pull Requests against our documentation.
 
-## Markdown
+## Hugo & Markdown
 
-GitBook (which we use for documentation) uses Markdown as its file format.  
+Hugo (which we use for documentation) uses Markdown as its file format.  
 
 * All files should end in `.md`. They should be all-lower-case and match the title of the page, or in some cases an abbridged version of it.
 * All pages should start with a title, denoted with # (first level header).
```
Crell, line 12: This should be updated for the YAML frontmatter.

CONTRIBUTING.md
```diff
@@ -16,15 +16,186 @@ GitBook (which we use for documentation) uses Markdown as its file format.
 * Always use inline links.
 * Do not hard-wrap prose text.  Set your text editor to soft wrapping.
 * Internal links should be absolute (starting with `/`) and link to a file ending in `.md`, not `.html`. That makes them easier to find when needed to update.
+* Each page must contain a front matter section, whether it is a single or list (`_index.md`) file. See the **Front matter** section below for more details
+
+### Front matter
```
Crell, line 21: I think this can be brought up a header level, and thereby avoid the 4th level headers below.

CONTRIBUTING.md
```diff
@@ -16,15 +16,186 @@ GitBook (which we use for documentation) uses Markdown as its file format.
...
+`new_language.md` should contain at least the following in its front matter at the top of the page.
+
+```
+---
+title: "New language"
+weight: 4
+---
+```
+
+`title` sets the `<h1>` header for the page title, and weight defines in what order items in this subsection should be listed in the sidebar. Here the new page will appear fourth under the **Languages** section.
```
Crell, line 49: To clarify: Is it higher number before lower number, or lower before high?  Make that explicit, because different systems do it differently, just to piss you off.

CONTRIBUTING.md
```diff
@@ -16,15 +16,186 @@ GitBook (which we use for documentation) uses Markdown as its file format.
...
+weight: 6
+layout: single
+---
+```
+
+`_index.md` will now use the layout for single pages (`layouts/_default/single.html`).
+
+#### Sidebar and pages
+
+If you would like the title of a given page to appear differently in the sidebar - say, a shorter title there, but a more descriptive title on the page). You can use the `sidebarTitle` attribute in the front matter to define it, while the regular `title` attribute will only be used on the page itself.
```
Crell, line 83:
```suggestion
If you would like the title of a given page to appear differently in the sidebar &mdash; say, a shorter title there, but a more descriptive title on the page &mdash; you can use the `sidebarTitle` attribute in the front matter to define it, while the regular `title` attribute will only be used on the page itself.
```

CONTRIBUTING.md
```diff
@@ -16,15 +16,186 @@ GitBook (which we use for documentation) uses Markdown as its file format.
...
+{{< note >}}
+Lorem ipsum
+{{< /note }}
+```
+
+This shortcode will automatically create the **Note:** first line for you.
+
+`note` accepts two optional parameters:
+
+- `theme`: default none, standard note. (`info`, `warning`). This will change the color and the default first line.
```
Crell, line 124:
```suggestion
- `theme`: May be one of `note, `info`, or `warning`.  Default is `note`.  This will change the color and the default first line.
```

CONTRIBUTING.md
```diff
@@ -16,15 +16,186 @@ GitBook (which we use for documentation) uses Markdown as its file format.
...
+- `title`: default dependent on theme. Over-rides that default.
```
Crell, line 125:
```suggestion
- `title`: default dependent on theme.  The default is "Note", "Info", or "Warning" depending on the `theme`.
```

CONTRIBUTING.md
```diff
@@ -16,15 +16,186 @@ GitBook (which we use for documentation) uses Markdown as its file format.
...
+<!-- Scaled image -->
+![alt-title](/images/current_image.png "0.5")
+
+<!-- Scaled image displayed inline, such as an icon in a sentence -->
+![alt-title](/images/current_image.png "0.1-inline")
+```
+
+#### Videos & asciinema
+
+Displaying videos and asciinema recordings requires their own shortcodes, but they have identical parameters:
```
Crell, line 150: Are there parameters beyond src?  (HTML video has a whole crapload of other options.)

CONTRIBUTING.md
```diff
@@ -16,15 +16,186 @@ GitBook (which we use for documentation) uses Markdown as its file format.
...
+highlight=python
+markdownify=false
+---
+
+from jwcrypto import jws, jwk
+
+{{< /codetabs >}}
+```
+
+The first tab *Elasticsearch* will read from a local file and highlight it for PHP, whereas the second tab will apply Python highlighting to the content below `---`. Individual tabs are separated with the `<--->` line.
```
Crell, line 188: What's markdownify all about, and what is the default?  (The default seems to be true, since everything I've seen specifies false, which feels backwards.)

CONTRIBUTING.md
chadwcarlson: I'm not sure I know where you're referring to here

CONTRIBUTING.md
chadwcarlson: 👍

CONTRIBUTING.md
chadwcarlson: lower number = higher in the sidebar. I'll  include

CONTRIBUTING.md
chadwcarlson: At the moment no. What would you  want to  add?

CONTRIBUTING.md
chadwcarlson: It will apply markdown by default to everything, except if the `file` you're using is a markdown file. Then you have to explicitly say make it markdown.

CONTRIBUTING.md
Crell: "Pages should start with a title denoted with #".  That's GitBook-ese.  Hugo uses the YAML frontmatter instead.

chadwcarlson: Ah, okay gotcha

CONTRIBUTING.md
Crell: At minimum I think `type` is required, but there's a whole lot more: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video

Ideally we'd just allow these to pass-through rather than explicitly supporting each of them?  (Only applies to video, not asciinema, which has a different set of possible attributes.)

chadwcarlson: Ya right now it's just set firm at `type="video/mp4"`. I'd say we set a few reasonable defaults and then accept everything to be overwritten.

CONTRIBUTING.md
```diff
@@ -6,25 +6,197 @@ In order to maintain a consistent style and voice throughout our documentation p
...
+
+`markdownify` is `false` by default, although everything placed below `---` will be Markdown-rendered properly. This attribute really only needs to be `true` in cases where the `file` you are using for the tab's content is itself a `.md` file.
+
+All of these parameters need not be present for `codetabs` to function properly. You could very easily only include `title` and `highlight` for a code snippet.
```
Crell, line 190:
```suggestion
Not all of these parameters need to be present for `codetabs` to function properly.  The only required properties are `title` and `highlight`. 
```

CONTRIBUTING.md
```diff
@@ -6,25 +6,197 @@ In order to maintain a consistent style and voice throughout our documentation p
...
+title=Elasticsearch
+file=static/files/fetch/examples/php/elasticsearch
+highlight=php
+---
+
+<--->
+
+---
+title=Memcached
+file=none
```
Crell, line 177:
```suggestion
```

PR #1396, opened 2020-05-03
Restore items lost in migration to Hugo

chadwcarlson:


chadwcarlson:
Final commit actually should be `1379-Tiny typo`

PR #1403, opened 2020-05-07
It's cleaner to install this as a dependency.

OriPekelman:


Crell:
Does this make the CLI available in the path?  I think it would only do so if the Drupal template's .environment file is present.

pjcdawkins:
The "dependency" there uses the equivalent of `composer global install` which is far from clean.

As soon as you add more than one PHP dependency, that means the versions of each tool's dependencies are then tied up together and will resolve to the lowest supported denominator, if you see what I mean. And quite frequently, those dependencies will conflict.

The Phar is far cleaner, it gives everyone the same code.

SerheyDolgushev:
Sorry, but in my personal opinion, the cleanest approach would be to have CLI preinstalled by default in all images.

We talked to @OriPekelman about this before, and I understand there are some cases/customers who won't be very excited to have it preinstalled. But maybe there is an option for such kind customers to have a hidden `build` hook, which will remove the Platform.sh CLI, just to keep them happy?

Crell:
Per @pjcdawkins I'm closing this one.  We'll stick with the phar file.

PR #1404, opened 2020-05-07
Clarify endpoints, service and relationship names.

chadwcarlson:


docs/layouts/shortcodes/endpoint-clarification.html
```diff
@@ -0,0 +1,37 @@
+
+{{ $data := index .Site.Data.registry ( .Get 0 )}}
+{{ $link := ""}}
+{{ $multiples := "" }}
+{{ if or ( eq (.Get 0) "solr") (eq (.Get 0) "mariadb") }}
+  {{ $link = (.Get 1) }}
+  {{ $multiples = (.Get 2)}}
+{{ end }}
+
+<div class="alert alert-primary" role="alert">
```
Crell, line 10: This code looks like an unmaintainable mess.  I can't even read what it's saying from the source.  What is it...?

At minimum this file needs documentation itself, but just glancing at it I don't even know what it's doing.

docs/layouts/partials/examples/relationship.html
```diff
@@ -0,0 +1,9 @@
+<!-- Creates an example relationships block from the registry's example naming in `.docs` -->
+<!-- Note: this will only replace the endpoint and service type for demonstration purposes. -->
+<!-- To use this partial, pass the image data as the context. -->
+<!-- Ex: partial "examples/relationship" $data  -->
```
Crell, line 4: 1) Will these comments show up in the output?  Doesn't the template syntax have its own comment syntax to keep that from happening?
2) Why 4 comments instead of one multi-line comment?

(Applies to other files, too.)

chadwcarlson: 1. They won't show up in the output. The only thing that will is what's passed to return.

chadwcarlson: 2. because. I fix

docs/layouts/shortcodes/endpoint-clarification.html
```diff
@@ -0,0 +1,51 @@
+
```
Crell, line 1: The name "clarification" seems wrong to me.  Anything that needs to be clarified should be explained better in the first place. :smile:  endpoint-description?  endpoint-example?

chadwcarlson: Fixed

PR #1410, opened 2020-05-12
Clarifying which port should be used and when. 

matthiaz:
Customers are sometimes trying to use port 465 for our SMTP_HOST which is not available.
The earlier mention of port 465 is correct. But a section that clarifies this close to the connection instructions might help.

### DIFF
--- a/docs/src/development/email.md
+++ b/docs/src/development/email.md
@@ -58,3 +58,12 @@ mounts:
         source: local
         source_path: spool
 ```
+
+
+### Ports
+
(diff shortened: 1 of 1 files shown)


docs/src/development/email.md
```diff
@@ -58,3 +58,12 @@ mounts:
         source: local
         source_path: spool
 ```
+
+
+### Ports
```
Crell, line 63:
```suggestion
## Ports
```

PR #1414, opened 2020-05-14
Fixes links.

chadwcarlson:
- Fixes old/default link styling that doesn't account for uglyURLs
- List pages don't have a `.md` file, so needs to be explicit `.html` or `/`.
- Fixes DevRel drift link to use  shortcode
- Updates contributing.md


CONTRIBUTING.md
```diff
@@ -15,8 +15,8 @@ Hugo (which we use for documentation) uses Markdown as its file format.
 * Inline code statements, file names, and keys that would appear in a file should use backticks ``like this``.
 * Longer code samples should be denoted with triple backticks before and after, with no extra whitespace between the backticks and the code block. Always specify the language of the code block.  See the [highlight.js docs](https://highlightjs.org/static/demo/) for available language options.  (`yaml`, `bash`, and `php` are the most common we're likely to see).
 * Always use inline links.
-* Do not hard-wrap prose text.  Set your text editor to soft wrapping.
-* Internal links should be absolute (starting with `/`) and link to a file ending in `.md`, not `.html`. That makes them easier to find when needed to update.
+* Do not hard-wrap prose text. Set your text editor to soft wrapping.
+* Internal links should be absolute (starting with `/`). Single pages (i.e. `/configuration/services/elasticsearch.md` should be linked using the `.md` suffix. List pages `/configuration/services/_index.md`, which will be served from `/configuration/services.html` can use either the explicit `.html` suffix or no suffix at all).
```
Crell, line 19: We should be consistent, otherwise it's bad for SEO.  I'm not sure which I like less. :smile:

docs/src/configuration/app/upgrading.md
```diff
@@ -71,7 +71,7 @@ directly by the web server. Responses served from passthru URLs continue to use
 whatever caching headers were set by the application..
 
 To enable caching on your static files, make sure you include an `expires` key
-in your [web configuration](/configuration/app/web/),
+in your [web configuration](/configuration/app/web.md),
```
Crell, line 74: Seems we should remove the needless line break here while we're at it.

docs/src/configuration/services/memcached.md
```diff
@@ -38,7 +38,7 @@ Now add a relationship in your `.platform.app.yaml` file:
 
 {{< endpoint-description "memcached" >}}
 
-If you are using PHP, configure the relationship and enable the [PHP memcached extension](/languages/php/#php-extensions/) in your `.platform.app.yaml`.  (Note that the `memcached` extension requires `igbinary` and `msgpack` as well, but those will be enabled automatically.)
+If you are using PHP, configure the relationship and enable the [PHP memcached extension](/languages/php.html#php-extensions/) in your `.platform.app.yaml`.  (Note that the `memcached` extension requires `igbinary` and `msgpack` as well, but those will be enabled automatically.)
```
Crell, line 41:
```suggestion
If you are using PHP, configure the relationship and enable the [PHP memcached extension](/languages/php/extensions.md) in your `.platform.app.yaml`.  (Note that the `memcached` extension requires `igbinary` and `msgpack` as well, but those will be enabled automatically.)
```

docs/src/development/faq.md
```diff
@@ -79,7 +79,7 @@ Dedicated projects can deploy production to any public AWS or Azure region.
 
 'sh' is the short version of shell.
 
-According to Wikipedia™, in computing, a [shell](https://en.wikipedia.org/wiki/Shell_(computing\)) is a user interface for access to an operating system's services. Generally, operating system shells use either a [command-line interface ](https://en.wikipedia.org/wiki/Command-line_interface) (CLI) or [graphical user interface](https://en.wikipedia.org/wiki/Graphical_user_interface) (GUI).  This is exactly what Platform.sh is about: Giving developers tools to build, test, deploy, and run great websites!
+According to Wikipedia™, in computing, a [shell](https://en.wikipedia.org/wiki/Shell_(computing)) is a user interface for access to an operating system's services. Generally, operating system shells use either a [command-line interface ](https://en.wikipedia.org/wiki/Command-line_interface) (CLI) or [graphical user interface](https://en.wikipedia.org/wiki/Graphical_user_interface) (GUI).  This is exactly what Platform.sh is about: Giving developers tools to build, test, deploy, and run great websites!
```
Crell, line 82: The \) was needed to keep it from mis-rendering the closing ), which is supposed to be part of the link.  Does Hugo handle that differently than GitBook?

Crell:
So I know you already switched the index pages to be .html links, but I feel like a directory link would be better if both work, as then we don't have to keep track of which links are html for a reason and which are because someone goofed and didn't put .md.

That said, doesn't Hugo have its own internal-links syntax?  Shouldn't we be using that instead?

chadwcarlson:
We can sure, and that's ideal because the build will fail if the link's invalid.  

It means replacing internal links like this:

```
[Something]({{< relref "/languages/nodejs/_index.md#support-libraries" >}})
```

docs/src/development/faq.md
chadwcarlson: Yeah, it doesn't show [correctly anyway](https://docs.platform.sh/development/faq.html#why-did-you-choose-the-sh-extension-for-your-domain) and removes the ) which breaks the link

Crell:
Kind of verbose, but it seems like it would give us the best result.  1) One syntax; 2) Compile-time checked;

PR #1415, opened 2020-05-15
Integrations

Crell:
Proposed reorganization of the integrations section into its own top-level.

I'm debating if we shouldn't group it more; source integrations (Github, BB, Gitlab), webhooks (activity scripts, webhooks, and related bits), and profiling (blackfire, tideways, new relic).  I don't know if that would be hiding things too much, though.  I also don't know what we'd want to put on the landing pages.  Open to discussion on this point.


docs/src/integrations/profiling/_index.md
```diff
@@ -0,0 +1,7 @@
+---
+title: "Profiling"
+weight: -3
+description: Platform.sh supports a number of profiling services for optimizing your code.
+---
+
```
chadwcarlson, line 6: So there may be some funkiness here we need to iron out. 

Right now, this is a list page, and will include those link blocks for Tideways, Blackfire and New Relic under this block of text. Since list pages don't normally have text in them, the `{{< description >}}` shortcode is automatically  placed at the top of the page if its present - so you have  duplicated text. 

So for this page (and `source-integrations/_index.md`)  I'd say either

- set `layout:single` to remove the block links. and the automatic descriptions.
- remove the duplicated first sentence from the main content and  keep everything else as is.

PR #1417, opened 2020-05-18
adds the lastest version

otaviojava:


Crell:
This needs more context.  It's about installing the latest version of npm, whereas the rest of the page is about installing a custom version of nodejs.  They're separate projects.

PR #1422, opened 2020-05-19
Update rabbitmq.md 

Azurency:
The hostname in `/etc/hosts` seems to be **rabbitmq.internal** and not **mq.internal**. 
I don't know if it's relative to a relationship name, if it is it might bu useful to mention it also.

### DIFF
--- a/docs/src/configuration/services/rabbitmq.md
+++ b/docs/src/configuration/services/rabbitmq.md
@@ -81,7 +81,7 @@ highlight=python
 For debugging purposes, it's sometimes useful to be able to directly connect to a service instance. You can do this using SSH tunneling. To open a tunnel, log into your application container like usual, but with an extra flag to enable local port forwarding:
 
 ```bash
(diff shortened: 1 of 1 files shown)


Crell:
The host name is always the relationship name, dot internal.  We have a standard convention for the docs for how to name things, but if your relationship is named differently then the domain will be named differently as well.

Thanks for the PR attempt, though.  We do appreciate it!

Azurency:
Ok, thank you for your answer, it's nice to know. It might be a good idea to actually say exactly that in the documentation. 
It's not explained anywhere and it's not even coherent with the relationship name of the example relation a few lines above. Should I make another small pull request with this info ?

Crell:
Mm, looking at the page again, you're right.  It's internally inconsistent.  It should have a relationship name of `rabbitmqqueue`.

Can you update this PR or create a new one (either way) that changes it to `rabbitmqqueue`?

Azurency:
I updated it here. The shh command now to uses `rabbitmqqueue` and not `mq` 👍

Crell:
Thanks!

PR #1429, opened 2020-05-27
Note that the Outbound Firewall is going to be tier gated.

Crell:
We may also want to change the styling on these blocks.  Open to suggestions there.

### DIFF
--- a/docs/src/configuration/app/firewall.md
+++ b/docs/src/configuration/app/firewall.md
@@ -1,11 +1,18 @@
 ---
-title: "Outbound firewall"
+title: "[Beta] Outbound firewall"
+sidebarTitle: "Outbound firewall"
 weight: 15
 ---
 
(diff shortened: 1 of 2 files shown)


docs/src/configuration/app/firewall.md
```diff
@@ -1,12 +1,19 @@
...
+title: "[Beta] Outbound firewall"
+sidebarTitle: "Outbound firewall"
 weight: 15
 ---
 
 In some situations, compliance regulations may require you to limit outbound traffic from your application.  The `firewall` property allows you to do so.
 
 This setting has no impact on inbound requests to your application.  For that, use the environment access control settings in the Management Console.
 
+{{< note >}}
```
chadwcarlson, line 11: Do you mean updating the style of these blockquotes/notes? If so you can pass `theme="warning"` or `theme="info"` to them.

Crell: note is what was used before.  We may want to update these to something else; I'm open to suggestions, now or later.

PR #1437, opened 2020-06-02
Update _index.md

tristanblattmansh:
just a tweak


Crell:
This seems incorrect.  The "it" is the subject of the clause.  Without it, the clause is not referring to anything.

PR #1441, opened 2020-06-03
Update self-indexing: valid json, some cleanup

chadwcarlson:
- all self-indexing moved out  of theme
- `index.json.json` uses [jsonify](https://gohugo.io/functions/jsonify/) to ensure final object is valid json. breaks otherwise
- `json/createIndex` removed, broken up into three smaller  partials.
- all `registry` pages + `other` list page not searchable

### DIFF
--- a/docs/layouts/_default/index.json.json
+++ b/docs/layouts/_default/index.json.json
@@ -1,9 +1,11 @@
 {{- $index := slice -}}
 {{- range .Site.Pages -}}
-  {{- $url := chomp ( partial "meilindex/getURL" . ) -}}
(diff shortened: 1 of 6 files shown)


docs/layouts/_default/index.json.json
```diff
@@ -0,0 +1,11 @@
+{{- $index := slice -}}
```
Crell, line 1: Woah, why is GitHub marking all of this is red?

chadwcarlson: Maybe  because the  json template is `.json` but  isn't itself valid json?

PR #1442, opened 2020-06-04
automatic https redirection sometimes needs to be more explicit.

dman-coders:
Clarify that sometimes you need to explicitly redirect from http://(no-www).domain.tld to  https://www.domain.tld explicitly in some cases. The 'automatic' redirect fails if they are enforcing a www prefix.
As per #devrel 
https://platformsh.slack.com/archives/CDCCS5ZUN/p1591250957364400


Crell:
I'm pretty sure this is a code bug, not doc bug.  The behavior described in docs now has worked in the past without issue.  "In some cases" suggests we need to identify and correct those cases.

PR #1457, opened 2020-06-12
Document how to avoid deadlock issues on TYPO3

Crell:
cf:
https://github.com/platformsh-templates/typo3/pull/7
https://github.com/platformsh/template-builder/pull/270

### DIFF
--- a/docs/src/frameworks/typo3/_index.md
+++ b/docs/src/frameworks/typo3/_index.md
@@ -6,16 +6,32 @@ layout: single
 toc: false
 ---
 
-## Prerequisites
+The supported way to run TYPO3 on Platform.sh is through [Composer](https://getcomposer.org/download/).  If you do not already have it installed then you should do so before proceeding.
 
-**Composer**
+## Avoiding deadlock
 
-Composer is a tool for dependency management in PHP. It allows you to declare the dependent libraries your project needs and it will install them in your project for you.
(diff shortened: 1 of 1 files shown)


docs/src/frameworks/typo3/_index.md
```diff
@@ -6,16 +6,32 @@ layout: single
...
+A default TYPO3 installation has a risk of deadlocks when run on low-worker PHP-FPM configurations.  Specifically, TYPO3 handles 403 and 404 error pages by issuing a full HTTP request back to itself with no timeout, which can lead to process starvation.  There are two required steps to avoid this problem:
 
-- [Install Composer](https://getcomposer.org/download/)
+1. Add the line
+
+    ```php
+    $GLOBALS['TYPO3_CONF_VARS']['HTTP']['timeout'] = 3;
+    ```
+
+to your `typo3config/AdditionalConfiguration.php` file, which will set an HTTP timeout of 3 seconds instead of the default several minutes.  (You may select a different number, but keep it under 10 seconds.)
```
chadwcarlson, line 21: may want to match the indentation here

docs/src/frameworks/typo3/_index.md
```diff
@@ -6,16 +6,32 @@ layout: single
...
+
+2. Install and enable the [Local Page Error Handler](https://extensions.typo3.org/extension/pxa_lpeh/) plugin for TYPO3.  The easiest way to do so is through composer:
+
+    ```bash
+    composer require pixelant/pxa-lpeh
+    ```
+
+that will provide a non-loopback way to handle error pages which avoids this race condition entirely.
```
chadwcarlson, line 29: and for these  two paragraphs

PR #1449, opened 2020-06-10
Updates `whitelist` to `safe-list` across docs

JGrubb:

PR #1458, opened 2020-06-12
Clean up and clarify list terms

pjcdawkins:
Follows on from #1449 

Removes `blacklist` and `whitelist` and uses the term `safelist` as a noun or verb where there isn't a clearer alternative.


Crell:
Safelist is not a meaningful term, as noted in the previous issue.

PR #1459, opened 2020-06-15
rm white and blacklist, adopt allow/deny nomenclature

robertDouglass:
Resolves https://github.com/platformsh/platformsh-docs/pull/1449

### DIFF
--- a/docs/src/frameworks/wordpress/_index.md
+++ b/docs/src/frameworks/wordpress/_index.md
@@ -11,6 +11,6 @@ Platform.sh strongly recommends starting new WordPress projects from our [WordPr
 
 ## Plugin compatibility
 
-Platform.sh does not blacklist any WordPress plugins.  However, some plugins are known to require write access to their own file system as part of their setup process.  That is not possible on Platform.sh.  The file system where code lives is read-only for security reasons and cannot be written to from the application, only during the [build hook]({{< relref "/configuration/app/build.md" >}}).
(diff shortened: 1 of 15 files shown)


docs/src/frameworks/wordpress/_index.md
```diff
@@ -11,6 +11,6 @@ Platform.sh strongly recommends starting new WordPress projects from our [WordPr
 
 ## Plugin compatibility
 
-Platform.sh does not blacklist any WordPress plugins.  However, some plugins are known to require write access to their own file system as part of their setup process.  That is not possible on Platform.sh.  The file system where code lives is read-only for security reasons and cannot be written to from the application, only during the [build hook]({{< relref "/configuration/app/build.md" >}}).
+Platform.sh does not deny any WordPress plugins.  However, some plugins are known to require write access to their own file system as part of their setup process.  That is not possible on Platform.sh.  The file system where code lives is read-only for security reasons and cannot be written to from the application, only during the [build hook]({{< relref "/configuration/app/build.md" >}}).
```
Crell, line 14:
```suggestion
Platform.sh does not explicitly block any WordPress plugins.  However, some plugins are known to require write access to their own file system as part of their setup process.  That is not possible on Platform.sh.  The file system where code lives is read-only for security reasons and cannot be written to from the application, only during the [build hook]({{< relref "/configuration/app/build.md" >}}).
```

PR #1469, opened 2020-06-26
Update search to index sections within pages

chadwcarlson:
Partials added to split pages into *documents by section*, in an attempt to 
- give better linking
- overcome Meilisearch's 1000 word limit per attribute, that leave some pages un-searchable. https://github.com/meilisearch/MeiliSearch/issues/556#issuecomment-605333369
- ^ originally observed by @ralt that `request_buffering` was not found in https://docs.platform.sh/configuration/app/web.html#locations. 
- updates the search app to Python 3.8
- Adds a distinct attribute, which keeps multiple results for documents from the same page showing up.
- Small syntax changes for creating the Meilisearch client from upstream.

*WIP*
- ~`request_buffering` is still not found for the locations section, despite it being ~ 761 words (< 1000 word limit).~
- ~Checking with Meilisearch on possible bug, issue filed here https://github.com/meilisearch/MeiliSearch/issues/807~


docs/themes/avocadocs/layouts/shortcodes/repolist.html
```diff
@@ -10,25 +10,25 @@
 
     <div class="card">
       <div class="card-header-template" id="heading{{ .shortname }}">
-        <h5 class="template-card">
+        <div class="template-card">
```
Crell, line 13: 3 divs inside each other seems like a code smell.  Shouldn't some of them be section or article or aside or something with more semantic meaning?  Do we even need all 3?  Seems sketchy.

chadwcarlson: I can change this back. This is just left over from testing the split.

Crell: Let's change it back for now at least.  We can look into a markup cleanup later.

chadwcarlson: Forgot to resolve, this is back to an h5 now

PR #1470, opened 2020-06-29
update solr.md with latest developments

fruitshakes:
according to the latest release notes here https://www.drupal.org/project/search_api_solr/releases/4.1.2, the 4.1 version of search api solr no longer needs the extra steps to install solarium v6, however requires drupal 8.8 or higher, and php 7.2 or higher.


Crell:
Good to hear, thanks!

PR #1477, opened 2020-07-10
Add a note about deployments in worker section

Nek-:


Crell:
Thanks, @Nek- .  It's a bit more complex than that, though, so I've opened another PR after talking with Engineering to confirm the details: #1478 1478

PR #1492, opened 2020-07-21
Add a warning about elasticsearch deployments

Nek-:
Maybe I can also add that it depends also on plugins? Just tell me I'll do it.


Nek-:
This is wrong ES does not restart on each deploy if the configuration is not moved.

PR #1506, opened 2020-07-30
List new templates, double multi-apps. Remove utms.

chadwcarlson:
- updates `data/templates.yaml` with newer templates
- updates `fetch-templates.js`
   - moves `parseErrorTemplates` to global var
   - adds global var `multiAppTemplates` for overriding templates that would fail parse (script expects root `.platform.app.yaml`)
   - removes `utm` parameters from DoP button link in `templateData`
   - adds `addTemplateToData` function, pulling out new attribute logic to handle `multiAppTemplates`
- updates `repolist.html` (template accordian shortcode)
   - multiapp templates expanded in both languages. Shortcode adjusted with a  new hash `id` to fix that.
   - cleaned up the random  hanging  last  `li` created in  features block.


docs/static/scripts/fetch-templates.js
```diff
@@ -78,7 +86,7 @@ async function getTemplateRuntime(template){
       "repo": "https://github.com/platformsh-templates/" + template.name,
       "description": "",
       "image": "",
-      "deploy": `https://console.platform.sh/projects/create-project?template=https://raw.githubusercontent.com/platformsh/template-builder/master/templates/${template.name}/.platform.template.yaml&utm_content=${template.name}&utm_source=github&utm_medium=button&utm_campaign=deploy_on_platform`
+      "deploy": `https://console.platform.sh/projects/create-project?template=https://raw.githubusercontent.com/platformsh/template-builder/master/templates/${template.name}/.platform.template.yaml`
```
Crell, line 89: @christianyates Please confirm this is all we need for the links in docs?

PR #1512, opened 2020-08-10
Remove mention of restrict_robots

ollie-platform:
The restrict_robots setting does not have any impact on the x-robots-tag header on non-production environments by design. The setting also does not have any impact on the x-robots-tag header in production environments currently (which is the result of a bug). Removing references to restrict_robots in the public documentation in the meantime.


Crell:
Do we know when the bug on production is going to be fixed?  Simply having the toggle there at all seems problematic if it's not going to do anything, and then having the docs dance around that seems iffy to me.

PR #1518, opened 2020-08-13
Rename Introduction section have a path of introduction.

Crell:
Because because who who needs needs duplicates duplicates?


ralt:
Please make sure redirects are handled properly, SEO matters.

PR #1549, opened 2020-09-16
Typo3

chadwcarlson:
Complete WIP. Will resubmit as to not conflict with the Drupal guide. 

- project creation? instruct to do when  creating  the account, or hold off to do via the CLI
- integration assumption? Since these are primarily supposed  to be migration guides, shouldn't our assumption be that the  code they are migrating is on GH/GL/BB, and the final "deployment" steps are setting up the integration?
- these code snippets are pretty long  with all the comments
   - is it all that useful to intro a file, then say read  the comments?
   - That  may be our only choice when showing an app's configuration  files (i.e. `PlatformshConfiguration.php`), but our config yamls get  real long with comments, and we should be able to find a way to parse the remote yaml so that we can show individual keys within a snippet, breaking up the shortcode descriptions into smaller chunks. I don't know if that's desirable or not.


Crell:
Closing in favor of #1553.

PR #1550, opened 2020-09-22
note-about-timezone-in-logs

matthiaz:
We don't specially mention that logs will always be UTC, no matter the time zone you set.

### DIFF
--- a/docs/src/configuration/app/timezone.md
+++ b/docs/src/configuration/app/timezone.md
@@ -38,3 +38,7 @@ Setting the application timezone will only affect the application itself, not sy
 {{< note >}}
 In the vast majority of cases it's best to leave all timezones in UTC and store user data with an associated timezone instead.
 {{< /note >}}
+
+### Logs
+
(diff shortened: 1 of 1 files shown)


docs/src/configuration/app/timezone.md
```diff
@@ -38,3 +38,7 @@ Setting the application timezone will only affect the application itself, not sy
 {{< note >}}
 In the vast majority of cases it's best to leave all timezones in UTC and store user data with an associated timezone instead.
 {{< /note >}}
+
+### Logs
+
+[All logs](https://docs.platform.sh/development/logs.html) will always be logged in UTC. You cannot change the timezone for them and changing any of the above time zone settings will not affect this.
```
Crell, line 44:
```suggestion
[All logs](https://docs.platform.sh/development/logs.html) will always be logged in UTC. You cannot change the timezone for them and changing any of the above time zone settings will not affect log timestamps.
```

PR #1554, opened 2020-09-23
Use `#` for Python comments

a1phyr:
I hope it does not break Markdown


Crell (approved):

Crell (approved):
I was wrong, it's fine. :-)

PR #1570, opened 2020-10-06
Update Alernative Node.js installation instructions

raupie:
Remove redundant deploy hook step
Remove nvm current command as purely informational


raupie:
The deploy hook is redundant as it already picks up the new NodeJS version through the `.environment` file.

Also the `nvm current` is purely informational debug purposes.

docs/src/languages/nodejs/nvm.md
```diff
@@ -25,13 +25,7 @@ hooks:
         curl -o- https://raw.githubusercontent.com/creationix/nvm/$NVM_VERSION/install.sh | dash
         export NVM_DIR="$PLATFORM_APP_DIR/.nvm"
         [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
-        nvm current
```
Crell, line 28: Why do we need to remove this line?

raupie: I believe all it does is print the version. In this case `current` is what I noticed. Purely informational.

Crell (approved):

PR #1572, opened 2020-10-08
clarify with example

matthiaz:
How to set up the timezone confuses me. (and apparently some customers as well). So I'm clarifying with an example.

### DIFF
--- a/docs/src/configuration/app/timezone.md
+++ b/docs/src/configuration/app/timezone.md
@@ -17,6 +17,19 @@ It defines **relative to which timezone** your cron is being run (to avoid havin
 Its value is one of the [tz database region codes](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones) such as `Europe/Paris` or `America/New_York`.  This key will apply to all cron tasks defined in that file.
 This entry is only meaningful on cron specs that specify a particular time of day, rather than a "time past each hour".  For example, `25 1 * * *` would run every day at 1:25 am in the timezone specified.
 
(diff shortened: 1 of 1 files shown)


docs/src/configuration/app/timezone.md
```diff
@@ -17,6 +17,19 @@ It defines **relative to which timezone** your cron is being run (to avoid havin
 Its value is one of the [tz database region codes](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones) such as `Europe/Paris` or `America/New_York`.  This key will apply to all cron tasks defined in that file.
 This entry is only meaningful on cron specs that specify a particular time of day, rather than a "time past each hour".  For example, `25 1 * * *` would run every day at 1:25 am in the timezone specified.
```
Crell, line 18:
```suggestion
This entry is only meaningful on cron specs that specify a particular time of day, rather than a "time past each hour".  For example, the sample below would run every day at 1:25 am New York time (automatically adjusting for daylight savings).
```

docs/src/configuration/app/timezone.md
```diff
@@ -17,6 +17,19 @@ It defines **relative to which timezone** your cron is being run (to avoid havin
...
 
+For example:
```
Crell, line 20:
```suggestion
```

Crell:
Let's couple the text with the example better so that they flow nicely.

chadwcarlson (approved):

PR #1573, opened 2020-10-08
eZ Systems -> Ibexa

dabrt:
Update the article to account for a company name change. Might be a good idea to change this elsewhere as well. I will look into that.


Crell:
Off hand I think that's the only mention of the name; we refer to the software in a few other places, but that's probably the only company name mention.  Thanks.

Crell (approved):

PR #1581, opened 2020-10-21
Use a safer way to install the CLI

opdavies:


Crell:
Why is this safer?  You're still running Arbitrary Code From The Internet(tm).

pjcdawkins:
It would only be safer security-wise if it verified the installer's code via a hash.

I think @opdavies means it's more compatible with shells that for some bizarre reason don't support `| php`, a notable example being MINGW64, which comes with Git Bash. Unfortunate. See https://github.com/platformsh/platformsh-cli/pull/992

Note Composer's download page goes for the complicated multiple-PHP-command option, probably having learned from experience that the simple one-liner isn't good enough: https://getcomposer.org/download/

opdavies:
It was a comment from one of our DevOps team. I asked him for some more information and he sent me this:

1. It gives you a way to review it.
2. If the download of the script is interrupted, it won't execute a partially downloaded script. Less problematic for php when compared to bash due to not wanting to execute rm -rf / when it aborts the download

Composer's one checks if the download of the script worked. Platform's doesn't.

pjcdawkins:
True. As he says it's less problematic for PHP. At the moment I can't see a way that it could produce any adverse effect if only part of the file were downloaded: in most cases there would be a syntax error, in rare cases a fatal error before any files are touched. (It may be a micro-enhancement to move the execution of `run()` to the end of the installer file)

The one-liner install command is pretty valuable. I've noticed another unfortunate thing about the default terminal with Git Bash: it seems to only let you copy and paste one line of code at a time.

pjcdawkins:
We could add `-f` to the command:

```sh
curl -sfS https://platform.sh/cli/installer | php
```

This ensures that `php` won't receive any input from `curl` if there was any download error, and so it'll do what PHP normally does with an empty input (it will print an empty output).

Re. reviewing the installer code, I think it's reasonably clear that it's an option available to users who would want to do that, but this isn't the place - if you are copying and pasting an "install command" you expect it to run.

docs/src/development/cli/api-tokens.md
```diff
@@ -58,7 +58,8 @@ Second, add a build hook to your `.platform.app.yaml` file to download the CLI a
 ```yaml
 hooks:
     build: |
-        curl -sS https://platform.sh/cli/installer | php
+        curl -o /tmp/platform-installer.php -sS https://platform.sh/cli/installer
+        php /tmp/platform-installer.php
```
pjcdawkins, line 62:
```suggestion
        curl -fsS https://platform.sh/cli/installer | php
```
see other comments for reasoning

PR #1593, opened 2020-10-30
Small updates

chadwcarlson:
- `/search`
   - Updates cryptography in `search` - Resolves https://github.com/platformsh/platformsh-docs/pull/1589
   - https://github.com/platformsh/platformsh-docs/pull/1589 was failing in the search app, due to upgrades in Meilisearch. Fix there was clearing the mount. 
   - Update to `update_search` to clear `data.ms`. In this issue above, upstream Meili was updated, but there was old data on the mount that wasn't compatible and broke everything. So this line will avoid that in the future, since we're reindexing on every deploy anyway.
- `/docs`
   - Update `docs` dependencies
   - Upgrades Hugo to 0.77.0
   - Updates template data to include newer templates
   - `s/Services/Features` on template accordians
   - Moves `repolist` accordian shortcode out of the theme, because it never really did belong there in the first place, did it?


search/Pipfile
```diff
@@ -6,7 +6,7 @@ verify_ssl = true
 [dev-packages]
```
Crell, line 6: Should this get shifted to Poetry as well while we're at it?

Crell (approved):

PR #1595, opened 2020-11-03
Lock meilisearch version

chadwcarlson:
Locks Meilisearch version instead of using the [`download-latest`](https://github.com/meilisearch/MeiliSearch/blob/master/download-latest.sh) script, since this resulted in a break previously. Other small cleanup.


RudyWeber (approved):
✨

PR #1579, opened 2020-10-19
Improvement to TYPO3 Deployment Guide

mabolek:
Special focus on making the guide applicable to users moving an existing site to Platform.sh. The difference between this approach and that used in the deploy template is quite considerable.

(This PR is input to @chadwcarlson and shouldn't be merged outright.)

PR #1597, opened 2020-11-04
TYPO3 guide adjustments

chadwcarlson:
Resolves: https://github.com/platformsh/platformsh-docs/pull/1579 

- `config-app.html`, `drupal9/deploy/configure.md`, & `typo3/deploy/configure.md`
   - Mathias included a [reference](https://github.com/platformsh/platformsh-docs/pull/1579/files#diff-b08d730cd3aa18529e8cb13c98ad2ee16bbc4b22b0ee8e1d9af005643e0dd32c) to a line within the `.platform.app.yaml` file that should be included. As we have it now, we would need to place it after the `config-app` shortcode is used, and these files are just too long for that to look right. This updates `config-app` to accept inner content for detail like this, and updates Drupal/configure to use the same. 
- `config-desc.html`
  - We have 3 starting points, and two of them will error `mkdir: .platform: File exists` on the command given. `mkdir .platform` is removed.
  - Since this won't be the case for integrations, I added a line at the end of `From a third party Git provider` in `guides/signup` that instructs users to `set-remote` to enable SSH so all are on the same page.
- `customize.md`
  - Replicated Mathias change [here](https://github.com/platformsh/platformsh-docs/pull/1579/files#diff-9479988e7992b1562f7c98a2a88c63be9e3032b164a68d51da948d01114be097)
- No change/follow-up:
   - Mathias removed the [post-install steps](https://github.com/platformsh/platformsh-docs/pull/1579/files#diff-1f68c8160245fbc75db0b95d10dd6e47ac452103b74da1019b494cd9cb5efb65) since migrators won't need to update their admin passwords. I don't know, this still seems relevant to include to me.
   - Mathias also removed most of the build and deploy hooks in the [hard-coded `.platform.app.yaml` file](https://github.com/platformsh/platformsh-docs/pull/1579/files#diff-b08d730cd3aa18529e8cb13c98ad2ee16bbc4b22b0ee8e1d9af005643e0dd32c).  Again here, they seem relevant to the other two cases, but just looking at it I would still think these are steps that you're going to want to run on first deploy, and then migrate data. Unless I'm missing something here, hopefully Mathias can clarify.


docs/layouts/shortcodes/guides/config-desc.html
```diff
@@ -14,7 +14,7 @@
 
 <div class="highlight">
 <pre class="chroma"><code class="language-bash" data-lang="bash"><span class="c1"># Create empty Platform.sh configuration files</span>
-$ touch .platform.app.yaml && mkdir .platform && touch .platform/routes.yaml && touch .platform/services.yaml</code></pre>
+$ touch .platform.app.yaml && touch .platform/routes.yaml && touch .platform/services.yaml</code></pre>
```
Crell, line 17: Why would `.platform` already exist?  That doesn't make sense to me.

docs/layouts/shortcodes/guides/signup.html
```diff
@@ -1,16 +1,31 @@
 <p>Depending on whether or not you have an existing project you want to migrate to Platform.sh there are three ways to get started.</p>
 
-<h3>With a template</h3>
+<h3>From a third party Git provider</h3>
```
Crell, line 3: I'm not convinced about making this the first option, because if you haven't setup the config files you'll get an error immediately upon adding the integration.  That is not a good user experience.  Also, we deliberately are trying to balance new and migration users in these guides.  Shifting it entirely to a migration-centric focus is a change of direction I'm not comfortable with at all.

docs/layouts/shortcodes/guides/config-desc.html
chadwcarlson: - If you clone from an initialized template, it's gonna have .platform
- When we give them the steps to create an empty project, we give them set-remote, which will create .platform
- So integration was the odd man out here.

docs/layouts/shortcodes/guides/signup.html
chadwcarlson: That's a fair point

docs/src/guides/typo3/deploy/configure.md
```diff
@@ -31,4 +31,8 @@ If a service stores persistent data then it will also have a `disk` key, which s
 
 {{< guides/config-app template="typo3" >}}
 
+Note that the command `php vendor/bin/typo3cms install:generatepackagestate` is run during the build hook. The command will make sure all installed extensions are enabled and that they can be ommitted if you commit your own [`PackageStates.php`](https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ExtensionArchitecture/ExtensionManagement/Index.html#package-manager) file.
+
```
Crell, line 35: Why not just move this text above the config-app include statement?  Why the inside-out inclusion for it?

chadwcarlson: Because we have a paragraph of shared text in that shortcode - this seemed like the best way to include any framework-specific notes after that paragraph

docs/layouts/shortcodes/guides/signup.html
chadwcarlson: Reverted this file back to original order

docs/layouts/shortcodes/guides/config-desc.html
Crell: * The guides say if you are using a template then all of this stuff is there already, just read through for understanding. So that's not relevant.
* Wait, `set-remote` creates `.platform`? ... For the local dir, right.  Blargh.  Is there a mkdir alternative that doesn't fail if the directory already exists?  That would be the better solution on that one.

docs/src/guides/typo3/deploy/configure.md
Crell: Hm, fair.

docs/layouts/shortcodes/guides/config-desc.html
chadwcarlson: Good point. `mkdir -p .platform`

Crell: Let's do that, then.

PR #1598, opened 2020-11-05
Update postgresql.md

rhubinak:
Added mention about correct DB dump format

### DIFF
--- a/docs/src/configuration/services/postgresql.md
+++ b/docs/src/configuration/services/postgresql.md
@@ -128,7 +128,7 @@ platform db:dump --stdout | bzip2 > dump.sql.bz2
 
 ## Importing data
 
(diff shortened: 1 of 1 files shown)


docs/src/configuration/services/postgresql.md
```diff
@@ -128,6 +128,12 @@ platform db:dump --stdout | bzip2 > dump.sql.bz2
 
 ## Importing data
 
+Please make sure that the imported file contains objects with cleared ownership and `IF EXISTS` clauses. For example, you can create a DB dump with following parameters:
```
Crell, line 131:
```suggestion
Make sure that the imported file contains objects with cleared ownership and `IF EXISTS` clauses. For example, you can create a DB dump with following parameters:
```

PR #1599, opened 2020-11-05
Keys endpoint instead of ns

chadwcarlson:
Since the poetry changeover on search, there have been things to iron out to get search working properly again:
- https://github.com/platformsh/platformsh-docs/pull/1593 was the original switch
- https://github.com/platformsh/platformsh-docs/pull/1594 Better handled venv creation to unbreak 1593
- https://github.com/platformsh/platformsh-docs/pull/1595 Added a locking variable so Meilisearch stuck to a single version, since the previous `get-latest` script caused a mismatch with existing data that caused the whole thing to break. 

Right now, a network storage mount is being used to share 
- the public api key from Meilisearch with docs, so search can be placed from within the docs search bar
- the Hugo-generated docs index with search, so its included in the final index

At the moment, network storage isn't visible via SSH, and so there is no token to get search working properly. This PR will pick up where I had originally left off - relying instead on:
- Meilisearch's [`/keys`](https://docs.meilisearch.com/references/keys.html#get-keys) endpoint to retrieve the token in docs
- The served `docs.platform.sh/index.json` to retrieve the Hugo index for the search app

There has been weirdness with NS in the past with this work-around, and with the `/keys` endpoint its no longer necessary, and we can get around this problem now and hopefully going forward.


chadwcarlson:
Closing in favor of https://github.com/platformsh/platformsh-docs/pull/1601

PR #1601, opened 2020-11-06
Search fixes and cleanup

chadwcarlson:
Cleaner resubmit of https://github.com/platformsh/platformsh-docs/pull/1599

Since the poetry changeover on search, there have been things to iron out to get search working properly again:
- https://github.com/platformsh/platformsh-docs/pull/1593 was the original switch
- https://github.com/platformsh/platformsh-docs/pull/1594 Better handled venv creation to unbreak 1593
- https://github.com/platformsh/platformsh-docs/pull/1595 Added a locking variable so Meilisearch stuck to a single version, since the previous `get-latest` script caused a mismatch with existing data that caused the whole thing to break. 

Right now, a network storage mount is being used to share 
- the public api key from Meilisearch with docs, so search can be placed from within the docs search bar
- the Hugo-generated docs index with search, so its included in the final index

At the moment, network storage isn't visible via SSH, and so there is no token to get search working properly. This PR will pick up where I had originally left off - relying instead on:
- Meilisearch's [`/keys`](https://docs.meilisearch.com/references/keys.html#get-keys) endpoint to retrieve the token in docs
- The served `docs.platform.sh/index.json` to retrieve the Hugo index for the search app
- The templates index data file (`data/templates.yaml`) moved into the built Hugo site so it too can be pulled by search. 

There has been weirdness with NS in the past with the original work-around, and with the `/keys` endpoint its no longer necessary, and we can get around this problem now and hopefully going forward.


docs/deploy.sh
```diff
@@ -0,0 +1,17 @@
+
+#!/usr/bin/env bash
+
+createSearchConfig(){
```
Crell, line 4:
```suggestion
createSearchConfig() {
```

docs/deploy.sh
```diff
@@ -0,0 +1,17 @@
...
+    # Get the data
+    MEILI_CONFIG_DEST=public/scripts/xss/dist/config/config.json
+    MEILI_INDEX="docs"
+    MEILI_TOKEN=$(curl -s -H "X-Meili-API-Key: $PLATFORM_PROJECT_ENTROPY" -X GET "search.internal/keys" | jq -r ".public")
+    MEILI_URL=$(echo $PLATFORM_ROUTES | base64 --decode | jq -r 'to_entries[] | select(.value.id == "search") | .key')
+    # Delete config file in the mount if it exists
+    [ ! -e $MEILI_CONFIG_DEST ] || rm $MEILI_CONFIG_DEST
```
Crell, line 11:
```suggestion
    rm -f $MEILI_CONFIG_DEST
```

I think this does the same thing (rm and don't error if it's not there), but is shorter.

search/main.py
```diff
@@ -57,60 +56,47 @@ def __init__(self):
...
+    
+    def add_documents(self, index):
+        documents = [f for f in glob.glob("{}/*.json".format(self.scrape_dir))]
+        for docs in documents:
+            self.add(docs, index)
 
+    def add(self, docs, index):
+        with open(docs) as scraped_index:
+            data = json.load(scraped_index)
+            index.add_documents(data)
```
Crell, line 92:
```suggestion
    def add_documents(self, index):
        documents = [f for f in glob.glob("{}/*.json".format(self.scrape_dir))]
        for doc in documents:
            self.add(doc, index)

    def add(self, doc, index):
        with open(doc) as scraped_index:
            data = json.load(scraped_index)
            index.add_documents(data)
```

This confused me for a while, as it seemed you were reopening a group of files over and over again.

Crell: Is this in any way parallelizable?

search/ABOUT.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Build
+
+### `docs`
+
+1. Building autocomplete
+
+    Within Hugo there is a React autocomplete app at `doc/static/scripts/xss`. The relevant files here are:
+
+    - `.../xss/src/components/[Suggestions|SuggestionsPrimary.js`: This defines the template for search results appearing in the dropdown. They differ between Primary results (results from the documentation), and other results (from the API Docs, marketing site, templates, etc.).
```
Crell, line 13:
```suggestion
    - `.../xss/src/components/[Suggestions|SuggestionsPrimary].js`: This defines the template for search results appearing in the dropdown. They differ between Primary results (results from the documentation), and other results (from the API Docs, marketing site, templates, etc.).
```

I assume?

search/ABOUT.md
```diff
@@ -0,0 +1,84 @@
...
+    - `.../xss/src/containers/Search.js`: Where most of the configuration for the search app happens. All configuration needed to query Meilisearch is done here. 
+        ```js
+        let config = {}
+        const request = async () => {
+            const response = await fetch("/scripts/xss/dist/config/config.json");
+            config = await response.json();
+        }
+        request();
+        ```
+        The `config.json` file does not yet exist and is only later built during the deploy hook when the `search` container becomes available. Webpack isn't a fan of reading from `config-reader-nodejs` or environment variables here if they are not yet set, but a file works just fine. We also have defined a mount in the *built* Hugo site at `public/scripts/xss/dist/config/` do allow writing this file in the deploy hook.
```
Crell, line 23:
```suggestion
        The `config.json` file does not yet exist and is only later built during the deploy hook when the `search` container becomes available. Webpack isn't a fan of reading from `config-reader-nodejs` or environment variables here if they are not yet set, but a file works just fine. We also have defined a mount in the *built* Hugo site at `public/scripts/xss/dist/config/` to allow writing this file in the deploy hook. 
```

search/ABOUT.md
```diff
@@ -0,0 +1,84 @@
...
+        ```
+        The `config.json` file does not yet exist and is only later built during the deploy hook when the `search` container becomes available. Webpack isn't a fan of reading from `config-reader-nodejs` or environment variables here if they are not yet set, but a file works just fine. We also have defined a mount in the *built* Hugo site at `public/scripts/xss/dist/config/` do allow writing this file in the deploy hook. 
+
+2. Copy the templates index
+
+    We want to include the templates index (`docs/data/templates.yaml`) in the search results, but Hugo doesn't include `data` in its final build. This moves a copy of the template index file into `static` so that it can be requested post-build. 
+
+3. Build the Hugo site
+
+    Hugo allows you to define custom output formats besides HTML. We've defined a `json` output format as well for the root `_index.md` location, which uses the below files to index the documentation itself and serve that index at `docs.platform.sh/index.json` (`public/index.json` when built).
```
Crell, line 31:
```suggestion
    Hugo allows you to define custom output formats besides HTML. We've defined a `json` output format as well for the root `_index.md` location, which uses the files below to index the documentation itself and serve that index at `docs.platform.sh/index.json` (`public/index.json` when built).
```

Crell (approved):

PR #1604, opened 2020-11-12
Update session example to use Persistent redis

andrerom:
Unsure where to update the YML, but session example should rather use Persistent to make sure sessions aren't easily lost.
Or maybe there should be doc exampling pros/cons of using cache or persistent redis flavour here.

### DIFF
--- a/docs/src/configuration/services/redis.md
+++ b/docs/src/configuration/services/redis.md
@@ -177,13 +177,13 @@ Using the same configuration but with your Redis relationship named `sessionstor
 
 `.platform/services.yaml`
 
-{{< readFile file="src/registry/images/examples/full/redis.services.yaml" highlight="yaml" >}}
+{{< readFile file="src/registry/images/examples/full/redis-persistent-session.services.yaml" highlight="yaml" >}}
 
 `.platform.app.yaml`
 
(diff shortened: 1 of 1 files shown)


Crell:
The most common redis usage is caching, not sessions, for which the ephemeral redis is fine.  This PR is also removing the mention of caching, which is a no-go.

Adding a note on the page somewhere that sessions are a good candidate for persistent-redis would be a better approach.  (New PR, probably.)

andrerom:
> The build failure is that there is no redis-persistent-session.services.yaml file. There's no "-session" in the file name.

This is mentioned in the PR description ;) As in I don't know where these files are and need a hint on that. To be able to add one where the service is not called `data` is is the case in ´redis-persistent.services.yaml´ _(which again, I can't seem to find)_

Crell:
Those files are auto-generated by the build process.  Try `redis-persistent.services.yaml`.  I think that file does get generated.  (Or I have a copy of it in my local, anyway.)

chadwcarlson:
Thank you @andrerom! I resubmitted with the right filename in https://github.com/platformsh/platformsh-docs/pull/1613, so closing this one.

PR #1606, opened 2020-11-12
Less PHP heavy

RudyWeber:
Our documentation is a bit php-oriented.
This PR suggests changes to make it more language agnostic.

Also, there's some typos fixes, a bit of rephrasing and some light restructuring.

Follow the commits if you want to review these changes separately.


docs/src/configuration/app/build.md
```diff
@@ -138,45 +138,19 @@ What is "safe" to run in a `post_deploy` hook vs. in a `deploy` hook will vary b
 
 The `post_deploy` hook logs to its own file in addition to the activity log, `/var/log/post-deploy.log`.
 
-## How do I compile Sass files as part of a build?
```
Crell, line 141: This is a common task and request.  We can update it if needed, but shouldn't drop it.

docs/src/configuration/app/workers.md
```diff
@@ -18,10 +18,10 @@ workers:
         size: S
         commands:
             start: |
-                php worker.php
+                ./worker
```
Crell, line 21:
```suggestion
                ./worker.sh
```

I hate bare executable names. I never know what to do with them... :smile:

docs/src/configuration/app/workers.md
```diff
@@ -18,10 +18,10 @@ workers:
...
 ```
 
-That defines a single worker named `queue`, which will be a "small" container, and wil run the command `php worker.php` on startup.  If `worker.php` ever exits it will be automatically restarted.
+That defines a single worker named `queue`, which will be a "small" container, and wil run the command `./worker` on startup.  If `worker` ever exits it will be automatically restarted.
```
Crell, line 24:
```suggestion
That defines a single worker named `queue`, which will be a "small" container, and will run the command `./worker.sh` on startup.  If `worker` ever exits it will be automatically restarted.
```

docs/src/configuration/routes/https.md
```diff
@@ -73,11 +73,6 @@ https://{default}/:
 
 ### `min_version`
 
-{{< note >}}
-This directive was put into place when Platform.sh supported older versions of TLS for customers.
-Currently only TLS v1.2 is supported. Support for TLS v1.3 will be added in the near future.
-{{< /note >}}
```
Crell, line 79: Why remove?  It's worth telling people that you can't use <1.2 versions no matter what, especially for the security-conscious.

docs/src/configuration/routes/https.md
```diff
@@ -149,13 +144,12 @@ tls:
 
 If the Let's Encrypt certificate is due to expire in less than one month then it will be renewed automatically during a deployment.  That makes it feasible to set up regular auto-renewal of the Let's Encrypt certificate.  The caveat is that, like any deploy, there is a very brief downtime (a few seconds, usually) so it's best to do during off-hours.
 
-You will first need to install the CLI in your application container.  See the section on [API tokens](/development/cli/api-tokens.md) for instructions on how to do so.
```
Crell, line 152: This is a rather important instruction.  Keep it.

This whole section goes away soon, though, as automatic LE renewal is Coming Very Soon(tm).

docs/src/configuration/routes/https.md
```diff
@@ -174,7 +168,7 @@ The above cron task will run on the 1st and 15th of the month at 10 am (UTC), an
 It is very important to include the `--no-wait` flag.  If you do not, the cron process will block waiting on the deployment to finish, but the deployment will be blocked by the running cron task.  That will take your site offline until you log in and manually terminate the running cron task.  You want the `--no-wait` flag.  We're not joking.
 {{< /note >}}
 
-The certificate will not renew unless it has less than one month remaining; trying twice a month is sufficient to ensure a certificate is never less than 2 weeks from expiring.  As the redeploy does cause a momentary pause in service we recommend running during non-peak hours for your site.
+The certificate will not renew unless it has less than one month remaining; trying twice a month is sufficient to ensure a certificate is never less than 2 weeks from expiring.
```
Crell, line 171: This is an important caveat.  Please don't remove it.

Though again, this section goes away soon.

docs/src/configuration/services/_index.md
```diff
@@ -150,8 +143,6 @@ Close tunnels with: platform tunnel:close
 
 The `tunnel:open` command will connect all relationships defined in the `.platform.app.yaml` file to local ports, starting at 30000.  You can then connect to those ports on `localhost` using the program of your choice.
 
-In this example, we would connect to `localhost:30001`, database name `main`, with username `user` and an empty password.
```
Crell, line 153: Why remove?

docs/src/configuration/routes/https.md
RudyWeber: The `note` block right after says the same thing, with the same link.

docs/src/configuration/routes/https.md
RudyWeber: We already said this a few paragraphs before: L145 : `The caveat is that, like any deploy, there is a very brief downtime (a few seconds, usually) so it's best to do during off-hours.`

docs/src/configuration/services/_index.md
RudyWeber: I just moved it. See the addition right under.
When reading, I felt like it makes more sense to display the example result of the command and then explain what it means/allows, rather than saying "In this example [...]" when you haven't seen the example yet.

docs/src/development/email.md
```diff
@@ -45,6 +45,13 @@ When SMTP support is enabled the environment variable `PLATFORM_SMTP_HOST` will
 Changing the SMTP status will not take effect immediately.  You will need to issue a new *build*, not just a new deploy, for the changes to take effect. That means making a trivial change to some file in Git for the application.
 {{< /note >}}
 
+## Ports
```
Crell, line 48: Why reorder?

docs/src/development/variables.md
```diff
@@ -23,7 +23,7 @@ All of those may be simple strings or base64-encoded JSON-serialized values.  In
...
+Variables may be [set in code](/configuration/app/variables.md), using the `.platform.app.yaml` file.  These values of course will be the same across all environments and present in the Git repository, which makes them a poor fit for API keys and such.  This capability is mainly to define values that an application expects via an environment variable that should be consistent across all environments.  For example, some runtimes or frameworks use environment variables to condition their behaviours. You might want to set those as application-provided variables to ensure a consistent build across every environment. They may also be used to set [PHP configuration values](#php-specific-variables).
```
Crell, line 26:
```suggestion
Variables may be [set in code](/configuration/app/variables.md), using the `.platform.app.yaml` file.  These values of course will be the same across all environments and present in the Git repository, which makes them a poor fit for API keys and such.  This capability is mainly to define values that an application expects via an environment variable that should be consistent across all environments.  For example, some runtimes or frameworks use environment variables to configure their behavior, such as a `dev` vs `production` flag. You might want to set those as application-provided variables to ensure a consistent build across every environment. They may also be used to set [PHP configuration values](#php-specific-variables).
```

docs/src/overview/build-deploy.md
```diff
@@ -36,11 +36,11 @@ The live environment is composed of multiple containers&mdash;both for your appl
 
 Based on your application type the system will select one of our pre-built container images and run the following:
 
-1. First, any dependencies specified in the `.platform.app.yaml` file are installed.  Those include tools like Sass, Gulp, Drupal Console, or any others that you may need.  
+1. First, any dependencies specified in the `.platform.app.yaml` file are installed. Those include tools like Sass, Webpack, Drupal Console, or any others that you may need.
```
Crell, line 39: Is Gulp not cool anymore? I can't keep up with JS fashion... :smile:

docs/src/overview/pricing/_index.md
```diff
@@ -8,10 +8,6 @@ aliases:
   - "/overview/pricing.html"
 ---
 
-{{< description >}}
```
Crell, line 11: The description part needs to be kept, IIRC.  It doesn't get injected automatically.

Crell:
Lots of little comments, but overall looks solid.  Mostly, "don't remove as much stuff." :smile:

docs/src/configuration/app/build.md
RudyWeber: Mh, weird.
The reasons why I've removed this are:
- If I have a `package.json` file, I don't see why I would run `grunt` on its own during the build process and not use npm scripts. I'd rather avoid having my build steps scattered between `package.json` and `.platform.app.yaml`, especially here, since grunt is a required dependency in order to build (a part of) my app, but `npm install` wouldn't install it because it's been installed before? What's the benefit in doing that? My `package.json` file becomes dependant on `.platform.app.yaml`.
- Involving a specific task runner just to compile Sass files seems a bit overkill and adds unnecessary assumptions.

Also, [Ruby sass is deprecated](https://sass-lang.com/ruby-sass) and even though Grunt is still alive and used (kinda), modern applications mostly rely on Webpack / Rollup / Parcel.

This example feels a bit convoluted and outdated.

If it's a common question, and to stay in the idea of this example, maybe this would work:
```yaml
dependencies:
  nodejs:
    sass: "^3.4.21"

hooks:
  build: |
    npm install
    npm run build-css
```
?

docs/src/overview/pricing/_index.md
RudyWeber: Woops!

docs/src/development/email.md
RudyWeber: Structure-wise, I thought it made more sense to have all general information before the language specific ones rather than [general stuff] [PHP specific] [Java specific] [general stuff].

docs/src/overview/build-deploy.md
RudyWeber: Yeah modern webapps configurations don't really use grunt/gulp anymore. All major frameworks use and include documentation for Webpack.

docs/src/configuration/routes/https.md
RudyWeber: First, I thought about just updating this `note` regarding TLS v1.3, as it is now supported, but right after this block, we talk about automatic rejections of connections with version lower than `min_version` and also say `Legal values are `TLSv1.2` and `TLSv1.3`.`
So it should be clear that no version lower than 1.2 can be used. No?

docs/src/configuration/app/build.md
Crell: Wait, Sass was fully ported from Ruby to JS?  Why not to a proper compiled language?  I missed this memo...

This tutorial was written several years ago, so it's quite possible that the common practices have shifted.  Last I looked, Grunt and Gulp were the common JS task runners.  Does no one use those anymore in favor of `npm run`?  How about non-JS people who are going to be more than 6 months behind JS conventions? :smile: 

I'm fine with updating the tutorial if the conventions have shifted, but we should keep Sass compilation as an example.  (It was nice as a multi-lingual example, but if that's not the current state of the art, meh.)

docs/src/configuration/routes/https.md
Crell: Ah, fair point.  In that case, we should probably also update the example to use 1.3, so it's a non-default value.  And note that versions older than 1.2 will always be rejected, period.

docs/src/configuration/routes/https.md
Crell: ... I suspect that's because people never saw either one, and forgetting to setup the key was the most common thing people forgot. :smile:  Hence, say it twice.   But, fair point.

docs/src/configuration/routes/https.md
Crell: ACK.

docs/src/development/email.md
Crell: ACK

docs/src/configuration/app/build.md
RudyWeber: It's written in [Dart](https://dart.dev/), so you can either use the binary or the JS version.

`npm run X` will just run the npm script `X` declared in your `package.json` in `scripts`. You can do anything in there (it's just shell commands). So you can rely on a task runner / bundler or run a binary directly. Here, saying `npm run build-css` just assumes that we have a `build-css` npm script that does what it needs to do in order to build the css. Whether it uses a task runner or the sass binary directly is an implementation detail. (It's called "npm" scripts but Yarn would work just as well by doing `yarn build-css`.)

docs/src/configuration/routes/https.md
RudyWeber: Updated in d4f41c8
WDYT?

docs/src/configuration/app/build.md
RudyWeber: Adapted in 3eff045

docs/src/configuration/app/build.md
```diff
@@ -140,43 +140,36 @@ The `post_deploy` hook logs to its own file in addition to the activity log, `/v
 
 ## How do I compile Sass files as part of a build?
 
-As a good example of combining dependencies and hooks, you can compile your SASS files using Grunt.
+As a good example of combining dependencies and hooks, you can compile your SASS.
```
Crell, line 143:
```suggestion
As a good example of combining dependencies and hooks, you can compile your Sass files.
```

RudyWeber: Huhu, thank you.

docs/src/configuration/routes/https.md
```diff
@@ -73,19 +73,15 @@ https://{default}/:
...
 
 ```yaml
 tls:
-    min_version: TLSv1.2
+    min_version: TLSv1.3
 ```
 
-The above configuration will result in requests using older TLS versions to be rejected.  Legal values are `TLSv1.2` and `TLSv1.3`.
+The above configuration will result in requests using older TLS versions to be rejected. Legal values are `TLSv1.2` and `TLSv1.3`.
+Any version older than 1.2 will be rejected.
```
Crell, line 84:
```suggestion
TLS versions older than 1.2 are not supported by Platform.sh and will be rejected regardless of the setting here.
```

Crell (approved):

PR #1625, opened 2020-12-01
Document error message for source operations.

GuGuss:
Capture d’écran 2020-12-01 à 14 35 25


chadwcarlson:
Resubmitted in https://github.com/platformsh/platformsh-docs/pull/1650 with some cleanup since this section was recently removed. Thanks @GuGuss

PR #1626, opened 2020-12-01
Update registry.json

trolologuy:
Updating doc to PHP 8.0


Crell (approved):

trolologuy (commented):
need to verify the information about PHP 8.0's availability on dedicated. I will add a comment once I confirmed that information.

trolologuy:
> need to verify the information about PHP 8.0's availability on dedicated. I will add a comment once I confirmed that information.

Didn't get the reply about PHP 8.0's availability on dedicated yet. Rollback'ed to the grid-only update.

PR #1644, opened 2020-12-10
Fix 10.5 mysql lack of support.

chadwcarlson:
`mysql` 10.5 not supported. Uses newest registry parser to separate into distinct `mysql` registry object.

### DIFF
--- a/docs/data/registry.json
+++ b/docs/data/registry.json
@@ -350,6 +350,40 @@
       ]
     }
   },
+  "mysql": {
+    "description": "A manufacture-based container for MariaDB",
+    "repo_name": "mariadb",
+    "disk": true,
+    "docs": {
+      "relationship_name": "database",
+      "service_name": "db",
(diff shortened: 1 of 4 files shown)


Crell:
Let's go ahead and remove the `mysql` example block from further down the page.  It's not needed anymore.

docs/data/registry.json
```diff
@@ -350,6 +350,40 @@
       ]
     }
   },
+  "mysql": {
+    "description": "A manufacture-based container for MariaDB",
```
robertDouglass, line 354: Who is the audience of this? Isn't "manufacture-based" an internal detail?

chadwcarlson: It's not publicly shown anywhere other than this file, but ya it's from our repo description. An artifact of trying to generate this file programmatically from GitLab I reckon - fine to remove if we want to.

robertDouglass (approved):

PR #1645, opened 2020-12-10
Split build.sh.

chadwcarlson:
Splits `build.sh` into
- `build_examples.sh`: fetches example files from templates, and builds configuration snippets using the registry-parser.
- `build_docs.sh`: all things specific to hugo. 

While I was in there:
- all files in `registry` were ignored previously, which should have caused problems with the project wizard since example yamls weren't being served. Don't know, fixing it here. 
- Because of the change above ^, Hugo verifies pages. They were breaking for language list pages (java, nodejs, php), so those links in the registry were also updated.


robertDouglass (approved):
I have absolutely no way of testing this and don't eyeball any problems.

PR #1663, opened 2020-12-24
Update redirects.md

trolologuy:
Since the upstream is missing the key (here `app` for e.g):
```
"https://www.{default}/":
  type: upstream
  upstream: "app:http"
  redirects:
```
the snippets will not work out of the box.

That should be clearer, hence adding the comment to help clarify that.


Crell:
Agreed with the idea, but these are indented wrong.  Let's just use `# ...` (not an array, as that could be misleading) and align it with the `type` and `redirect` lines.

trolologuy:
Should be clearer now, let me know what you think :)

Crell (approved):

PR #1715, opened 2021-02-03
rework new relic doc

RudyWeber:
- add Node.js New Relic doc
- split the doc so there's one page per language configuration
- adapted the existing doc to use environment level variables so people know they can control which environments are monitored (see [this question](https://community.platform.sh/t/how-to-conditionally-disable-new-relic-based-on-git-branch/718))


docs/src/integrations/profiling/new-relic/_index.md
```diff
@@ -0,0 +1,13 @@
...
+  Platform.sh supports [New Relic APM](https://newrelic.com/products/application-monitoring) for profiling your applications.
+---
+
+{{< description >}}
+
+## On a Dedicated cluster
+
+Open a support ticket and let us know what your license key is.  Our support team will install it and let you know when it is complete.
+
+## On a Grid plan, configure your project
```
Crell, line 13: This seems to just stop in the middle here?

RudyWeber: It then displays the list of the supported languages but hugo does that automatically. Looks weird, I agree. 🤷

Crell: Oh ew. I really don't like the default giant-button design there. :smile: It could still use some additional intro text, like "the setup varies slightly by language, select the appropriate language from the list below" or something.

Crell:
Another point: This would be the deepest page in the hierarchy.  We've to date managed to avoid going more than 2 levels deep.  As the docs grow we may not be able to avoid that much longer, but it's something we've tried to do.  I don't have a counter-proposal, though, so mostly I'm flagging it so we know we're crossing a rubicon here.

Crell (approved):

PR #1725, opened 2021-02-18
Update nvm.md

rhubinak:
The $PLATFORM_APP_DIR/.nvm directory is not always created (for example with NVM_VERSION: v0.34.0), therefore creating it explicitly.

### DIFF
--- a/docs/src/languages/nodejs/nvm.md
+++ b/docs/src/languages/nodejs/nvm.md
@@ -22,6 +22,7 @@ variables:
 hooks:
     build: |
         unset NPM_CONFIG_PREFIX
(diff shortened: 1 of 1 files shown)


RudyWeber:
Thanks for your suggestion @rhubinak!
However, this is a version-specific behaviour. It has been fixed in nvm 0.35.0 (see [this PR](https://github.com/nvm-sh/nvm/pull/1986)).
So this change isn't useful for nvm ^0.35.0 and our documentation currently suggests version 0.36.0.

If you need a specific version and this version has a specific behaviour, then you will have to change the configuration accordingly.

PR #1734, opened 2021-02-27
Add missing step to default_branch change instructions

kay-o:
without this step I get the following result

```% platform environment:delete master --no-delete-branch
The environment master has children and therefore can't be deleted.
Please delete the environment's children first.```

With the step I get the expected behavior.


chadwcarlson (approved):

kay-o:
tested just now and can confirm that section of docs is complete and works as expected

chadwcarlson:
Thanks @kay-o !

PR #1736, opened 2021-03-01
Remove another mention of certificate renewal cron

pjcdawkins:
No longer needed since https://platform.sh/blog/2021/https-and-tls-certificates-always-served-fresh/


chadwcarlson:
Resubmitted in https://github.com/platformsh/platformsh-docs/pull/1741 since the integration hiccupped on this one.

PR #1749, opened 2021-03-08
Add info about redis module depend igbinary

spk:
Add comment to ensure people add igbinary when adding php redis module

### DIFF
--- a/docs/src/languages/php/extensions.md
+++ b/docs/src/languages/php/extensions.md
@@ -11,7 +11,7 @@ runtime:
     extensions:
         - raphf
         - http
-        - igbinary
+        - igbinary  # redis module depend on igbinary
         - redis
     disabled_extensions:
         - sqlite3


chadwcarlson:
Thanks @spk - I wasn't aware that this was a requirement for the `redis` extension until now - its not enabled in our [Drupal 9 template](https://github.com/platformsh-templates/drupal9/blob/e59d8b48c22e3a6be5e59374656c098833f3409d/.platform.app.yaml#L19) for example. 

What's the error you get here that I can test on that template?

spk:
> Thanks @spk - I wasn't aware that this was a requirement for the `redis` extension until now - its not enabled in our [Drupal 9 template](https://github.com/platformsh-templates/drupal9/blob/e59d8b48c22e3a6be5e59374656c098833f3409d/.platform.app.yaml#L19) for example.
> 
> What's the error you get here that I can test on that template?

On the build log we get a warning and the module is not working for example:

```
//build log      
        Found a `composer.json`, installing dependencies.
        W: PHP Warning:  Cannot load module "redis" because required module "igbinary" is not loaded in Unknown on line 0
        W: Loading composer repositories with package information
```

PR #1753, opened 2021-03-10
we've adjusted our data retention standards 

rinchen:
we've adjusted our data retention standards various legal and standard requirements


rinchen:
I've also removed the payment processing entry because we do not do payment processing.

damz (approved):

PR #1770, opened 2021-03-29
Remove duplicated extension

fabpot:
The  `php` extension is listed twice in the file, not sure if one of the names is wrong or if we should merge the versions. I let you investigate further.


chadwcarlson:
Hi @fabpot - verified that we support 5.4, 5.5, 7.3, 7.4, and 8.0. Closing this in favor of https://github.com/platformsh/platformsh-docs/pull/1772. Thank you!

PR #1773, opened 2021-03-29
Platform.sh WAF

chadwcarlson:


docs/src/security/waf.md
```diff
@@ -0,0 +1,145 @@
...
+
+Depending on the characters added to the response header (such as line feeds and carriage returns), an attacker can create and return additional responses to legitimate user requests, hence the name response *splitting*. 
+
+{{< note title="Protection rules">}}
+The WAF monitors for requests that include the carriage return (CR; `%0d`) and line feed (LF; `%0a`) characters. The presence of these characters indicate an attempt to inject data in the response header and potentially force intermediate proxy servers to treat the message as two separate responses. 
+{{< /note >}}
+
+### HTTP Splitting
+
+[Need some help with intro description here. Seems like much of the above, only specific to filename POSTs. Or Rewrites, but that seems specific to Apache/PHP? Not really sure.]
```
chadwcarlson, line 53: some description if we have one.

docs/src/security/waf.md
```diff
@@ -0,0 +1,145 @@
...
+
+- **Missing/Empty Host Headers**
+
+    The Gateway proxy implements route mapping based on host names. Because of this, requests are not relayed that do either do not include the host header, or if that header exists but is empty. 
+
+- **File upload limit**
+
+    File upload limits are enforced by the Gateway proxy, but are configured by the application's configuration in `.platform.app.yaml` using the `max_requst_size` attribute in [`web.locations`](/configuration/app/web.md#locations).
+
+    [Do we set some default here if not set? Some default limit on multipart/form-data related to https://docs.platform.sh/development/troubleshoot.html#large-file-upload-failing-10mb-limit?]
```
chadwcarlson, line 83: clarification I thought of if we know the answer here

docs/src/security/waf.md
```diff
@@ -0,0 +1,145 @@
...
+#### [PRODSECBUG-2403](https://magento.com/security/patches/magento-2.3.3-and-2.2.10-security-update)
+
+| Type:  | Remote Code Execution |
+|-----|-----|
+| **Details:** | An unauthenticated user can insert a malicious payload through Page Builder template methods.  |
+| **Rule(s):** | Suspicious requests are filtered based on the request path. |
+
+#### [PRODSECBUG-2432](https://magento.com/security/security-update-potential-vulnerability-magento-admin-url-location)
+
+| Type:  | Information Disclousure  |
```
chadwcarlson, line 142:
```suggestion
| Type:  | Information Disclosure  |
```

docs/src/security/waf.md
chadwcarlson: (HTTP splitting) what are we protecting against when we monitor for \r and \n in REQUEST_FILENAME, and where is this variable?

docs/src/security/waf.md
chadwcarlson: (File upload limit) users can set max_request_size to configure this, but what is it if unset (10MB like https://docs.platform.sh/development/troubleshoot.html#large-file-upload-failing-10mb-limit?)

docs/src/security/waf.md
chadwcarlson: removed, not relevant for us.

docs/src/security/waf.md
```diff
@@ -0,0 +1,139 @@
...
+tier:
+  - Enterprise
+  - Elite
+---
+
+{{< description >}}
+
+## HTTP protocol attacks
+
+Platform.sh's WAF implements a number of request filtering rules for common security vulnerabilities on the HTTP protocol, which are implemented either on the Gateway proxy or the nginx Router service.
```
chadwcarlson, line 16:
```suggestion
Platform.sh's WAF implements a number of request filtering rules for common security vulnerabilities on the HTTP protocol. 
```

docs/src/security/waf.md
```diff
@@ -0,0 +1,139 @@
...
+
+{{< description >}}
+
+## HTTP protocol attacks
+
+Platform.sh's WAF implements a number of request filtering rules for common security vulnerabilities on the HTTP protocol, which are implemented either on the Gateway proxy or the nginx Router service. 
+
+### Request smuggling
+
+The [HTTP specification](https://tools.ietf.org/html/rfc2616) allows for two ways to define where a request ends within its header: Content-Length and Transfer-Encoding. Both can be used, but the specification additionally outlines that if both headers are present in a single request, Transfer-Encoding should be used over Content-Length. HTTP request smuggling occurs when
```
RudyWeber, line 20:
```suggestion
The [HTTP specification](https://tools.ietf.org/html/rfc2616) allows for two ways to define where a request ends within its header: `Content-Length` and `Transfer-Encoding`. Both can be used, but the specification additionally outlines that if both headers are present in a single request, `Transfer-Encoding` should be used over `Content-Length`. HTTP request smuggling occurs when
```

Just for consistency with L61.

RudyWeber (approved):

PR #1777, opened 2021-04-02
Better handle Meilisearch upgrades.

chadwcarlson:
Background:

- Previously, all index data (saved in the mount `data.ms`) was deleted during the `cleanup` function of the `search` app's post deploy hook. The main reason for this is that if Meilisearch was upgraded (`MEILISEARCH_VERSION` var in `.platform.app.yaml`) the database would be incompatible with the new version, and deployment would fail. 
- This was removed here https://github.com/platformsh/platformsh-docs/pull/1723, because if anything went wrong and Meili needed to restar, there would be no data until another redeploy.
- Upgrading Meilisearch makes those old database files incompatible with the new version, and a failed deploy. This happened on the most recent PR #1776, and required sshing in, deleting the contents of `data.ms`, and then a redeploy. 

Update:
- start command calls new file, `start.sh` which includes previous start command
- new function `deleteIndexPreMeilisearchUpgrade` added in that file. compares downloaded version (`MEILISEARCH_VERSION` env var) to "current db" in the `data.ms` mount (`data.ms/VERSION` file.)

### DIFF
--- a/search/.platform.app.yaml
+++ b/search/.platform.app.yaml
@@ -22,7 +22,6 @@ hooks:
 web:
     commands:
       # Run the Meilisearch server
-    #   start: "./meilisearch --http-addr localhost:${PORT} --master-key $PLATFORM_PROJECT_ENTROPY"
       start: !include 
             type: string
             path: start.sh
--- a/search/.platform.app.yaml
+++ b/search/.platform.app.yaml
@@ -6,7 +6,7 @@ variables:
     env:
         # Do not update meilisearch version without having a plan to
         # delete the .mdb files in data.ms/ on the mounts as well.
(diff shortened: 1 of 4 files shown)


chadwcarlson:
Check downgrading to 0.19.0

```
  > Downloaded version: 0.19.0
  > Current database version: 0.20.0
  > Meilisearch was upgraded. Deleting old index before start.
```

chadwcarlson:
Restore 0.20.0

```
  > Downloaded version: 0.20.0
  > Current database version: 0.19.0
  > Meilisearch was upgraded. Deleting old index before start.
```

RudyWeber (approved):

PR #1791, opened 2021-04-30
Build environment variables.

chadwcarlson:
DEVREL-998

Environment variables can be given the same `--visible-build` and `--visible-runtime` flags as project variables.


docs/src/development/variables.md
```diff
@@ -53,15 +53,17 @@ Project variables may also be marked `--sensitive true`.  That flag will mark th
...
 
-There are two additional flags available on environment variables: `--inheritable` and `--sensitive`.
+There are a few additional flags available on environment variables:
 
 * Setting `--inheritable false` will cause the variable to not be inherited by child environments.  That is useful for setting production-only values on the `master` branch, and allowing all other environments to use a project-level variable of the same name.
 * Setting `--sensitive true` flag will mark the variable to not be readable through the management console once it is set.  That makes it somewhat more private as requests through the Platform.sh CLI will not be able to view the variable.  However, it will still be readable from within the application container like any other variable.
+* The `--visible-runtime` (default: `true`) flag defines that variable as accessible during runtime. 
+* The `--visible-build` (default: `true`) flag defines that variable as accessible at build time. Build environment variables change the application's build configuration ID, and triggers a rebuild of the application the same way that a commit would.
```
maria-webb, line 66:
```suggestion
* The `--visible-build` (default: `true`) flag defines that variable as accessible at build time. Build visible environment variables change the application's build configuration ID, and triggers a rebuild of the application in the same way that a commit would. 
```

docs/src/development/variables.md
```diff
@@ -73,25 +75,54 @@ If you omit the variable `--value` from the command line as above, you will be p
...
+
+```bash
+$ platform variable:create -l environment -e master --prefix env: --name NODE_ENV --value production --visible_build true --inheritable false
+$ platform variable:create -l environment -e staging --prefix env: --name NODE_ENV --value development --visible_build true --inheritable true
+```
+
+With these two commands, while `NODE_ENV` will be `production` on the default branch, it will be `development` on staging as well as each of its child environments. Note that build environment variables change the application's build configuration ID - value updates will trigger a rebuild of the application the same way that a commit would.
```
maria-webb, line 87:
```suggestion
With these two commands, while `NODE_ENV` will be `production` on the default branch, it will be `development` on staging as well as each of its child environments. Note that build visible environment variables change the application's build configuration ID, and triggers a rebuild of the application in the same way that a commit would.. 
```

docs/src/development/variables.md
```diff
@@ -73,25 +75,54 @@ If you omit the variable `--value` from the command line as above, you will be p
...
+* **PLATFORM_APPLICATION**: A base64-encoded JSON object that describes the application. It maps certain attributes from the `.platform.app.yaml` that you have in Git and it has a few subkeys.
+
+`PLATFORM_APPLICATION` is a special case to keep in mind in how it differs between the build and runtime container. Every environment's build is associated with a configuration ID that uniquely identifies it. It's this ID which allows us to reuse builds on merges. The ID itself is a product of your application code, as well as its configuration for Platform.sh in `.platform.app.yaml`. 
+
+Not every attribute in `.platform.app.yaml` is relevant to builds - that is, not every attribute is accessible to the build container in `PLATFORM_APPLICATION`, nor will commits which edit those attributes result in a full rebuild of the application when pushed. Because of this, not all attributes defined in your `.platform.app.yaml` file will be accessible at build time from `PLATFORM_APPLICATION`, only those actually relevant to builds. Some of those attributes that will **not** be available in `PLATFORM_APPLICATION` during buids include:
```
maria-webb, line 113:
```suggestion
Not every attribute in `.platform.app.yaml` is relevant to builds - that is, not every attribute is accessible to the build container in `PLATFORM_APPLICATION`, nor will commits which edit those attributes result in a full rebuild of the application when pushed. Because of this, not all attributes defined in your `.platform.app.yaml` file will be accessible at build time from `PLATFORM_APPLICATION`, only those actually relevant to builds. Some of those attributes that will **not** be available in `PLATFORM_APPLICATION` during builds include:
```

maria-webb (changes requested):
Some minor fixes

docs/src/development/variables.md
chadwcarlson: @maria-webb this was redundant, but I ended up rewriting/keeping the second mention instead.

maria-webb (approved):

PR #1792, opened 2021-05-04
Update grid.md

rhubinak:
It's possible to set cron to run every minute on Elite plans on Grid.


MariadeAnton (approved):
Thanks for removing internal naming @rhubinak !

PR #1817, opened 2021-05-25
Delete periods

ErikaBustamante:


zendoodles:
@MariadeAnton Do we have a style guide that covers this? If I remember correctly, the AP style guide prefers capitalization and periods for bulleted or numbered lists. It looks like there are examples of proper sentences, fragments with periods, and fragments without in the rest of the docs though.

rinchen (approved):

PR #1818, opened 2021-05-25
Misc. Cleanup

zendoodles:
- Removes `docs/old` directory, because if we actually need it later, git remembers.
- Adds inline comments for services so I know where to go to update versions next time.
- Commit an odd thing in the changelog that looks like nothing is changed, but apparently it is? Likely a line-ending.


MariadeAnton (approved):

MariadeAnton:
Thanks for the clean up and useful comments in Markdown files @zendoodles! The changelog one does seem like a trailing whitespace after the dot.

chadwcarlson (approved):

PR #1828, opened 2021-06-09
Update language to match BAA appendix

ErikaBustamante:


rinchen (approved):
Revised wording to match BAA appendix

PR #1834, opened 2021-06-24
languages/php: Improve YAML quote consistency

damz:
Try to apply the following standard:

* Do not quote map keys when they are property names (i.e. they are keys of a struct, not keys of a map);
* Always quote map values unless they are a closed-list enumeration (like `upstream.socket_family: tcp`)


royallthefourth (approved):
This change removes unnecessary quotes from YAML keys and adds them to the values in alignment with damz's style recommendations

PR #1847, opened 2021-07-26
Rewrite various unnecessary uses of the word "leverage"

sruggier:


MariadeAnton (approved):
These changes simplify the phrases in which the word leverage was used. 🎉  Thank you for these, @sruggier

PR #1888, opened 2021-09-07
Ruby and Rails improvements

spk:
Improve docs for latest rails version with assets pipeline and storage

ref https://github.com/platformsh/template-builder/pull/390


CollierCZ (approved):
Thanks for the improvements, @spk! I made some small styling changes, but nothing major. Glad we could see this incorporated. :bow:

PR #1892, opened 2021-09-10
docs: add brief comparison of CLI and Git commands

CollierCZ:


docs/src/development/cli/_index.md
```diff
@@ -76,13 +76,21 @@ Options:
  --shell (-s)          Launch the shell
 ```
 
+### Comparison with Git commands
+
+Some CLI commands (especially many within the `environment` namespace) have some overlap with Git commands. Generally, they offer more options than the Git commands alone. For example, `platform push` offers options such as `--activate` (to active an environment before pushing) and `--no-wait` (so you can continue working without waiting for the push to complete).
+
+For all of them, you don't need to configure a Git remote. It is enough to have a project ID.
+
+An example of how this affects commands, when you run `platform merge`, it doesn't affect your local codebase. You don't even need the code locally. The code is merged between environments remotely.
```
chadwcarlson, line 85:
```suggestion
An example of how this affects commands is that when you run `platform merge`, it doesn't affect your local codebase. You don't even need the code locally. The code is only merged between environments remotely.
```

docs/src/development/cli/_index.md
```diff
@@ -38,12 +38,12 @@ When you are logged in, a list of your projects appears, along with some tips fo
 **Your command-line tools are now ready to use with Platform.sh.**
 
 {{< note >}}
-Please consult [the full documentation on CLI Authentication](https://github.com/platformsh/platformsh-cli#authentication) on the public CLI Github repository for further details.
+Please consult [the full documentation on CLI Authentication](https://github.com/platformsh/platformsh-cli#authentication) on the public CLI GitHub repository for further details.
```
maria-at-platform, line 41: I would drop the 'please'. It's just a personal preference. Since we tend to use a direct style, we would have to use 'please' quite a bit for the sake of consistency, so I tend to drop it.  : ) 

Learn more about the public CLI Github repository [here](https://github.com/platformsh/platformsh-cli#authentication).

maria-at-platform (approved):

docs/src/development/cli/_index.md
CollierCZ: I agree about dropping `please`. I only changed the line because it was caught by spellcheck, so I didn't even really read it.

I don't agree with the link in `here`, for the reasons noted in [the style PR](https://github.com/platformsh/platformsh-docs/pull/1902/files#diff-cd95459fff6e7161b4478a42b001be388b9e7cffa0902ae17e0985be1eb61ee8R71), but I can definitely improve the sentence. :+1:

PR #1895, opened 2021-09-13
fix: add clarification on ephemeral cache

CollierCZ:


docs/src/configuration/app/build.md
```diff
@@ -101,9 +101,11 @@ There are no constraints on what can be downloaded during your build hook except
...
+This hook should be used when something needs to run once for all instances of an app when deploying new code. It is not run when a host is restarted (such as during region maintenance), so anything that needs to run each time an app starts (regardless of whether there's new code) should go in the `start` key in [your web configuration](/configuration/app/web.md#commands).
+
+Be aware: The deploy hook blocks the site accepting new requests. If your deploy hook is only a few seconds then incoming requests in that time are paused and will continue when the hook completes, effectively appearing as the site just took a few extra seconds to respond. If it takes too long, however, requests cannot be held and will appear as dropped connections. Only run tasks in your deploy hook that have to be run exclusively, such as database schema updates or some types of cache clear (those where the code must match what's on the disk). A post-deploy task that can safely run concurrently with new incoming requests should be run as a `post_deploy` hook instead.
```
maria-at-platform, line 108: Again, great work. I'm thinking we might want to find a better way to display/convey critical information. I'll definitely note this as an example to discuss for any upcoming work related to a new style guide.💯 I'm keen to know your thoughts on the usage of 'note:', be aware, and so on. Do you have any personal preferences?

maria-at-platform (approved):

docs/src/configuration/app/build.md
CollierCZ: Interestingly, I was writing about this in the [style PR](https://github.com/platformsh/platformsh-docs/pull/1902/files#diff-cd95459fff6e7161b4478a42b001be388b9e7cffa0902ae17e0985be1eb61ee8R156) while you were making this comment :slightly_smiling_face: 

I think the notes (or callouts, admonitions, whatever you want to call them) work well when they're short. For long information, or things that are more a part of the flow, I think inline things work better.

We could discuss it more in the style PR if you have more ideas.

PR #1905, opened 2021-09-23
fix: workaround for broken examples

CollierCZ:
Builds currently fail because examples aren't returned. This PR should allow the builds to complete,
by adding a timeout and using the current examples as fallbacks if the examples can't be fetched.
They should be overwritten by anything new.


chadwcarlson:
Closes #1906

chadwcarlson (approved):

PR #1907, opened 2021-09-24
Multidb pg 10 11 12

spk:
Its going to be available soon :tm:


CollierCZ:
Hi @spk. Do you have insight into when this will be available? Just to know when we should think about merging it.

spk:
Hi @CollierCZ not sure when maybe I should keep it as draft until then ?

spk:
Checked all 10 11 12 has been released with multidb

CollierCZ (approved):

PR #1910, opened 2021-09-29
feat: add how-to template

CollierCZ:
The other option would be to have separate template files and explanation files, like the Good Docs Project does: https://github.com/thegooddocsproject/templates/tree/dev/how-to I prefer having the instructions in the file and demonstrating as much as possible through examples rather than explanations. But I'm open to other ways.

### DIFF
--- a/docs/templates/how-to.md
+++ b/docs/templates/how-to.md
@@ -17,39 +17,44 @@ How to use
 -->
 
 Introduce the topic in 1 or 2 sentences.
-Include reasons why a person might want to do it
+Include reasons why the reader might want to do it
 (the most common use cases).
 Also include what the goal of the task is.
 
 ## Before you begin
 
 Include here any terms the reader might need
-(either defined here or linked to a definition).
-Also any prerequisites they might need to install/know,
+that they might not be familiar with
+(like terms specific to Platform.sh).
+Either define them here or link to a definition elsewhere in the docs.
+
+Also include any prerequisites they might need to install/know,
 such as having the CLI installed or having created a project.
+
 For tasks like installing the CLI,
 link to instructions rather than including them here.
 
-## 1. Do this first
+## 1. Do this step first
 
-This starts the list of steps readers must do in order.
-Order them with numbers.
+This starts the list of steps readers must do in order
+(each step is a separate section in the document).
+Order the steps with numbers.
 The titles should be bare infinitives: https://developers.google.com/style/headings?hl=en
 
 Introduce the purpose of this step very briefly.
 
-Then use an ordered list for the actions to take:
+Then use an ordered list for the actions to take to complete the step:
 
-1. Do this first.
-1. Do this next.
-1. Do this third.
-1. Do this fourth.
+1. Do this action first.
+1. Do this action next.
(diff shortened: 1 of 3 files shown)


maria-at-platform:
> The other option would be to have separate template files and explanation files, like the Good Docs Project does: https://github.com/thegooddocsproject/templates/tree/dev/how-to I prefer having the instructions in the file and demonstrating as much as possible through examples rather than explanations. But I'm open to other ways.

I personally prefer having both in the same file. I eases the learning process and it's cleaner anyway.

docs/templates/how-to.md
```diff
@@ -0,0 +1,117 @@
...
+  https://diataxis.fr/how-to-guides/ 
+
+How to use
+  1. Copy this template into the right directory in /src/docs/.
+  2. Rename it to match the title.
+  3. Replace the following content with your own.
+-->
+
+Introduce the topic in 1 or 2 sentences.
+Include reasons why a person might want to do it
```
maria-at-platform, line 20: I'd change person for user.

docs/templates/how-to.md
```diff
@@ -0,0 +1,117 @@
...
+(the most common use cases).
+Also include what the goal of the task is.
+
+## Before you begin
+
+Include here any terms the reader might need
```
maria-at-platform, line 26: I feel this could lead some contributors to mistake. How will they know which terms to define? what's the standard for what's obvious and what's not? We might want to wait until we have a glossary to prompt contributors to define terms. Just an idea though.

docs/templates/how-to.md
```diff
@@ -0,0 +1,117 @@
...
+1. Do this third.
+1. Do this fourth.
+
+Include a short example of what the expected outcome of this step is,
+such as a code example, a description of what the user can do now,
+or (if absolutely necessary) a screenshot.
+
+## 2. Do this second
+
+Split into separate steps when there are more than 5--7 actions in a step.
```
maria-at-platform, line 54: I think external contributors might not understand at first what we mean by split into separate steps. It could be creating 2 different tasks, using substeps or something else. If they don't have experience creating how to content, they might get confused.  I've taken the liberty to add an alternative explanation. 

"Before you start writing, think about the number of steps a task might have. If it has more than 5-7 steps, you might want to consider breaking the tasks into two separate tasks or using substeps. 
Also, split the steps when each set has a different outcome"

maria-at-platform (approved):
Thanks for putting together the page, it looks great.

docs/templates/how-to.md
CollierCZ: I'm not sure there's an absolute standard other than things readers might need. I added a little more for explanation, but that does make it longer. :shrug:

docs/templates/how-to.md
CollierCZ: I totally understand the potential for confusion. I was intentionally trying to avoid terms like "substeps" because I don't think they really add clarity. Here, "steps" are the sections and "actions" are the items in the list. I can make that more clear, especially since I didn't use that in step 1 :facepalm: .

PR #1917, opened 2021-10-01
remove MASTER_USE_GTID from start command example

georanma:
With the setup provide by this guide, the "slave_pos" is an empty string when starting the replication database. This causes an error in the reading of the bin logs, as it can not find the '' location in the log. Removing this line from the slave definition allows the database to start and copy as expected.


CollierCZ:
Thanks so much for your contribution!

CollierCZ (approved):

georanma:
Not a problem. Hopefully, it saves someone else some time.

PR #1924, opened 2021-10-06
improve: use cache for nvm

CollierCZ:
Improve instructions for NVM.

The steps should be explained by the code comments. If you think it would help to break it into separate steps, I can do that too.

[DEVREL-1012](﻿https://platformsh.atlassian.net/browse/DEVREL-1012)

### DIFF
--- a/docs/src/languages/nodejs/_index.md
+++ b/docs/src/languages/nodejs/_index.md
@@ -13,7 +13,7 @@ layout: single
 |----------------------------------|---------------|
 |  {{< image-versions image="nodejs" status="supported" environment="grid" >}} | {{< image-versions image="nodejs" status="supported" environment="dedicated" >}} |
 
-If you need other versions or a version in a non-Node container,
+If you need other versions or a version in a non-Node.js container,
 [use a version manager like `nvm`](/languages/nodejs/nvm.md).
 
 ## Deprecated versions


docs/src/languages/nodejs/nvm.md
```diff
@@ -1,41 +1,62 @@
 ---
-title: "Alternative Node.js install"
+title: "Manage Node.js versions"
 weight: 1
 toc: false
+description: See how to manage different Node.js versions in your Platform.sh containers.
 ---
 
-## How to use NVM to run different versions of Node.js
+You may have instances where you need to use a specific version of Node.js that isn't available as a Platform.sh container.
```
chadwcarlson, line 8:
```suggestion
You may have instances where you need to use a specific version of Node.js that isn't available in an application container.
```

chadwcarlson (dismissed):

docs/src/languages/nodejs/_index.md
```diff
@@ -13,27 +13,32 @@ layout: single
 |----------------------------------|---------------|
 |  {{< image-versions image="nodejs" status="supported" environment="grid" >}} | {{< image-versions image="nodejs" status="supported" environment="dedicated" >}} |
 
-If you need other versions, take a look at our [options for installing them with NVM](/languages/nodejs/nvm.md).
+If you need other versions or a version in a non-Node container,
```
MariadeAnton, line 16:
```suggestion
If you need other versions or a version in a non-Node.js container,
```

PR #1926, opened 2021-10-07
Postgres rm comment

spk:
Remove comment from mysql option not present on postgres (PDO::MYSQL_ATTR_USE_BUFFERED_QUERY) cf https://github.com/platformsh/platformsh-docs/blob/7b3d6e09774e19fdb967d097e78dd97a600b4551/docs/static/files/fetch/examples/php/mysql#L23

Remove trailing on mysql file

### DIFF
--- a/docs/static/files/fetch/examples/php/mysql
+++ b/docs/static/files/fetch/examples/php/mysql
@@ -36,9 +36,9 @@ try {
     $conn->query($sql);
 
     // Insert data.
-    $sql = "INSERT INTO People (name, city) VALUES 
-        ('Neil Armstrong', 'Moon'), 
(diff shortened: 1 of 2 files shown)


CollierCZ:
@spk , if you want the comment to disappear in the docs, you have to remove it from the [language examples repository](https://github.com/platformsh/language-examples/blob/master/php/examples/MySQL.php#L22). The examples in this repository are just backups in case the URL is unavailable for some reason (so the build can still continue). But they're overridden by the examples from the other repository.

CollierCZ:
Sorry, I was on the wrong branch. :facepalm: I'll merge the changes here, too.

CollierCZ (approved):

PR #1928, opened 2021-10-07
Google search verification

chadwcarlson:
Adds google search verification tag at request on Jon P.

### DIFF
--- a/docs/themes/avocadocs/layouts/partials/head/meta-tags.html
+++ b/docs/themes/avocadocs/layouts/partials/head/meta-tags.html
@@ -1,6 +1,6 @@
 <!--  Meta tags -->
 
-# Site verification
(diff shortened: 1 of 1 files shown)


docs/themes/avocadocs/layouts/partials/head/meta-tags.html
```diff
@@ -1,5 +1,8 @@
 <!--  Meta tags -->
 
+# Site verification
```
CollierCZ, line 3:
```suggestion
<!-- Site verification -->
```

I don't think it's necessary in the final build. It also makes the HTML invalid.

CollierCZ (approved):
Great. It took me a while to find before, because the tag was appearing in `<body>` rather than `<head>`, but now it seems just fine :+1:

PR #1938, opened 2021-10-19
Add AU-2 Region

tristanblattmansh:
Add AU-2 region to IP list


CollierCZ:
Closing in favor of #1919 (which already has some additional styling improvements).

PR #1957, opened 2021-10-27
WIP: Multi-App with same source code

OriPekelman:
This is one of the root use-cases for which we built this feature and there are some gotchas here.


cs278:
Just stumbled across this, I opened a ticket (149891) about this very problem a while back which is still open, hadn't thought to try `/` so will give it a go and see if it fixes our problem.

trolologuy:
Since it's also unclear in which order multi-apps are built (seems to be alphabetical), whilst addressing this one, we might add a note about that as well.

CollierCZ:
An example of multiple apps from the same source is already present at: https://docs.platform.sh/create-apps/multi-app.html#unified-app-configuration

PR #1967, opened 2021-11-03
:memo: Storage downsize is impossible only in older regions

rhubinak:


CollierCZ (approved):
Thanks for the fix, @rhubinak! (And the use of gitmoji :heart_eyes: )

PR #1987, opened 2021-11-11
 📝 Update sentence for configuration block

SabriPlatform:
## Why

Closes #{ISSUE_NUMBER}

## What's changed

I just changed the sentence from either to neither/nor

### DIFF
--- a/docs/src/configuration/services/mysql/_index.md
+++ b/docs/src/configuration/services/mysql/_index.md
@@ -204,7 +204,7 @@ configuration:
                 main: admin
 ```
 
-If either schemas or endpoints are defined, then no default will be applied and you must specify the full configuration.
+If neither schemas nor endpoints are defined, then no default will be applied, and you must specify the full configuration.
 
 ## Adjusting database configuration
 


CollierCZ:
Thanks for the suggestion! Appreciate the effort to make it great no matter how big or small the issue. :muscle: 

I think in this case, though, the change inverts the meaning. And I think it was correct before. At least, this is my understanding:
If you don't specify anything, you get defaults for both schemas and endpoints.
if you specify anything (either schemas or endpoints), you get no defaults. None. Not even one you didn't specify.

Whereas I think you change implies:
If you don't specify anything, you do _not_ get defaults. 

So I think it should stay as either/or, but I might be wrong. :shrug:

CollierCZ:
@SabriPlatform , I'm going to close this. Feel free to reopen if you disagree and let me know why. :slightly_smiling_face:

SabriPlatform:
Hi @CollierCZ, I meant to get back and confirm - I agree with your take! I wasn't quite sure while reading the docs with the wording, but your understanding and explanation make complete sense. Thank you for the insight!

PR #1989, opened 2021-11-12
:fire: Replace widget with simple links

CollierCZ:
## Why

Got rid of old links to accounts page and widget that doesn't match the console.

## What's changed

Replaced the widget with links and replaced links to accounts page with links to page to register for a trial.


docs/layouts/partials/header/upper-widget.html
```diff
@@ -1,2 +1,7 @@
-<!-- Platform.sh User Widget -->
-<div id="user-widget"></div>
+<!-- Links to log in or start a trial with Platform.sh -->
+<div class="console_links">
+  <div class="console_links__wrapper">
+    <a class="console_links__login" href="https://console.platform.sh/" rel="noopener noreferrer">Log in</a>
+    <a class="console_links__free-trial" href="https://auth.api.platform.sh/register" rel="noopener noreferrer">Free trial</a>
+  </div>
+</div>
```
maria-at-platform, line 7: It might not be a bad idea to add a line here.

CollierCZ: Is there a process you think this affects? I added one, but it's definitely not the only file we have without an empty line at the end and we don't have code styling guidelines.

maria-at-platform: It's not a big deal, but just to ensure that there aren't a picky POSIX system that could complain.

docs/static/styles/user-customizations.css
```diff
@@ -236,3 +236,59 @@
...
+  line-height:34px;
+  font-weight: 700;
+  transition:background-color 0.25s,color 0.25s
+}
+
+.console_links__free-trial:hover,:focus{
+  text-decoration: none !important;
+  background-color:#28d8b2;
+  color:#fff
+}
```
maria-at-platform, line 294: I assume you copied this from the widget CSS and therefore, it should be fine. However, I've tried to compare the   prod version and the new version locally and the button shows up broken in the mobile view ( I assume it's because of the search field missing).
In the old version (with the search and in the mobile version) the button doesn't show up at all. I'm not sure if that is as intended.
Screenshot 2021-11-15 at 10 21 47
Screenshot 2021-11-15 at 10 33 45

maria-at-platform:
Since we're not linking the widget anymore, should we remove its source code or any other item?

docs/static/styles/user-customizations.css
CollierCZ: The CSS selector was overly broad (all focus, rather than just that element). 

Also, there's no need to test it locally without search. You can use the deployed version (see the checks for the PR). That's the benefit of dogfooding :dog2:

CollierCZ:
We certainly can. It's not actually used anymore, so it doesn't give real benefits, but can definitely get rid of it.

docs/layouts/partials/header/upper-widget.html
CollierCZ: Given the >100 files like this we have, I don't think such a system is an issue.

maria-at-platform (approved):

PR #1990, opened 2021-11-12
Changes the environment tree image

aleksey-likhodievski:
## Why

We want to use the "Main" instead of the "Master" environment name. 

## What's changed

The envrironment tree screenshot is changed.


CollierCZ (approved):

CollierCZ:
Awesome! Thanks for the improvement, @aleksey-likhodievski. :bow:

PR #1997, opened 2021-11-12
Update backup-and-restore.md

chadwcarlson:
## Why

Listed backups are a list of activities, rather than backups that can actually be restored to an environment. There is a `restorable` attribute on the `/environments/ENV/backups` endpoint that we can use with `jq` to filter out available backups. [Slack thread](https://platformsh.slack.com/archives/C0JHEUHQD/p1636738196443800)

## What's changed

Expanded the note on *Restore* with a jq tip.


CollierCZ (approved):

CollierCZ:
Thanks for the improvement, @chadwcarlson!

PR #1998, opened 2021-11-15
:zap: Improve search experience

CollierCZ:
## Why

It wasn't working well
* The top search term (`cli`) did not return the expected page in the top 10 results. There are other examples.
* People requested a full page for search and to be taken there on pressing enter

## What's changed

* Updated instructions and dependencies
* Added a page for search
* Made the input redirect to the page on submission
* Added a button so people can see up to 200 results if they want
* Updated Meilisearch to increase accuracy of results


trolologuy:
what could be a nice addition too, would be to add a “search in community” button or ... even add a checkbox to include search results from the community page :slightly_smiling_face:

Since the search query parameter is using the same format, the “search in community” button should be easy to implement:
https://pr-1998-vk5qcxa-652soceglkw4u.eu-3.platformsh.site/search.html?q=git
vs
https://community.platform.sh/search?q=git

CollierCZ:
Added a button for that at the top of the search page

trolologuy:
For improved UX we could reuse the button styling from [the homepage](https://platform.sh/) to have the same behavior or hover and onclick events.
I'd propose to also a "Search doc site (default)" or something similar, so that it is clear that you can search on different places.
Or maybe even replace the `Search Community site` with a checkbox and add a "global" `Search` button (that would cover both use cases).

chadwcarlson:
This all looks great to me - thank you for finally getting this in! Only things I'm seeing aside are
- maybe the search bar needs a border/shadow on the "full page" 
- I second Paul - not sure the regular side-bar search should disappear on full page.

chadwcarlson (approved):

CollierCZ:
I didn't come up with that text. The results do include items from the website and such, so I don't think there's specific other copy that would cover it. What's wrong with `Other resources`?

MariadeAnton (approved):
OMG !! Thanks so much for fixing this @CollierCZ  and everyone for reviewing until it's fully ready 💪🏼 💯 
image

PR #2002, opened 2021-11-16
:memo: Update PHP documentation with Composer 2

### LINKED ISSUE #2001: Update PHP doc to use Composer 2 phar instead

* [PHP documentation](https://docs.platform.sh/languages/php.html)

The actual documentation for [PHP](https://docs.platform.sh/languages/php.html) says that to use Composer 2, it should be defined as a dependency in `platform.app.yml`, which will download Composer 2 with Composer 1.

Downloading it with Composer 1 may lead to some backward compatibilities issues due to:
* Composer 2 installed globally
* Composer's downloaded dependencies versions are not those that are locked by Composer 2 in its `composer.lock`
* Since Composer 2 is installed globally and it uses its global dependencies which can lead to some backward compatibilities issues

I have run into this issue while using a composer plugin (here Drupal Paranoia). It defines a Command that is based on `symfony/console`'s Command. The `execute` function return value has changed from the `symfony/console:2.8.52` that is locked by `composer.lock` in Composer 2 to the `symfony/console:5.3.10` that is been downloaded with the required Composer 2 by Composer 1.

This difference in the dependency version breaks this plugin, but it might also break other things due to the difference between the dependencies versions.

To fix that, I think we should update the documentation for PHP, to let people know that there is already a PHAR available for composer 2 in the PHP image.

Curryed:
Change the way to use Composer 2, use it with the PHAR available.

## Why

Closes #2001

## What's changed

Change the way the user uses Composer 2 in PHP images

### DIFF
--- a/docs/src/languages/php/_index.md
+++ b/docs/src/languages/php/_index.md
@@ -117,19 +117,7 @@ That allows you to install a forked version of a global dependency from a custom
 PHP images use the `composer` build flavor by default,
 which runs `composer --no-ansi --no-interaction install --no-progress --prefer-dist --optimize-autoloader` if a `composer.json` file is detected.
 
-Note that by default, all PHP containers include the latest Composer 1.x release. If you wish to use Composer 2.x, add it as a `dependency` (see the section below).
-
-```yaml
-dependencies:
-    php:
-        composer/composer: '^2'
-```
-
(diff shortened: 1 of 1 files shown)


CollierCZ:
Closing in favor of #2003, which is the recommended way of doing it. Thanks so much for making us investigate this.

Issue #2005, opened 2021-11-18
Change wording on the public IP page.


maria-at-platform:
### Where on docs.platform.sh should be changed?

https://docs.platform.sh/development/public-ips.html

### What exactly should be updated?

I think the term definition is what is missing from the docs page.  Out of all these items above, the only one that is important is that data never leaves the territory.  And it’s defined as “customer data” not log data. In practice though, I thought it was also log data.
### Additional context

[Conversation]([url](https://platformsh.slack.com/archives/CCKPD2HTQ/p1637165279376800))

PR #2007, opened 2021-11-18
:memo: Cleanup, add clarification to Public IPs

chadwcarlson:
## Why

- The public IPs was a long scroll, when it could be presented a little cleaner.
- Removing ~Location guarantee lacks definition, needed clarification. (https://github.com/platformsh/platformsh-docs/issues/2005)~

## What's changed

- ~short glossary with DLG clarification, linked to in relevant regions (https://github.com/platformsh/platformsh-docs/issues/2005)~ - depending on this conversation, this block can be removed entirely
- cleans page into tables per region, trying to stick to P.sh web system design
- pulls data from added data file `data/regions.yaml` for easy editing/adding new regions
- adds `region-ips` shortcode, which handles DLG definition links; retains per-region anchor links in current page

Closes https://github.com/platformsh/platformsh-docs/issues/2005


0xDSousa:
Since we are not disclosing specific cloud providers, naming regions as "North 5" might be confusing for end users as they can't really map that to location/region. Also, the 5 carries little significance as we don't have North 1-2-3-4.

Console shows "Europe - Stockholm" for North 5 which would be a much clearer describer for potential customers that want to host their services as close as possible to their users.
For services focused on a specific country, it still makes sense to enumerate (e.g., FR-1, FR-2, FR-3).

docs/layouts/shortcodes/region_ips.html
```diff
@@ -0,0 +1,81 @@
+<!-- INDIVIDUAL REGION TABLE SHORTCODE -->
```
CollierCZ, line 1: I find Markdown to be much more readable than HTML, usually even with tables.

docs/layouts/shortcodes/region_ips.html
```diff
@@ -0,0 +1,81 @@
...
+        <th>Region</th>
+        <th>Outbound IPs</th>
+        <th>Inbound IPs</th>
+      </tr>
+      <!-- List individual regions -->
+      {{ range $data }}
+      <!-- Give the region a row -->
+      <tr>
+        <!-- Region overview -->
+        <td style="background: #f4f2f3;">
```
CollierCZ, line 18: Given the added CSS, could this be done there rather than with inline styles?

docs/layouts/shortcodes/region_ips.html
```diff
@@ -0,0 +1,81 @@
...
+                            <h4><a href="#{{ $regionAnchor }}" id="{{ $regionAnchor }}" aria-label="Anchor">{{ .name }}</a></h4>
+                        </td>
+                    </tr>
+                    <tr>
+                        <td><code style="background: white;">{{ .label }}</code></td>
+                    </tr>
+                    <!-- Data location guarantee when relevant -->
+                    {{ if .location_guarantee }}
+                    <tr>
+                        <td class="guarantee-note">(<a href="#glossary">Data location guarantee</a>)</td>
```
CollierCZ, line 34:
```suggestion
                        <td class="guarantee-note">(<a href="#glossary">Data Location Guarantee</a>)</td>
```

To fit the glossary. Or change it there.

docs/src/development/public-ips.md
```diff
@@ -1,254 +1,46 @@
...
+# Data: Add and edit regions in `docs/data/regions.yaml`.
+# Styles: See `docs/static/styles/user-customizations.css`.
+# Table shortcode: See `docs/layouts/shortcodes/region_ips.html`.
 title: "Public IP addresses"
 weight: 14
 sidebarTitle: "Public IPs"
 description: |
   Platform.sh regions reach the outside through a limited number of IP addresses.
+  These IP addresses are stable, but not guaranteed to never change.
+  Prior to any future change, all affected customers will receive ample warning.
```
CollierCZ, line 11: I think this information is important, but I don't think  it really works in the description, which should be about making people want to click to read details or whatever.

I'd also switch "all affected customers" so that it's in the second person. Maybe something like "Before any change, you're notified well in advance if you're affected."?

docs/src/development/public-ips.md
```diff
@@ -1,254 +1,46 @@
...
 ---
 
 {{< description >}}
 
-{{< note >}}
+## Glossary
 
-These IP addresses are stable, but not guaranteed to never change.
-Prior to any future change, all affected customers will receive ample warning.
+**Data Location Guarantee:** Customer data never leaves the region.
```
CollierCZ, line 18: I feel who the customer is might be slightly ambiguous. Like if it's an ecommerce site or something -- is it site data or the data of their customers or something else. I also feel like the original issue showed that it wasn't clear with log data, so maybe that should be explicitly added?

CollierCZ (commented):
A few small things I noticed in passing. I saw that this might not be done soon?

docs/layouts/shortcodes/region_ips.html
chadwcarlson: I get you - some of the feedback in these sections we've gotten from other teams here is that MD tables get messy, and difficult to format. Part of the reason I split off into a data file + shortcode so they didn't have to deal with it. 

Aside from that, I was trying to clean up while preserving individual region anchors, in case someone was used to linking a specific region (AU-2, for example).

docs/layouts/shortcodes/region_ips.html
chadwcarlson: related to your separate comment, we've resolved to remove the definition entirely. So alongside cleanup, which I'd be curious to see what you think of, this PR removes the definition rather than giving it Glossary space.

chadwcarlson:
@CollierCZ Tried to address all your points here. PR 1) removed DLG reference, and 2) does some cleanup of the page. 

If you think it make sense to revisit [Diogo's comment](https://github.com/platformsh/platformsh-docs/pull/2007#issuecomment-977888753) above here, let me know. Otherwise it may be time to revisit region naming in a follow-up.

docs/layouts/shortcodes/region_ips.html
CollierCZ: I definitely like the use of YAML (although with all the quotes it looks more like JSON :see_no_evil: ). I was just thinking even the template could be Markdown, but it's not a big deal.

docs/src/development/public-ips.md
```diff
@@ -1,4 +1,7 @@
 ---
+# Data: Add and edit regions in `docs/data/regions.yaml`.
+# Styles: See `docs/static/styles/user-customizations.css`.
+# Table shortcode: See `docs/layouts/shortcodes/region_ips.html`.
```
CollierCZ, line 4: :+1: Nice

CollierCZ:
Given that it's not the first comment to mention the names, I think it makes sense to address it. I'd use the countries that we have listed publicly already: https://platform.sh/regions/ And take out numbers of the names when there's only one in the country.

CollierCZ (approved):
I changed the names in Europe and made some small styling adjustments.

trolologuy:
Maybe the names could be the same as the ones returned by the API?

For e.g:
```
vagrant:~/ $ platform api:curl regions | jq '.regions[] | .id + ": " + .provider + " - " + .zone + " - " + .timezone' | grep 'eu-2'                                                                                                                [10:52:50]

9:"eu-2.platform.sh: AWS - Europe - Europe/Paris"```

CollierCZ:
They mostly are. I just didn't feel the need to repeat "Europe" twice. Also, the ones for the US are less helpful in the API. "North America" I believe.

PR #2011, opened 2021-11-22
Update registry.json

chadwcarlson:
## Why

Some example snippets are generated in docs, which then get reused in console within the Project Setup Wizard. For Ruby, there was some escaped double-quotes that were causing issues when a user copy-pastes the resulting `.platform.app.yaml` snippet. 

The same is happening for `root: "public"`, so these escape quotes aren't being seen on the console side.

## What's changed

Removed double quotes to give a workable config file.

### DIFF
--- a/docs/data/registry.json
+++ b/docs/data/registry.json
@@ -805,7 +805,7 @@
           "socket_family": "unix"
         },
         "commands": {
-          "start": "\"unicorn -l $SOCKET -E production config.ru\""
+          "start": "unicorn -l $SOCKET -E production config.ru"
         },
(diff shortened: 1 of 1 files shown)


chadwcarlson:
**Current** (https://docs.platform.sh/registry/images/examples/full/ruby.web.app.yaml)

```yaml
web:
    upstream:
        socket_family: unix
    commands:
        start: "unicorn -l $SOCKET -E production config.ru"
    locations:
        /:
            root: "public"
            passthru: true
            expires: 1h
            allow: true
```

PR #2015, opened 2021-11-23
📝 Update migrating doc and change master to main

SabriPlatform:
Just changing use of Master to Main.

## Why

I am just changing the use of Master to Main for the default branch.

## What's changed

Change master to main where mentioned


CollierCZ (approved):
Thanks for the catch, @SabriPlatform!

PR #2021, opened 2021-11-30
:bug: Render text in shortcodes with render hooks

CollierCZ:
## Why

To make sure links and such get transformed.

https://docs.platform.sh/guides/wordpress/deploy/configure.html#application-container-platformappyaml The link on `mounts` went to the `.md` file rather than the built page.

## What's changed

Switched from `Markdownify` to `.Page.RenderString`. These hooks aren't available in `Markdownify`

### DIFF
--- a/.vale.ini
+++ b/.vale.ini
@@ -6,6 +6,6 @@ MinAlertLevel = suggestion # suggestion, warning, or error
 
 [*.md]
 BasedOnStyles = Platform,Vale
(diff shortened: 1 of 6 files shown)


trolologuy:
Previously broken link under `been given corresponding mounts defined`:
/guides/wordpress/deploy/configure.html#application-container-platformappyaml

Is now working as expected.
Enabling the merge ✅

trolologuy (approved):

PR #2030, opened 2021-12-03
:arrow_up: Upgrade Node and dependencies

CollierCZ:
## Why

There were some security warnings, partly for unused dependencies. This should get things up to date.

## What's changed

* Updated to Node 16 (the current LTS version)
* Removed some unused dependencies
* Fixed ESLint for XSS


trolologuy:
Some dependencies are indeed in need of further updates, either because they are outdated or deprecated. 
Namely:
- https://console.platform.sh/platformsh-internal/652soceglkw4u/pr-2030/log/sqjlzpnminyzs#L8-L11
- https://console.platform.sh/platformsh-internal/652soceglkw4u/pr-2030/log/sqjlzpnminyzs#L167

For now, it doesn't seem to break functionality.
This may have to be manually addressed at some point. 
Maybe worth adding an issue and trying to update each warning manually and make sure the site is still functioning until all of the warnings disappear?

Four other warnings would need to be investigated as well:
- https://console.platform.sh/platformsh-internal/652soceglkw4u/pr-2030/log/sqjlzpnminyzs#L182
- https://console.platform.sh/platformsh-internal/652soceglkw4u/pr-2030/log/sqjlzpnminyzs#L189
- https://console.platform.sh/platformsh-internal/652soceglkw4u/pr-2030/log/sqjlzpnminyzs#L191
- https://console.platform.sh/platformsh-internal/652soceglkw4u/pr-2030/log/sqjlzpnminyzs#L191

CollierCZ:
Yes, there are dependencies of dependencies that are out of date, but I don't think we can spend so much time chasing them all down. I was just hoping for some quick wins and at least stopping the critical security alerts.

In general, many modern JS apps will have notifications like that, but it's not always the best sign of an issue. For example, in [previous logs](https://console.platform.sh/platformsh-internal/652soceglkw4u/main/log/5h4xrbw4qfcpk#L181) you can see it was using `core-js@2.6.12`. It was only _after_ upgrading other packages that the warning about that version being deprecated came out. So the existence of a warning isn't by itself an issue.

But in this case, one of the issues should be easy enough to fix. We'll see if search still works without it (it should, since it was deprecated years ago).

The other warnings are known issues, but not major things. They've mostly come up in most logs before (except the one that is because I got the linting working that wasn't before, but it's just that errors are logged to the console, and I don't want to set up an entire error handling system before this gets merged).

CollierCZ:
I fixed some of the additional small warnings, @trolologuy . I would leave gulp and its dependencies as they are for now.

chadwcarlson (approved):

PR #2037, opened 2021-12-07
:memo: Improve SSH docs

### LINKED ISSUE #1973: Update SSH doc

https://docs.platform.sh/development/ssh.html

Update for:
- Content accuracy
- Includes what it should
- Fits template
- Fits style guide

Incorporate info also on SSH connections to services: https://docs.platform.sh/configuration/services.html#open-an-ssh-tunnel-directly
Link back to this section from that page so it's clear from services how to SSH to a service.

### Additional context

Focus on the goal of connecting to the application.

Keep SSH through CLI as the main recommendation. API keys should be described, but as an alternative, not the best way.

Can consider splitting into separate pages (setting up authentication and then using SSH). Not necessarily, but maybe.

CollierCZ:
## Why

Closes #1973

## What's changed

* Restructed SSH index page to be more focused
* Split SSH keys into a separate page with a link from the main SSH page
* Numbered the headings on hte new SSH keys page, but otherwise mostly left as is.
* Moved info on SSH tunnels from services page to SSH page.

  The flow for connecting to a service is now somewhat interrupted (it's on two pages). But it makes sense to me because each topic is different: one about the service information and the other covering everything about SSH connections.
* Removed duplicate instructions from `access-site.md`, but didn't go through the entire page.

  Similarly as with tunnels, this moves some flows to spreading over two pages.
* Removed errors noted by Vale on services page (but didn't touch other parts of the page).

We could use transclusion to repeat commands rather than having the flows be broken up. I think it works as is, but would welcome other thoughts.


MariadeAnton (dismissed):
.

MariadeAnton (approved):
All good to go from my PoV. 

You've done a fantastic job splitting up the existing SSH keys setup guide into their own how-to guide, focusing the SSH into their application and the simplification on phrases around SSH clients 👏🏼 

As to SSHing into services splitting up, I see your point, I think we could make this guide specifiy which credentials specifically they need from the list linked, so that it's easier to follow - they may even have those stored somewhere already. Or is it that we need all of them? I'm not sure @trolologuy  might know better than I.

MariadeAnton (approved):
🚀   final touches are definitely sparkling ✨ , this PR is so ready to go. Thanks @CollierCZ !

PR #2041, opened 2021-12-09
:memo: Add an additional reference to the templates

trolologuy:
## Why

Attempt to make the usage of templates more obvious.

## What's changed

Added one reference to the templates.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,19 +11,19 @@ For style and formatting guidance, see:
 * [Content style guide](contributing/content-style.md)
 * [Markup reference](contributing/markup-format.md)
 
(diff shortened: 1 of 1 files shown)


CollierCZ:
I like the idea, but a link to the templates is already [under the heading Adding pages](https://github.com/platformsh/platformsh-docs/blob/main/CONTRIBUTING.md?plain=1#L24). I think that's where it belongs in the workflow. Is it not clear enough? I'm certainly open to making it clearer, but I don't think it's general guidance like the style guides are.

trolologuy (commented):
It is much better like that, the context is clearer ✅

CollierCZ (approved):

PR #2044, opened 2021-12-09
:bug: Limit region table CSS to that table

CollierCZ:
## Why

In #2007, I unfortunately entered the CSS for region tables too broad so they applied everywhere. See the table here for an example:
https://docs.platform.sh/configuration%2Fservices.html#type

## What's changed

Made the CSS for the region table only apply to that table and not cells everywhere.

### DIFF
--- a/docs/static/styles/user-customizations.css
+++ b/docs/static/styles/user-customizations.css
@@ -302,7 +302,7 @@
   border-spacing: 2px;
 }
 
-.region-table > tr, td, th {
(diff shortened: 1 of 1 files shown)


trolologuy (approved):
The table looks much nicer [now](https://pr-2044-oxvniya-652soceglkw4u.eu-3.platformsh.site/configuration%2Fservices.html#type).
✅  Merge approved.

PR #2051, opened 2021-12-13
:bug: Fix typos

trolologuy:
## Why

Fix typos


CollierCZ (approved):

CollierCZ:
* [x] Small improvements to templates, which have no effect on the built site and so other things don't need to be checked.

PR #2055, opened 2021-12-16
Update gitlab.md

michielkenis:
Updated with new format of GitLab

## Why

Closes #{ISSUE_NUMBER}


CollierCZ (approved):
Thanks for the fix, @michielkenis! We're having a temporary problem with our pipeline. As soon as that's taken care of, we'll get this merged in.

CollierCZ:
The issue seems to be on the GitLab side with permissions. Will update the docs with a note when the cause has been more precisely located.

CollierCZ:
Handled in #2056

PR #2057, opened 2021-12-20
:memo: Updating troubleshooting for git push out of disk issue

### LINKED ISSUE #2026: :memo: Add too big git message to troubleshooting

https://docs.platform.sh/development/troubleshoot.html

It's possible to get the following error message when trying to push to git:

```
There is not enough free space to complete the push
```

This can block deployments (even production ones). This means the git repository is taking up too much space. In such cases, users should open a support ticket.

### Additional context

_No response_

trolologuy:
## Why

Closes #2026

### DIFF
--- a/docs/src/development/troubleshoot.md
+++ b/docs/src/development/troubleshoot.md
@@ -128,7 +128,7 @@ Make sure that the paths for files like media files, dependencies, and databases
 
 If large files are already in the repository, the open-source tool [bfg-repo-cleaner](https://rtyley.github.io/bfg-repo-cleaner/)
 can assist in cleaning up the repository by purging older commits, removing unnecessary files, and more
-Create a [support ticket](https://console.platform.sh/-/users/~/tickets/open) to have the disk space temporarily increased.
+To have the disk space temporarily increased, create a [support ticket](https://console.platform.sh/-/users/~/tickets/open).
 
 ## Stuck build or deployment
 
--- a/docs/src/development/troubleshoot.md
+++ b/docs/src/development/troubleshoot.md
@@ -123,9 +123,8 @@ They can be used for your data: file uploads, logs, and temporary files.
 If you see the following message when attempting to do a `git push`:
 `There is not enough free space to complete the push`
 
-It usually indicates that large files are present in the repository.
-Media, dependencies, database, or large data files shouldn't be present in the repository.
-Make sure the paths for those are correctly set in the `.gitignore` file.
+It usually indicates that large files are present in the repository (where they shouldn't be).
+Make sure that the paths for Media, dependencies, database, or large data files are correctly set to be ignored in the `.gitignore` file.
 An open-source tool like [git-sizer](https://github.com/github/git-sizer) can help locate large files and provide general guidance.
 
 If large files are already in the repository, the open-source [bfg-repo-cleaner](https://rtyley.github.io/bfg-repo-cleaner/) can assist cleaning up the repository by purging older commits, removing unnecessary files, ...
--- a/docs/src/development/troubleshoot.md
+++ b/docs/src/development/troubleshoot.md
@@ -127,8 +127,7 @@ This usually indicates that large files are present in the repository (where the
 Make sure that the paths for files like media files, dependencies, and databases are set to be ignored in your `.gitignore` file.
 
 If large files are already in the repository, the open-source tool [bfg-repo-cleaner](https://rtyley.github.io/bfg-repo-cleaner/)
-can assist in cleaning up the repository by purging older commits, removing unnecessary files, and more
-To have the disk space temporarily increased, create a [support ticket](https://console.platform.sh/-/users/~/tickets/open).
+can assist in cleaning up the repository by purging older commits, removing unnecessary files, and more.
 
 ## Stuck build or deployment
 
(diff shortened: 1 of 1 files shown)


trolologuy:
Direct link to the preview:
https://pr-2057-kgfqc7q-652soceglkw4u.eu-3.platformsh.site/development/troubleshoot.html#git-push-fails-due-to-lack-of-disk-space

docs/src/development/troubleshoot.md
```diff
@@ -118,6 +118,17 @@ You can write to disk a `build` hook to generate anything you need later.
 Or you can declare writable [mounts](../configuration/app/storage.md#mounts), which are writable even during and after deploy.
 They can be used for your data: file uploads, logs, and temporary files.
 
+## Git push fails due to lack of disk space
+
+If you see the following message when attempting to do a `git push`:
+`There is not enough free space to complete the push`
+
+It usually indicates that large files are present in the repository (where they shouldn't be).
+Make sure that the paths for media files, dependencies, or databases, ... are set to be ignored in the `.gitignore` file.
```
CollierCZ, line 127:
```suggestion
Make sure that the paths for files like media files, dependencies, and databases are set to be ignored in your `.gitignore` file.
```

The ellipses is essentially like a less clear version of `etc.`, which [Google recommends against](https://developers.google.com/style/word-list?hl=en#etc).

docs/src/development/troubleshoot.md
```diff
@@ -118,6 +118,17 @@ You can write to disk a `build` hook to generate anything you need later.
...
+
+If large files are already in the repository, the open-source [bfg-repo-cleaner](https://rtyley.github.io/bfg-repo-cleaner/) can assist in cleaning up the repository by purging older commits, removing unnecessary files, ...
```
CollierCZ, line 129:
```suggestion
If large files are already in the repository, the open-source tool [bfg-repo-cleaner](https://rtyley.github.io/bfg-repo-cleaner/)
can assist in cleaning up the repository by purging older commits, removing unnecessary files, and more
```

docs/src/development/troubleshoot.md
```diff
@@ -118,6 +118,17 @@ You can write to disk a `build` hook to generate anything you need later.
...
+Create a [support ticket](https://console.platform.sh/-/users/~/tickets/open) to have the disk space temporarily increased.
```
CollierCZ, line 130:
```suggestion
To have the disk space temporarily increased, create a [support ticket](https://console.platform.sh/-/users/~/tickets/open).
```

This suggestion is about the general [clause order recommendations]. I'd also consider starting with some more context (like "if this doesn't work" or something).

Also, would it only be for a temporary increase? I thought it wasn't temporary. If it's only temporary, it seems like we should push the other fixes more?

CollierCZ (changes requested):
It seems like a clear explanation of the problem and what to do. 

I put in some style suggestions, mostly just to make sure we go forward on the same page. They wouldn't be blockers normally. Only thing I wonder about more seriously is the "temporarily" part of it.

docs/src/development/troubleshoot.md
trolologuy: The temporality of the fix depends on the size that is required. As far as I know, some sizes can not be set as a global fix (but I might be wrong). Anyway, repositories that have sizes over 4-8 GB are probable signs of issues in development. 
For e.g [limit in Github is 5GB](https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-large-files-on-github)

trolologuy:
It should be good to go now, @CollierCZ . Set to auto-merge once approved.

docs/src/development/troubleshoot.md
CollierCZ: The fact that its a sign of an issue seems to me a sign that we shouldn't be suggesting a temporary fix then. If it's a development issue, then that should be fixed. Otherwise, it's not a temporary fix: it's intentional and should be introduced differently.

CollierCZ (changes requested):
I would still not use "temporarily".

docs/src/development/troubleshoot.md
trolologuy: You are correct, I removed that part out of the documentation. It usually is a sign of a development issue, and the courtesy that support allows is probably not always a good approach.

docs/src/development/troubleshoot.md
```diff
@@ -118,6 +118,19 @@ You can write to disk a `build` hook to generate anything you need later.
...
+You might see the following message when attempting to run `git push`:
+`There is not enough free space to complete the push`
+
+This usually indicates that large files are present in the repository (where they shouldn't be).
+Make sure that the paths for files like media files, dependencies, and databases are set to be ignored in your `.gitignore` file.
+
+If large files are already in the repository, the open-source tool [bfg-repo-cleaner](https://rtyley.github.io/bfg-repo-cleaner/)
+can assist in cleaning up the repository by purging older commits, removing unnecessary files, and more.
+
+If none of the propositions above worked, create a [support ticket](https://console.platform.sh/-/users/~/tickets/open).
```
CollierCZ, line 132:
```suggestion
If none of these suggestions work, create a [support ticket](https://console.platform.sh/-/users/~/tickets/open).
```

I'd try to stay away from "propositions" cause it's not so common of a word and "above" for [accessibility reasons](https://developers.google.com/style/accessibility?hl=en) (avoiding "directional" language).

Otherwise, it all looks good to me.

CollierCZ (approved):

PR #2058, opened 2021-12-20
:lipstick: Small improvements to styling

CollierCZ:
## Why

* In long tables (like for [PHP extensions](https://docs.platform.sh/languages%2Fphp%2Fextensions.html)), it could be hard to see what column lined up with what heading.
* When a table of contents was long or the window not tall enough, a lot of the ToC was hidden and inaccessible.

## What's changed

* Made the headings sticky so they're still shown when scrolling down. Should be a progressive enhancement (just not appear if the browser doesn't support it) rather than a breaking change. [Example](https://pr-2058-csuicfi-652soceglkw4u.eu-3.platformsh.site/languages/php/extensions.html)
* Made long ToCs scrollable

### DIFF
--- a/docs/themes/avocadocs/static/styles/customizations.css
+++ b/docs/themes/avocadocs/static/styles/customizations.css
@@ -1083,6 +1083,12 @@ td, th {
   padding: 8px;
 }
 
+th {
+  position: sticky;
+  top: calc(1.5rem * 3);
+  background-color: var(--primary);
+}
+
 blockquote {
   color: #333;
   border: none;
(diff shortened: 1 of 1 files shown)


trolologuy (approved):
Maybe the three items (searchbox, heading, and toc) could be aligned?
Screenshot 2021-12-21 at 17 43 51

But I'm not sure what's the best practice to have in that case.

Besides that detail above, the rest is fine for me and it can be merged ✅ :
- [x] All links work.

CollierCZ:
I'd definitely like to have them aligned, but it'd require a lot of fiddling with more fragile CSS (creating white rectangles to cover things and such). So I probably wouldn't right now, though we can keep it in mind for a more comprehensive redesign.

PR #2059, opened 2021-12-20
:bug: Fix search query on search page

CollierCZ:
## Why

Search page was loading results for an empty string even when the URL had a query.

## What's changed

Change the useEffect hook to not rely on useState hook and just use the URL query directly.


trolologuy (commented):
- [x] The [final built page](https://pr-2059-pvtjjoq-652soceglkw4u.eu-3.platformsh.site/search.html?q=ssh) is working as expected.
- [x] All links work.

trolologuy (approved):

PR #2064, opened 2021-12-23
:memo: Clarify Let's Encrypt limits per env

CollierCZ:
## Why

Docs said that Let's Encrypt limited environments by domain and that exceeding the limits meant some domains would succeed and some fail. This isn't correct
[Additional context](https://github.com/orgs/platformsh/projects/3/views/1)

## What's changed

Clarified that the limit regards hostnames.

### DIFF
--- a/docs/src/configuration/routes/https.md
+++ b/docs/src/configuration/routes/https.md
@@ -81,11 +81,15 @@ The use cases for this configuration are few.
 More complex routing logic is also possible if the situation calls for it.
 
 {{< note >}}
-Let's Encrypt has a limit of 100 TLS certificates per environment.
-If you define both a `{default}` and `www.{default}` route for each domain you use, that will give you a limit of 50 domains.
-Adding more than that will result in a warning on deploy and some domains will not be issued a TLS certificate.
-If you need more than that, we recommend obtaining additional certificates or a wildcard certificate from another TLS provider.
-Alternatively, consider splitting your project up into multiple discrete Platform.sh projects.
+
+Let's Encrypt has a limit of 100 hostnames per certificate.
+Each of your environments has 1 certificate.
+If you define both a `{default}` and `www.{default}` route for each domain you use, you have a limit of 50 hostnames.
+Adding more than that results in a warning on deploy and no new TLS certificates are issued.
+
+If you need more hostnames than that, obtain additional certificates or a wildcard certificate from another TLS provider.
+Alternatively, consider splitting your project up into multiple Platform.sh projects.
+
 {{< /note >}}
 
 ## TLS configuration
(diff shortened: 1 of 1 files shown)


docs/src/configuration/routes/https.md
```diff
@@ -81,11 +81,15 @@ The use cases for this configuration are few.
 More complex routing logic is also possible if the situation calls for it.
 
 {{< note >}}
-Let's Encrypt has a limit of 100 TLS certificates per environment.
-If you define both a `{default}` and `www.{default}` route for each domain you use, that will give you a limit of 50 domains.
-Adding more than that will result in a warning on deploy and some domains will not be issued a TLS certificate.
-If you need more than that, we recommend obtaining additional certificates or a wildcard certificate from another TLS provider.
-Alternatively, consider splitting your project up into multiple discrete Platform.sh projects.
+
+Let's Encrypt has a limit of 100 hostnames per certificate.
```
krishna-kashyap, line 85: could we use domains instead of hostnames please?
```suggestion
Let's Encrypt has a limit of 100 domains (to be more precise 100 SAN or Subject Alternative Names) per certificate.
```

CollierCZ: @pjcdawkins requested hostnames instead of domains :smile:

pjcdawkins: because

1. A "domain" is already a defined thing you can add to your platform.sh project. Each domain can result in 0 or more names in the TLS certificate, not necessarily 1 name.
1. "Hostname" is the word Let's Encrypt uses on their page:

https://letsencrypt.org/docs/rate-limits/
> You can combine multiple hostnames into a single certificate, up to a limit of 100 Names per Certificate.

docs/src/configuration/routes/https.md
```diff
@@ -81,11 +81,15 @@ The use cases for this configuration are few.
...
+Each of your environments has 1 certificate.
```
pjcdawkins, line 86:
```suggestion
Each of your environments can have 1 of these automatically provisioned certificates.
```
or
```suggestion
Each of your environments can have 1 Let's Encrypt certificate.
```

docs/src/configuration/routes/https.md
trolologuy: Solving the thread without applying the suggestion since it makes sense to stick with Let's Encrypt terminology (even if domain would be more what the customer might expect).

trolologuy (approved):
- [x] The [final built page](https://pr-2064-yacar6i-652soceglkw4u.eu-3.platformsh.site/configuration/routes/https.html#using-https) is working as expected.

PR #2067, opened 2021-12-29
:art: Routes port limit info

### LINKED ISSUE #2035: :memo: Note port limitations for routes

https://docs.platform.sh/configuration/routes.html

Note that non-default ports (other than `80` and `443`) aren't supported and can't be includes in routes configuration. Try to find a good place to note this as a fact without needing to draw too much attention to it as most people won't care.

### Additional context

_No response_

trolologuy:
## Why

Closes #2035


trolologuy:
We should probably also add a link to the `/configuration/routes/https.md` page

docs/src/configuration/routes/_index.md
```diff
@@ -68,24 +68,18 @@ In the future, Platform.sh will support multiple endpoints per application.
 
 ## Route limits
 
-Although there is no fixed limit on the number of routes that can be defined,
-there is a cap on the size of generated route information.
-This limitation comes from the Linux kernel, which caps the size of environment variables.
-The kernel limit on environment variables is 32 pages. Each page is 4k on x86 processors,
-resulting in a maximum environment variable length of 131072 bytes.
-If your `routes.yaml` file would result in too large of a route information value it will be rejected.
+- {{% lets_encrypt_limitations %}}
```
CollierCZ, line 71: This won't work without #2069 first

trolologuy:
See [the changes in 2069](https://github.com/platformsh/platformsh-docs/pull/2069/files#diff-d178dfa638c867651b2acf1f91a1e9454e4009974d5b7a275aac5132edbaeda0) for the `{{% lets_encrypt_limitations %}}` placeholder.

docs/src/configuration/routes/_index.md
```diff
@@ -276,14 +276,13 @@ The full list of generated route information is often much larger than what's sp
...
-If your `routes.yaml` file is rejected for being too big,
-the best alternative is to move redirect routes to the application rather than relying on the router.
-You can also collapse multiple routes into a [regular expression-based redirect](./redirects.md#partial-redirects).
-
-Let's Encrypt has a limit of 100 hostnames per certificate and
-each of your environments can have 1 Let's Encrypt certificate.
-If you define both a `{default}` and `www.{default}` route for each domain you use, you have a limit of 50 hostnames.
-Adding more than that results in a warning on deploy and no new TLS certificates are issued.
+Should your `routes.yaml` file be rejected due to an excessive size, either:
+- move any redirect routes to the application,
```
CollierCZ, line 280:
```suggestion
- move redirect routes to the application,
```
`any` implies that there might not be any, but if the file is rejected for being too big, I think there have to be some.

CollierCZ (changes requested):
I like the idea of a list to present the two options. It slightly goes away from one recommendation and a backup, but I think that's probably fine like that. I don't really know myself.

The use of the transclusion is also :+1: And great to include the ports info, which seems to have gone missing :detective: 

I'd just point out some minor points to get the style consistent. Partly it seems to come from making things more formal (using modals and being longer). I'd try to aim for more short and direct:
* https://developers.google.com/style/word-list?hl=en#should Note the examples under "A state is actual" -- this kind of modal verb can be difficult for some language learners
* I feel like "too big" is clearer than "an excessive size"
* https://developers.google.com/style/lists?hl=en#capitalization-and-end-punctuation There's a habit that I see often in formal and academic writing to treat lists as part of the sentence that surround them. This can sometimes be harder to parse. Treat the list as separate from what surrounds it -- like someone might be scanning the page and stop at just the list.
* Remember to use relative links: https://github.com/platformsh/platformsh-docs/blob/main/contributing/markup-format.md#links

CollierCZ (approved):
Looks good to me :rocket:

PR #2073, opened 2022-01-05
:bug: Fix description of disk property for apps

CollierCZ:
## Why

Docs said the `disk` property could be a boolean, but it defaults to `null`, not `false`.

## What's changed

Changed it to `null`.


trolologuy (approved):

PR #2078, opened 2022-01-07
:bug: Fix region names for public IPs

CollierCZ:
## Why

Some of the region names were incorrect (apparently based on the timezone rather than the location. See [additional context](https://github.com/orgs/platformsh/projects/3/views/1).

## What's changed

Since cities seemed to be possibly wrong, switched to countries, which should be enough detail.

### DIFF
--- a/docs/data/regions.yaml
+++ b/docs/data/regions.yaml
@@ -1,6 +1,6 @@
 europe:
     eu:
-        name: Dublin
+        name: Ireland
         label: eu.platform.sh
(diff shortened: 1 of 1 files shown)


trolologuy (approved):
- [x] The final [built page](https://pr-2078-lcvoyzi-652soceglkw4u.eu-3.platformsh.site/development/public-ips.html#ip-addresses-by-region) is working as expected.

PR #2079, opened 2022-01-07
:sparkles: Add keywords to search

### LINKED ISSUE #2061: Add keywords for search

/configuration/app/app-reference.html

We'd like specific pages to be findable by searching for specific words. For example, the app reference when searching for `.platform.app.yaml` (this may also involve a problem with escaping characters like `.`). Add keywords or other to make these pages findable.

### Additional context

_No response_

CollierCZ:
## Why

Closes #2061

## What's changed

Added keywords to:
* JSON index builder
* A few pages
* Searchable attributes
* Results when present

Also made it so matches are styled to be seen. The specific design could be changed, but I think it helps sometimes to see why something is shown.

At some point, will need to work on making the top result for a given page be the top heading. Right now, it only shows one result per page (intentionally), but it's often lower on the page even for a keyword match. Maybe this is enough for now though?

Example: https://pr-2079-sll534y-652soceglkw4u.eu-3.platformsh.site/search.html?q=.platform.app.yaml


trolologuy:
I like how the keywords are displayed, however, the highlighting of the found match in italic is not fully correct.
https://pr-2079-sll534y-652soceglkw4u.eu-3.platformsh.site/search.html?q=ssh
`.sh` is found as a match.
vs:
https://docs.platform.sh/search.html?q=ssh

Also, why did the number of results change that drastically in the `ssh` example? + why is there no `additional results` button anymore?

CollierCZ:
> the highlighting of the found match in italic is not fully correct.
> https://pr-2079-sll534y-652soceglkw4u.eu-3.platformsh.site/search.html?q=ssh
> .sh is found as a match.
> vs:
> https://docs.platform.sh/search.html?q=ssh

I think this is conflating two different things. To account for typos and such, sometimes matches are based on not the complete word (it used to be even worse). So sometimes for "ssh" there's a match for "sh". I think that happens currently. It's just it wasn't highlighted before. The highlighting is just showing you what the match is now.

> Also, why did the number of results change that drastically in the ssh example? + why is there no additional results button anymore?

My hope is that it just wasn't up to date with `main`. We'll see soon.

CollierCZ:
Yes, the results are basically the same now for `ssh`.

CollierCZ:
And with the last change the use case for the original issue is better: https://pr-2079-sll534y-652soceglkw4u.eu-3.platformsh.site/search.html?q=.platform.app.yaml

trolologuy (approved):
Behavior is now exactly as one would expect.
We might also want to [remove the tables from the displayed preview](https://pr-2079-sll534y-652soceglkw4u.eu-3.platformsh.site/search.html?q=.platform.app.yaml)? (but still, include them in the search)

CollierCZ:
Not sure I see the tables there. But we could look at it as a separate issue. If it's important, open one.

PR #2081, opened 2022-01-10
:memo: Update supported Redis version

CollierCZ:
## Why

We now support version 6.2 of Redis

## What's changed

Updated the registry to reflect that.


MariadeAnton (approved):
⚡️

PR #2090, opened 2022-01-18
:arrow_up: Upgrade dependencies

CollierCZ:
And remove styled components, which was large and unnecessary

## Why

Some dependencies were out of date. [Additional context](https://github.com/orgs/platformsh/projects/3).

## What's changed

Updated dependencies for docs site and search app. Also removed styled components, which was only a single rule that was unnecessary and is large and complex.


trolologuy (approved):

PR #2101, opened 2022-01-24
Add troubleshot section to ruby for BUNDLE_DEPLOYMENT=1

spk:
## Why

Report by @Theosakamg on his project

## What's changed

Add troubleshot section to ruby for BUNDLE_DEPLOYMENT=1, this need a explanation for ruby users that are not expert on those issues

### DIFF
--- a/docs/src/languages/ruby.md
+++ b/docs/src/languages/ruby.md
@@ -28,17 +28,17 @@ A complete example is included at the end of this section.
 
 2. Setup environment variables.
 
-    Rails runs by default on a development environment.
-    You can change the Rails/Bundler via those environment variables,
-    some of which will be defaults at some point on Platform.sh.
+   Rails runs by default on a development environment.
+   You can change the Rails/Bundler via those environment variables,
+   some of which are defaults on Platform.sh.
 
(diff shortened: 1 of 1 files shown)


CollierCZ:
@spk , thanks for the suggestion! This means that `BUNDLE_DEPLOYMENT` is already set to `1` by default, right? Are any of the other variables [listed as "will be default"](https://pr-2101-crnshwi-652soceglkw4u.eu-3.platformsh.site/languages/ruby.html#unicorn-based-rails-configuration) already switched over or just this one?

spk:
@CollierCZ  Yes the project is on EU-3, not sure about the other regions, all of them should be the defaults soon

Should I change the `will be` by `the current default`?

CollierCZ:
I can change it, no worries. Just wasn't sure if it was for the one or for all of them.

spk:
> I can change it, no worries. Just wasn't sure if it was for the one or for all of them.

Ok no problemo, I think only this one will cause issues with builds failing the others should be fine with existing projects

CollierCZ (approved):

PR #2102, opened 2022-01-25
:memo: Add note about /tmp disk usage

trolologuy:
## Why

Disk space on `/tmp` reclaimed by an app or a user is not automatically deleted.
Additional context [can be found here](https://github.com/orgs/platformsh/projects/3/views/1)


trolologuy:
Adding @printerjam in the loop.

CollierCZ (approved):
I think its a helpful addition. I added some small suggestions for style, but it's completely up to you.

CollierCZ (approved):

PR #2105, opened 2022-01-26
⬆️ Update Meilisearch

### LINKED ISSUE #2084: :arrow_up: Update MeiliSearch

We should update MeiliSearch to version 0.25. This means we have to change how we manage API keys. It does mean we can create more keys for monitoring and the like.

### Additional context

[Blog post on version](https://blog.meilisearch.com/whats-new-in-v0-25/)
[Additional context](https://github.com/orgs/platformsh/projects/3)

CollierCZ:
## Why

Closes #2084 

## What's changed

Updated to version 27.1
Didn't add keys. Right now, the extra memory is enough and there seems to be an endpoint for status checks anyway. Can add keys on production, but I think the process wouldn't be to add a specific key, but rather to generate a key with a random ID, so I'm not sure scripting the process would help so much.

### DIFF
--- a/docs/config/_default/params.yaml
+++ b/docs/config/_default/params.yaml
@@ -1,6 +1,6 @@
 
 # Basics
-version: "1.0.0"
+version: "1.0.1"
 description: Platform.sh User Documentation
 author: Platform.sh
 title: User documentation
--- a/docs/deploy.sh
+++ b/docs/deploy.sh
@@ -4,12 +4,12 @@
 # Get the data
 MEILI_CONFIG_DEST=public/scripts/xss/dist/config/config.json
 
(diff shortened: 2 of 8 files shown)


CollierCZ:
This will require Python 3.10 for our search build to work.

trolologuy (approved):
Are there specific changes to look out for? Or specifics to test out? The [changelog](https://github.com/meilisearch/meilisearch/releases/tag/v0.27.0) brings some changes, like typo tolerance which might be interesting.

I tested for regression and that the page and the search work. ✅ 

CollierCZ:
Well, the original idea was to upgrade in large part for the ability to add API keys.

PR #2108, opened 2022-01-26
:memo: Clarify when domains are challenged

CollierCZ:
## Why

The docs said `www` routes were checked by default, but it's only if they're included

## What's changed

Clarified the behavior and added a link to where routes can be defined.


trolologuy (approved):
- [x] The [final built page](https://pr-2108-evwlsqa-652soceglkw4u.eu-3.platformsh.site/configuration/routes/https.html) is working as expected.

PR #2111, opened 2022-01-27
Update redis.md

### LINKED ISSUE #1841: Using Redis with WordPress : Documentation outdated
Hello,

I'm following this doc to add Redis to a Wordpress installation : https://github.com/platformsh/platformsh-docs/blob/master/docs/src/guides/wordpress/redis.md 

It seems that the package advised is old (no commit since 2019) and fetches an outdated version of the Redis Object Cache Drop-In (1.3.4), while being incompatible with newest updates of the Redis Object Cache plugin.

It was a great alternative for my website which lacks write access. I think this page should be updated or deleted, but I wanted to know if there is another package that does the same job, or another way to accomplish this ?

Thanks in advance,
Regards

gilzow:
Updates to add instructions for installing both available plugins

## Why

Closes #1841 

## What's changed

Added instructions on how to set up and configure the plugins WP-Redis and Redis Object Cache, updated version of redis from 5.0 to 6.0,

### DIFF
--- a/docs/src/guides/wordpress/redis.md
+++ b/docs/src/guides/wordpress/redis.md
@@ -6,66 +6,115 @@ description: |
     Configure Redis for your WordPress site.
 ---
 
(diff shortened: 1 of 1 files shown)


CollierCZ (approved):
Thanks for the improvement, @gilzow! I made some minor styling changes to fit with our new content model, but I appreciate your finding the time to test out the different options.

PR #2117, opened 2022-02-02
:memo: Clarify restriction on root location

CollierCZ:
## Why

Make it clear that the `root` can't be set to the app root to avoid conflicts with dot files and the like.
[Additional context](https://github.com/orgs/platformsh/projects/3)

## What's changed

Added a sentence of explanation, though left out the reason as it seemed maybe extraneous and the space in the table is small.

### DIFF
--- a/docs/src/configuration/app/app-reference.md
+++ b/docs/src/configuration/app/app-reference.md
@@ -252,7 +252,7 @@ The following table presents possible properties for each location:
 
 | Name                | Type                                                 | Default   | Description |
 | ------------------- | ---------------------------------------------------- | --------- | ----------- |
-| `root`              | `string`                                             |           | The directory to serve static assets for this location relative to the app root. Must be an actual directory and not just the app root. |
+| `root`              | `string`                                             |           | The directory to serve static assets for this location relative to the app's root directory. Must be an actual directory inside the root directory. |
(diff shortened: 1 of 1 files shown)


docs/src/configuration/app/app-reference.md
```diff
@@ -252,7 +252,7 @@ The following table presents possible properties for each location:
 
 | Name                | Type                                                 | Default   | Description |
 | ------------------- | ---------------------------------------------------- | --------- | ----------- |
-| `root`              | `string`                                             |           | The directory to serve static assets for this location relative to the app root. |
+| `root`              | `string`                                             |           | The directory to serve static assets for this location relative to the app root. Must be an actual directory and not just the app root. |
```
trolologuy, line 255:
```suggestion
| `root`              | `string`                                             |           | The directory to serve static assets for this location relative to the app root. Must be an actual directory and not just the app's root folder. |
```

CollierCZ: Great idea for a change! I wouldn't change only the second instance of it. So I'll change the others.

trolologuy (approved):
- [x] The [final built page](https://pr-2117-iv55gyq-652soceglkw4u.eu-3.platformsh.site/configuration/app/app-reference.html#locations) is working as expected.

PR #2124, opened 2022-02-03
:memo: Add webp in the list of extensions

trolologuy:
## Why

webp is not listed and is often a source of questions on the public slack channel.

## What's changed

Added it to the list of extensions.

### DIFF
--- a/docs/src/languages/php/extensions.md
+++ b/docs/src/languages/php/extensions.md
@@ -119,6 +119,7 @@ and turn off those on by default with the `disabled_extensions` key.
 | `tidy`            | Avail | Avail | Avail | Avail | Avail | Avail | Avail | Avail | Avail | Avail |
 | `tokenizer`       |       |       |       | Def   | Def   | Def   | Def   | Def   | Def   | Def   |
 | `uuid`            |       |       |       |       | Avail | Avail | Avail | Avail | Avail | Avail |
+| `webp`            |       |       |       |       |       |       |       |       | Avail | Avail |
 | `wddx`            |       |       |       | Avail | Avail | Avail | Avail | Avail |       |       |
 | `xdebug`          |       |       |       |       | Avail | Avail | Avail | Avail | Avail | Avail |
 | `xcache`          | Avail | Avail |       |       |       |       |       |       |       |       |


CollierCZ (changes requested):
As noted in the additional context, it's not a separate extension but part of `imagick`. Maybe a separate line next to `imagick`? Or asterisks and an explanation below? :thinking:

trolologuy:
> As noted in the additional context, it's not a separate extension but part of `imagick`. Maybe a separate line next to `imagick`? Or asterisks and an explanation below? 🤔

Hmm, asterisks might be confusing since they are always in the array globally for the default extensions.
Maybe going the same approach as `tideways` and `tideways-xhprof` might work to get the context right?

CollierCZ:
But `tideways-xhprof` is a separate extension: https://support.tideways.com/documentation/reference/tideways-xhprof/tideways-xhprof-extension.html 
I wouldn't list it as a separate extension unless it is.

CollierCZ (changes requested):
It could be a different symbol? But something other than implying they have to add something to the YAML that will be rejected or have no effect.

docs/src/languages/php/extensions.md
```diff
@@ -57,7 +57,7 @@ and turn off those on by default with the `disabled_extensions` key.
 | `http`            | Avail | Avail |       |       |       |       | Avail | Avail | Avail | Avail |
 | `iconv`           |       |       |       | Def   | Def   | Def   | Def   | Def   | Def   | Def   |
 | `igbinary`        |       |       |       | Avail | Avail | Avail | Avail | Avail | Avail | Avail |
-| `imagick`         | Avail | Avail | Avail | Avail | Avail | Avail | Avail | Avail | Avail | Avail |
+| `imagick`         | Avail | Avail | Avail | Avail | Avail | Avail | Avail | Avail | Avail[^1]  | Avail[^1]  |
```
CollierCZ, line 60:
```suggestion
| `imagick`         | Avail | Avail | Avail | Avail | Avail | Avail | Avail | Avail | Avail with<br />`webp`  | Avail with<br />`webp`  |
```

Or something of the sort?

docs/src/languages/php/extensions.md
```diff
@@ -173,3 +174,7 @@ work:
    ```ini
    extension=/app/spiffy.so
    ```
+
+## Exceptions
+
+[^1]: `webp` support is only available for PHP 8.0 and higher.
```
CollierCZ, line 180: This isn't needed anymore.
```suggestion
```

PR #2131, opened 2022-02-07
:memo: Add footnotes in the styleguide

trolologuy:
## Why

Document the usage of footnotes if we decide to use them in [2124](https://github.com/platformsh/platformsh-docs/pull/2124)


trolologuy:
That is indeed much better! OK to merge for me.

CollierCZ (approved):

PR #2133, opened 2022-02-08
:bug: Fix incorrect filename for relationships key in service docs

ajmalsiddiqui:
## What's changed

The services documentation incorrectly indicates that the `relationships` key should be present in the `.platform/services.yaml` file, when it should be present in the `.platform.app.yaml` file instead. This PR fixes that.

A correct example of using the `relationships` key can be found [here](https://docs.platform.sh/configuration/app.html#comprehensive-example).


CollierCZ (approved):
Thanks, @ajmalsiddiqui !

PR #2138, opened 2022-02-08
Link to section about MFA enablement for ent/elite

nickbennison:
Minor update with link


CollierCZ:
Will be fixed in #2140

PR #2145, opened 2022-02-10
:memo: Add activity log docs

### LINKED ISSUE #2012: Linkable log lines and detailed timing information

I'm not sure, we don't seem to have a section on Activity logs either through the API or the UI. 

The big change is that there's detailed timing information and that you can select from one line to the other clicking shift, like you generally do when sharing code. 

We may evaluate that it doesn't require new docs! I think the biggest gotcha will be for people to figure out they can use shift to select several lines :) 

### Additional context

The improvement is already live and can be tested in all projects and activity logs.

### LINKED ISSUE #2016: :bug: Resizing images doesn't work within codetabs

In the platform

Currently, we have this syntax for images:
```markdown
![alt text](/images/path "0.5")
```
Where the last number is about the sizing.

But this isn't used within codetabs. It should be investigated and fixed if it's easy enough.

### Additional context

_No response_

CollierCZ:
## Why

Closes #2012
Closes #2016

## What's changed

* Made render hooks work in code tabs, meaning images can be resized by title
* Restructured container logs docs, including making a reference table and outsourcing explanation of format
* Added activity logs docs
* Added icon short code to render console icons inline. Only included share icon so far, but can add as needed.


docs/src/development/logs.md
```diff
@@ -1,89 +1,127 @@
...
-Although the files in `/var/log` are writable, they should not be written to directly. Only write to it via standard logging mechanisms, such as your application's logging facility.  If your application has its own logging mechanism that should be written to a dedicated logs mount in your application.
+<--->
+---
+title=Using the CLI
+highlight=false
+file=none
+---
 
-All log files are trimmed to 100 MB automatically. But if you need to have complete logs, you can set up cron which will upload them to third-party storage. [Contextual Code](https://www.contextualcode.com/) made a [basic and well-described example](https://gitlab.com/contextualcode/platformsh-store-logs-at-s3) how to achieve it.
+1. Get a list of activities by running `platform act -e <ENVIRONMENT_NAME>`.
```
trolologuy, line 33: Maybe let's use the long version of the argument?
```suggestion
1. Get a list of activities by running `platform activity:list -e <ENVIRONMENT_NAME>`.
```

docs/src/development/logs.md
```diff
@@ -1,89 +1,127 @@
...
-## `app.log`
+When trying to identify and resolve failures, it can often help to have another person's perspective.
+Troubleshoot collaboratively by sharing the logs of specific activities.
 
-Any log messages generated by the application will be sent to this file.  That includes language errors such as PHP Errors, Warnings, and Notices, as well as uncaught exceptions.
+To share a log, open the log and click {{< icon share >}} **Copy URL**.
 
-## `cron.log`
+Share specific lines in the log on clicking the line number.
+Hold shift to select multiple lines.
```
trolologuy, line 49:
```suggestion
Hold <kbd>shift</kbd> to select multiple lines.
```
We might want to use the [kbd tag](https://www.w3schools.com/Tags/tag_kbd.asp) for readility? (even though it's against the markdown guidelines).

trolologuy (approved):
Great improvement over the current page! Besides a few comments, nothing to add.

- [x] The [final built page](https://pr-2145-ajjgq4q-652soceglkw4u.eu-3.platformsh.site/development/logs.html
) is working as expected.

docs/src/development/logs.md
CollierCZ: I'd go with the [Google recommendation](https://developers.google.com/style/ui-elements?hl=en#keyboard) for now until we establish a separate standard. They recommend backticks.

CollierCZ: And if we do switch, I'd change the default styling :see_no_evil:
Screenshot from 2022-02-14 09-37-53

CollierCZ:
Also, what do you think about the title? I realize now on looking at it that it might be confusing since we don't have all titles in the new form in the `Do thing` style. So it might look like it's only about logs that deal with access rather than how to access all logs.

docs/src/development/logs.md
trolologuy: Sure thing :)

trolologuy:
> Also, what do you think about the title? I realize now on looking at it that it might be confusing since we don't have all titles in the new form in the `Do thing` style. So it might look like it's only about logs that deal with access rather than how to access all logs.

The title is indeed a bit misleading on that part. Maybe a more generic title like `Rework log access docs` would do the trick?

trolologuy (approved):
So, for now, let's follow the Google recommendation for the keyboard shortcut, right?
Other than that ✅

CollierCZ:
I meant the title of the built page, not the title of the PR.

trolologuy:
fair point, than "Access _the_ logs" might be better?

CollierCZ:
That's a good idea. I might go with "your" in place of "the", but the determiner definitely works.

I'll add the tag in #2164

trolologuy (approved):
✅

PR #2160, opened 2022-02-11
typo: mariadb storage engine

suntropez:


CollierCZ (approved):
Thanks for the change, @suntropez!

PR #2161, opened 2022-02-11
🐛  Update support-related information

trolologuy:
## Why

Fixed some mistakes on the image.


trolologuy:
Should the svg be indented as it previously was the case? or can just totally overwrite it?

CollierCZ:
It seems to be indented properly now. Not a large difference either way.

It needs to be rebased or merged with the changes already in `main`.

CollierCZ (approved):

PR #2169, opened 2022-02-15
updating command to restart php

jww-sh:
## Why

Customers cannot use sv restart app and need to Kill PHP instead

### DIFF
--- a/docs/src/languages/php/tuning.md
+++ b/docs/src/languages/php/tuning.md
@@ -65,8 +65,8 @@ If you aren't using PHP 7.4, this is a good reason to upgrade.
 
 Note that the only way to clear the preload cache is by restarting PHP-FPM.
 PHP-FPM isn't restarted on every deployment automatically,
-so you might want to add that in a [`deploy hook`](../../configuration/app/hooks.md),
-such as by including `pkill -f php-fpm`.
+so you might want to add that in a [`deploy` hook](../../configuration/app/hooks.md),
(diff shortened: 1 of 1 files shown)


docs/src/languages/php/tuning.md
```diff
@@ -65,8 +65,8 @@ If you aren't using PHP 7.4, this is a good reason to upgrade.
 
 Note that the only way to clear the preload cache is by restarting PHP-FPM.
 PHP-FPM isn't restarted on every deployment automatically,
-so you might want to add that in a [`deploy` hook](../../configuration/app/hooks.md),
-such as by including `sv restart app`.
+so you might want to add that in a [`deploy hook`](../../configuration/app/hooks.md),
```
thomasdiluccio, line 68: The docs seem to mention the name of the hook in quotes: ``deploy` hook`.

jww-sh: yeah I made it look better so the whole link and codeblock cover `deploy hook` and not just `deploy`

CollierCZ: Links that have just code as their text aren't visibly obvious as links. If you look at the code with a link vs. the code without, there's not much of a distinction:
https://pr-2169-tyztlpq-652soceglkw4u.eu-3.platformsh.site/languages/php/tuning.html#enable-preloading
This is a design issue we hope to fix, but the "hook" shouldn't be part of the code.

CollierCZ (approved):

PR #2171, opened 2022-02-15
:bug: Get PR comments with links working

CollierCZ:
## Why

The workflow to add comments was falsely reporting a lack of an integration, ~~likely cause it wasn't waiting long enough to get a report~~. Because GitHub has different hashes for different things. 

## What's changed

- Added a looping check to see if a status is reported. Ends at 10 minutes to prevent running forever.
- Switched to the PR head commit for the hash
- Made link list into a checklist

### DIFF
--- a/.github/workflows/link-to-docs.yaml
+++ b/.github/workflows/link-to-docs.yaml
@@ -17,24 +17,30 @@ jobs:
           id: get_env_url
           run: |
             COMMIT_STATUS="pending"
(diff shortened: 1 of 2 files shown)


trolologuy (approved):
From the flow of the PR, it seems like it's working now.
I'm not sure how to exactly test the behavior in detail.

CollierCZ:
We keep doing what we're doing. That's how I got to this bug :slightly_smiling_face:

PR #2173, opened 2022-02-16
:bug: Run PR comment workflow only on opening

CollierCZ:
## Why

To keep PRs from getting cluttered with comments. Currently the links to the built site are run on every edit.

## What's changed

Changed to only run on open and reopen (so it can be manually triggered by closing and reopening).


trolologuy (approved):
I was going to suggest something like that, the inbox get spammed a lot otherwise. 
Nice addition!

PR #2182, opened 2022-02-21
:memo: Rework of the troubleshoot page

### LINKED ISSUE #1999: Troubleshooting breadcumbs

https://docs.platform.sh/development/troubleshoot.html

Add links to specific troubleshooting pages from general one. Make it clear to readers that we now have multiple troubleshooting pages and their specific problem might be on one of the other pages. Consider linking back to this page from that one -- just make the network of troubleshooting pages clear.

### Additional context

_No response_

trolologuy:
## Why

Closes #1999

## What's changed
 
More crosslinking between the troubleshooting pages.

### DIFF
--- a/docs/layouts/shortcodes/troubleshoot.md
+++ b/docs/layouts/shortcodes/troubleshoot.md
@@ -1,0 +1,1 @@
+For generic information regarding how to troubleshoot a project , see [Troubleshoot development](/development/troubleshoot.md).
--- a/docs/layouts/shortcodes/troubleshoot.md
+++ b/docs/layouts/shortcodes/troubleshoot.md
@@ -1,1 +1,1 @@
-For generic information regarding how to troubleshoot a project , see [Troubleshoot development](/development/troubleshoot.md).
+For more general information, see how to [troubleshoot development](/development/troubleshoot.md).
--- a/docs/src/development/troubleshoot.md
+++ b/docs/src/development/troubleshoot.md
@@ -5,7 +5,7 @@ weight: 16
 description: See some common solutions to issues you might run into in development.
 ---
 
-## Common tools
+## Common tasks
 
 ### Force a redeploy
 
@@ -49,37 +49,11 @@ The redeploy takes place after any scheduled activities (either *Running* or *Pe
 Despite the name, redeployment doesn't rerun the `deploy` hook, only the `post_deploy` hook.
 Both your `build` and `deploy` hooks are tied to individual commits in code.
 They're reused until another commit is pushed to the environment.
-See [more about hooks](../configuration/app/hooks/_index.md) and their reuse.
+See [more about hooks](/configuration/app/hooks/_index.md) and their reuse.
 
-To rerun the `build` and `deploy` hooks, [manually trigger a build](#manually-trigger-builds).
+To rerun the `build` and `deploy` hooks, [manually trigger a build](/configuration/app/hooks/_index.md#manually-trigger-builds).
 
 {{< /note >}}
-
-### Manually trigger builds
-
-To increase performance and keep applications the same across environments,
-Platform.sh reuses built applications if its code and build time configuration (variables and such) remain the same.
-
-There may be times where you want to force your application to be built again without changing its code,
-for example to test an issue in a build hook or when external dependencies change.
-To force a rebuild without changing the code,
-use an [environment variable](./variables/set-variables.md#create-environment-specific-variables).
-
-Assuming you want to do this for your `main` environment,
-first create a `REBUILD_DATE` environment variable:
-
-```bash
-platform variable:create -l environment -e main --prefix env: --name REBUILD_DATE --value "$(date)" --visible-build true
-```
-
-This triggers a build right away to propagate the variable.
-To force a rebuild at any time, update the variable with a new value:
-
-```bash
-platform variable:update -e main --value "$(date)" "env:REBUILD_DATE"
-```
-
-This forces your application to be built even if no code has changed.
 
 ### Clear the build cache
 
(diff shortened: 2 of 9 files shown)


CollierCZ:
Test comment

docs/layouts/shortcodes/troubleshoot.md
```diff
@@ -0,0 +1 @@
+For generic information regarding how to troubleshoot a project , see [Troubleshoot development](/development/troubleshoot.md).
```
CollierCZ, line 1: I wouldn't use the word "generic" when not necessary. It's not really very specific.

I'd also prefer not to have links as titles rather than just parts of the sentence. I feel like it's confusing.

So I'd suggestion something along these lines (but really feel free to adapt it -- at least fix the extra space):
```suggestion
For more general information, see how to [troubleshoot development](/development/troubleshoot.md).
```

docs/src/development/troubleshoot.md
```diff
@@ -49,7 +49,7 @@ The redeploy takes place after any scheduled activities (either *Running* or *Pe
 Despite the name, redeployment doesn't rerun the `deploy` hook, only the `post_deploy` hook.
 Both your `build` and `deploy` hooks are tied to individual commits in code.
 They're reused until another commit is pushed to the environment.
-See [more about hooks](../configuration/app/hooks/_index.md) and their reuse.
+See [more about hooks](/configuration/app/hooks/_index.md) and their reuse.
```
CollierCZ, line 52: Why this change?

docs/src/development/troubleshoot.md
```diff
@@ -128,23 +128,31 @@ To send large files, use the `multipart/form-data` header instead:
 $ curl -XPOST 'https://example.com/graphql' --header 'Content-Type: multipart/form-data' -F file=large_file.json
 ```
 
-## Permission error creating a database
+## Databases
+
+For MySQL specific errors, see [troubleshoot MySQL](../configuration/services/mysql/troubleshoot.md).
```
CollierCZ, line 133: As above: see how to

docs/src/development/troubleshoot.md
```diff
@@ -128,23 +128,31 @@ To send large files, use the `multipart/form-data` header instead:
...
+### Permission error creating a database
 
 If you try to use a user to create a database, you get an error saying `permission denied to create database`.
 The database is created for you
 and can be found in the `path` key of the `$PLATFORM_RELATIONSHIPS` [environment variable](./variables/use-variables.md#use-platformsh-provided-variables).
 
-## Can't write to file system
+## Storage
+
+If you're having trouble with storage, you can [troubleshoot mounts](../configuration/app/troubleshoot-mounts.md) and [disks](../configuration/app/troubleshoot-disks.md).
```
CollierCZ, line 143: I'd be consistent here and replace "you can" with "see how to".

docs/src/development/troubleshoot.md
```diff
@@ -128,23 +128,31 @@ To send large files, use the `multipart/form-data` header instead:
...
+
+### Can't write to file system
 
 If you attempt to write to disk outside a `build` hook, you may encounter a `read-only file system` error.
 Except where you define it, the file system is all read-only, with code changes necessary through git.
 This gives you benefits like repeatable deployments, consistent backups, and traceability.
 
-You can write to disk a `build` hook to generate anything you need later.
+You can write to disk a [`build` hook to generate anything you need later](../configuration/app/app-reference.md#writable-directories-during-build).
 Or you can declare writable [mounts](../configuration/app/app-reference.md#mounts#mounts), which are writable even during and after deploy.
```
CollierCZ, line 152: I think this is too much unrelated info inside a link -- it doesn't really describe what the link is about. Maybe something like:
```suggestion
To generate anything you need later, [write to disk during a `build` hook](../configuration/app/app-reference.md#writable-directories-during-build).
Or [declare mounts](../configuration/app/app-reference.md#mounts#mounts),
which are writable even during and after deploy.
```

docs/src/development/troubleshoot.md
```diff
@@ -175,7 +183,7 @@ If the result is still `running`, the build is stuck.
 In most regions, stuck builds terminate after one hour.
 In older regions (`us` and `eu`), create a [support ticket](https://console.platform.sh/-/users/~/tickets/open) to have the build killed.
 
-When a _deployment_ is blocked, you should try the following:
+When a *deployment* is blocked, you should try the following:
```
CollierCZ, line 186: Why this change? This actually goes against the Markdownlint rule I'd adopt (I'd go with underscore for emphasis and asterisks for bold).

docs/src/development/troubleshoot.md
```diff
@@ -197,7 +205,7 @@ Invisible errors during the build and deploy phase can cause increased wait time
 ### Build and deploy hooks
 
 [`build` and `deploy` hooks](/configuration/app/hooks/_index.md) can cause long build times.
-If they run into issues, they can cause the build to fail or hang indefinitely.
+If they [run into issues](../configuration/app/app-reference.md#hook-failure), they can cause the build to fail or hang indefinitely.
```
CollierCZ, line 208: I don't think that the link actually adds information to the context here. It's mostly about how to cause them to fail, which is the opposite of helpful here. The issues there are separate from the ones here.

docs/src/development/troubleshoot.md
```diff
@@ -218,3 +226,10 @@ That means long-running cron jobs block a container from being shut down to make
 
 Make sure your custom cron jobs run quickly and properly.
 Cron jobs may invoke other services in unexpected ways, which can increase execution time.
+
+## Language-specific troubleshooting
+
+For more in depth troubleshooting regarding app-related issues:
```
CollierCZ, line 232:
```suggestion
For language-specific troubleshooting for your apps:
```

CollierCZ (changes requested):
As I mentioned, I think the improvement to the structure on the general troubleshooting page is great. An improvement I would have skipped over. The transclusion is also a good touch.

I have some style points, but only the things like the extra space are totally necessary.

- [ ] The information is concise.
- [ ] The changes follow the content and markup style guides.

docs/src/development/troubleshoot.md
trolologuy: I was using the [standard markdownlint](https://github.com/DavidAnson/markdownlint/blob/v0.25.1/doc/Rules.md#md049), which triggered here. What default settings are you using besides underscore for emphasis and asterisks for bold?

docs/src/development/troubleshoot.md
trolologuy: That is very correct indeed, the link only adds more confusion! Thanks for spotting it, I removed it.

trolologuy (commented):
A few open questions left. Applied the proposed recommendations.

docs/src/development/troubleshoot.md
CollierCZ: I wouldn't personally make changes based on the standard settings without agreeing on the standard for here. :shrug:

CollierCZ: But I think it's not a big deal either way. Was just curious because I've used a different standard before.

CollierCZ (changes requested):
Revert the white space changes and then it seems good to go

CollierCZ (approved):

PR #2185, opened 2022-02-21
:fire: Remove outdated Symfony templates

CollierCZ:
## Why

The linked Symfony templates are archived and shouldn't be included.

## What's changed

Removed the outdated templates


fabpot (commented):
Let's do it in #2187 as removing the templates is not enough as they are referenced IIUC in the docs.

PR #2187, opened 2022-02-21
Symfony on Platform.sh docs update

fabpot:
This PR removes everything related to the old Symfony templates and redirects the user to the official docs for the Symfony integration.

PR #2188, opened 2022-02-21
:bug: Fix bugs in PR comment workflow

CollierCZ:
## Why

* PRs run from forks (like #2187) don't have permissions currently to comment on the PR.
* The check for no changed files wasn't catching when files weren't changed
* The links to the deployed environment weren't using HTTPS

## What's changed

- Added explicit permissions for job (if it doesn't work, I can try `pull_request_target`, but it didn't work immediately
- Check for the filler text instead of an empty string when no docs are actually changed.
- Switched to HTTPS

### DIFF
--- a/.github/workflows/link-to-docs.yaml
+++ b/.github/workflows/link-to-docs.yaml
@@ -39,6 +39,7 @@ jobs:
                 sleep 30
 
                 ENV_URL=$(curl -s https://api.github.com/repos/$GITHUB_REPOSITORY/statuses/$COMMIT_SHA  | jq -r '.[0].target_url')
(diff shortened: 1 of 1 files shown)


CollierCZ:
> Your Platform.sh environment has successfully deployed. :rocket:
> 
> See the site: http://pr-2188-gvw6ffy-652soceglkw4u.eu-3.platformsh.site/

This is how it should look without any docs changes, @trolologuy

trolologuy (approved):
Since the latest re-open triggered the comment and the PR is working as expected, it's probably safe to assume that it's working. ✅

CollierCZ:
Not sure about the forking one, but at least this didn't break the main flow :grin:

PR #2191, opened 2022-02-22
:bug: Split workflows for comments on forked PRs

CollierCZ:
## Why

PRs run from forks (like https://github.com/platformsh/platformsh-docs/pull/2187) don't have permissions currently to comment on the PR.

## What's changed

* Split into a separate workflow run from the base repository (rather than the forked one) so it has sufficient permissions to comment. (Inspired by https://github.com/nf-core/tools/issues/765)
* Started using a sticky comment action so there is only one comment on a PR but it gets updated when changes are pushed. So if new pages are added later, they'll be added to the list.


CollierCZ:
After trying again and again, I realized the workflow with the comment only runs from the code on the default branch. So the only way to fully test this is to merge it.

trolologuy (approved):
The change in behavior sounds meaningful. Let's test if it works as expected.

CollierCZ:
After the various other bug fixes, it now comments even on PRs from forks: https://github.com/platformsh/platformsh-docs/pull/2209#issuecomment-1052128821 :rocket:

PR #2193, opened 2022-02-23
Fix query parameter on Google SSO URL

mrkschan:
It should be `prompt=?` rather than the other way round ...


CollierCZ (approved):
Thanks for the fix, @mrkschan!

PR #2195, opened 2022-02-23
:bug: Download artifact from other workflow

CollierCZ:
## Why

Official GitHub action doesn't allow downloading from a separate workflow (rather than the main one).

## What's changed

Use a different action.


trolologuy (approved):
Let's try [the alternative](https://github.com/dawidd6/action-download-artifact)

PR #2196, opened 2022-02-23
:bug: Fix location of downloaded files

CollierCZ:
## Why

Files seem to have been unzipped to the root directory, unlike the docs which said they'd default to a directory with the name of the artifact.

## What's changed

Try from the root.


trolologuy (approved):
Let's 🧪

PR #2204, opened 2022-02-24
:memo: Update firewall docs

CollierCZ:
## Why

To clarify status and add examples for domain rules. See additional context for request.

## What's changed

Updated docs and added examples.

### DIFF
--- a/docs/src/configuration/app/app-reference.md
+++ b/docs/src/configuration/app/app-reference.md
@@ -395,7 +395,7 @@ Each rule has the following properties where at least one is required and `ips` 
 | --------- | ------------------- | --------------- | ----------- |
 | `ips`     | Array of `string`s  | `["0.0.0.0/0"]` | IP addresses in [CIDR notation](https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing). See a [CIDR format converter](https://ipaddressguide.com/cidr). |
 | `domains` | Array of `string`s  |                 | [Fully qualified domain names](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) to specify specific destinations by hostname. |
-| `ports`   | Array of `integer`s |                 | Ports from 1 to 65535 that are allowed. If specified, unspecified ports are blocked. If no ports are specified, all ports are allowed. Port `25`, the SMTP port for sending email, is always blocked. |
+| `ports`   | Array of `integer`s |                 | Ports from 1 to 65535 that are allowed. If any ports are specified, all unspecified ports are blocked. If no ports are specified, all ports are allowed. Port `25`, the SMTP port for sending email, is always blocked. |
 
 The default settings would look like this:
 
--- a/docs/src/configuration/app/app-reference.md
+++ b/docs/src/configuration/app/app-reference.md
@@ -442,7 +442,7 @@ You can filter outbound traffic by domain.
 Using domains in your rules rather than IP addresses is generally more specific and secure.
 For example, if you use an IP address for a service with a CDN,
 you have to allow the IP address for the CDN.
-THis means that you allowing potentially hundreds or thousands of servers also using the CDN.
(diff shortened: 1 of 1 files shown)


docs/src/configuration/app/app-reference.md
```diff
@@ -402,7 +395,7 @@ Each rule has the following properties where at least one is required and `ips`
 | --------- | ------------------- | --------------- | ----------- |
 | `ips`     | Array of `string`s  | `["0.0.0.0/0"]` | IP addresses in [CIDR notation](https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing). See a [CIDR format converter](https://ipaddressguide.com/cidr). |
 | `domains` | Array of `string`s  |                 | [Fully qualified domain names](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) to specify specific destinations by hostname. |
-| `ports`   | Array of `integer`s |                 | Ports from 1 to 65535 that are allowed. If specified, unspecified ports are blocked. If no ports are specified, all ports are allowed. |
+| `ports`   | Array of `integer`s |                 | Ports from 1 to 65535 that are allowed. If specified, unspecified ports are blocked. If no ports are specified, all ports are allowed. Port `25`, the SMTP port for sending email, is always blocked. |
```
trolologuy, line 398:
```suggestion
| `ports`   | Array of `integer`s |                 | Ports from 1 to 65535 that are allowed. If ports are specified, all unspecified ports are blocked. If no ports are specified, all ports are allowed. Port `25`, the SMTP port for sending email, is always blocked. |
```

trolologuy (approved):
- [x] If anything has been moved or removed, there are clear redirects to the new location or related content.

docs/src/configuration/app/app-reference.md
```diff
@@ -435,6 +436,51 @@ For most CDNs, routing is done via domain name, not IP address,
 so thousands of domain names may share the same public IP addresses at the CDN.
 If you allow the IP address of a CDN, you are usually allowing many or all of the other customers hosted behind that CDN.
 
+### Outbound traffic by domain
+
+You can filter outbound traffic by domain.
+Using domains in your rules rather than IP addresses is generally more specific and secure.
+For example, if you use an IP address for a service with a CDN,
+you have to allow the IP address for the CDN.
+THis means that you allowing potentially hundreds or thousands of servers also using the CDN.
```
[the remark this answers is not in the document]
CollierCZ: I'd keep the adverb after "allow" because it's definitely allowing them, it's just a question of number.

docs/src/configuration/app/app-reference.md
CollierCZ: It would more technically be "If this property is specified" and I don't think that adds a lot of clarity, so I'd prefer the shorter one, unless you can think of some ambiguity that people might actually take from it?

trolologuy: It took me a few times to re-read the sentence to be able to understand its meaning. 

```suggestion
| `ports`   | Array of `integer`s |                 | If specified, unspecified ports are blocked. If no ports are specified, all ports are allowed. By default, ports from 1 to 65535 are allowed with the exception of the always blocked port `25` (SMTP port for sending email). |
```

CollierCZ: This way is more confusing because it puts what can be specified there after it describes the consequences. Leading with what can be there seems better to me. I feel like the original is common phrasing in docs in general, including ours. Like here: https://github.com/platformsh/platformsh-docs/blob/main/contributing/markup-format.md?plain=1#L274

But I can change it.

trolologuy: okey yeah, that makes sense 🤔 
Thanks for adding the additional words in the latest push, it seems clearer at the first reading occurrence of that sentence which might help non-native speakers.

trolologuy (approved):
✅ 🚀

PR #2210, opened 2022-02-28
:memo: Update availability wording

CollierCZ:
## Why

See additional context

## What's changed

Removed "only" and changed preposition


trolologuy (commented):
Where exactly are the partials being loaded?

CollierCZ:
You can search for them. The specific example from the additional context: https://docs.platform.sh/security/waf.html

PR #2211, opened 2022-02-28
Update network-storage page, re: legacy regions

gilzow:
## Why

Indicate the regions not available are legacy, remove the list of suggested regions as those are also out-of-date

[Additional Context](https://github.com/orgs/platformsh/projects/3/views/1)

## What's changed
Noted the two regions where the service is not available are legacy. Removed the list of suggested regions where a legacy region should be migrated to.


CollierCZ (approved):
Thanks, @gilzow!

PR #2225, opened 2022-03-15
:bug: Only report deployed env if successful

CollierCZ:
## Why

The step `Comment without links` was running even when deployment wasn't successful (see comment on #2224). This meant overriding the failure message.

## What's changed

Added a condition to only run that step on success.


trolologuy (approved):
Let's try that ✅

PR #2226, opened 2022-03-16
:memo: Remove examples of manual SSH URI building

CollierCZ:
## Why

The recommendation to build a SSH URI manually led to confusion. (See additional context.) The CLI should handle the selection of the right app.

## What's changed

Changed examples to use the CLI for getting the URI.

### DIFF
--- a/docs/src/configuration/services/rabbitmq.md
+++ b/docs/src/configuration/services/rabbitmq.md
@@ -66,36 +66,51 @@ highlight=python
 
 {{< /codetabs >}}
 
-(The specific way to inject configuration into your application will vary. Consult your application or framework's documentation.)
+(The specific way to inject configuration into your application varies.
+Consult your application or framework's documentation.)
 
 ## Connecting to RabbitMQ
 
 ### From your local development environment
 
-For debugging purposes, it's sometimes useful to be able to directly connect to a service instance. You can do this using SSH tunneling. To open a tunnel, log into your application container like usual, but with an extra flag to enable local port forwarding:
+For debugging purposes, it's sometimes useful to connect directly to a service with an SSH tunnel.
+To open a tunnel, follow these steps.
 
-```bash
-ssh -L 5672:rabbitmqqueue.internal:5672 <projectid>-<branch_ID>@ssh.eu.platform.sh
-```
+1. Get the username and password to the service by checking your app's relationships:
 
-Within that SSH session, use the following command to pretty-print your relationships. This lets you see which username and password to use, and you can double check that the remote service's port is 5672.
+   ```bash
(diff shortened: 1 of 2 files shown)


docs/src/configuration/services/rabbitmq.md
```diff
@@ -66,36 +66,51 @@ highlight=python
...
 ### From your local development environment
 
-For debugging purposes, it's sometimes useful to be able to directly connect to a service instance. You can do this using SSH tunneling. To open a tunnel, log into your application container like usual, but with an extra flag to enable local port forwarding:
+For debugging purposes, it's sometimes useful to connect directly to a service with an SSH tunnel.
+To open a tunnel, follow these steps.
 
-```bash
-ssh -L 5672:rabbitmqqueue.internal:5672 <projectid>-<branch_ID>@ssh.eu.platform.sh
-```
+1. Get the username and password to the service by checking your app's relationships:
```
pjcdawkins, line 79: Hm I haven't tried specifically for rabbitmq but I think tunnel:single should output relationship information like this on its own

docs/src/configuration/services/rabbitmq.md
```diff
@@ -66,36 +66,51 @@ highlight=python
...
+
+2. Within that SSH session, get the username and password.
+   Use the following command to see your app's relationships:
+
+   ```bash
+   echo "$PLATFORM_RELATIONSHIPS" | base64 --decode | jq .
+   ```
 
-After you successfully established a connection, you should be able to open `http://localhost:15672` in your browser. You'll find the credentials like mentioned above.
+3. Open `http://localhost:15672` in your browser.
```
pjcdawkins, line 112: We should test for rabbitmq specifically, but this again is the same as using the tunnel:single command

docs/src/configuration/services/rabbitmq.md
CollierCZ: All I get is the following:

```bash
SSH tunnel opened to rabbitmq at: amqp://guest:guest@127.0.0.1:30000

Quitting this command (with Ctrl+C or equivalent) will close the tunnel.
```

CollierCZ: I missed that in the provided URI. :facepalm:

docs/src/configuration/services/rabbitmq.md
CollierCZ: The tunnel:single command seems to create a amqp connection string, but it didn't seem to create a link to the web UI. I didn't see a flag that would work for that.

trolologuy (approved):
Indeed it's much clearer! Works as expected. 🚀 ✅ 

PR #2227, opened 2022-03-16
Add file path to configure Bootsnap

spk:
## Why

Help user not used to configure bootsnap with the path

## What's changed

Add the path to the metadata of the markdown


CollierCZ (approved):
Thanks for the change, @spk !

PR #2231, opened 2022-03-18
:bug: Fix links to home from search

CollierCZ:
## Why

Links from search to the home page had `/.html` in them, causing a 404 error.

## What's changed

Replaced that string with just `/`.


trolologuy (approved):
Indeed clicking the first link in https://docs.platform.sh/search.html?q=git%20driven results in a 404 and the [pr version](https://pr-2231-6kwgyyy-652soceglkw4u.eu-3.platformsh.site/search.html?q=git%20driven) works.

PR #2234, opened 2022-03-21
Provide a .environment/rsync warning

chadwcarlson:
## Why

No existing issue open.

I've done this plenty of times, but never run into the issue before because I was not running `platform mount:download`. 

1. You want to track if some conditional thing has occurred on an environment. 
2. You put that logic in `.environment`.
3. To sanity check/troubleshoot, you add some feedback/output to that file with an `echo` command so you can more easily track whats going on. 

That's all fine, until you try to run `platform mount:download`. `rsync` doesn't expect the output coming from your `echo` command, resulting in the following error:

```bash
$ platform mount:download -A nextjs -e pr-1 -m deploy --target . -y

Downloading files from the remote mount deploy to .

Are you sure you want to continue? [Y/n] y

  protocol version mismatch -- is your shell clean?
  (see the rsync man page for an explanation)
  rsync error: protocol incompatibility (code 2) at /System/Volumes/Data/SWE/macOS/BuildRoots/220e8a1b79/Library/Caches/com.apple.xbs/Sources/rsync/rsync-55/rsync/compat.c(61) [receiver=2.6.9]

                                                                                                                                                
  [ProcessFailedException]                                                                                                                      
  The command failed with the exit code: 2                                                                                                      
                                                                                                                                                
  Full command: 'rsync' '--archive' '--compress' '--human-readable' '-v' '65o73exizvbli-pr-1-djjnuwy--nextjs@ssh.eu-3.platform.sh:deploy/' '.' 
```

Totally caused by the fact that this `.environment`:

```bash
if [ -f "deploy/environment.tracker.txt" ]; then 
    echo "File found."
    export DEPLOY='Friday'
else
    echo "File not found."
    export DEPLOY='Never on Friday'
```

prints `File found/not found` at the start of an SSH session. 

[Source](https://dev-maziarz.blogspot.com/2015/01/rsync-error-protocol-incompatibility.html)

## What's changed

Added a note to [`.environment` section](https://docs.platform.sh/development/variables/set-variables.html#set-variables-via-script) that 

1. reminds the user that they're dealing in dash, not bash.
2. recommends they don't do what I described, including the error so its searchable. (The dash addition seemed relevant since it's another time adding an `echo` is useful.

### DIFF
--- a/docs/src/development/variables/set-variables.md
+++ b/docs/src/development/variables/set-variables.md
@@ -131,8 +131,47 @@ to be run regardless of the current directory:
 export PATH=/app/vendor/bin:$PATH
 ```
 
+Like everything else, the sourced `.environment` file runs in dash (not bash), and you can include logic that dynamically defines environment variables based on the current environment.
+
+```bash {location=".environment"}
+URL=$(echo $PLATFORM_ROUTES | base64 --decode | jq -r 'to_entries[] | select (.value.id == "api") | .key')
+```
+
+In the above example, `jq` is used to retrieve a backend Drupal application's url for the current environment using it's `id` that has been defined in `routes.yaml`. 
+
 Note that the file is sourced after all other environment variables above are defined and so they're available to the script.
 This also means the `.environment` script has the last word on environment variable values and can override anything it wants to.
+
+{{< note >}}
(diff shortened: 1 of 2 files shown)


CollierCZ (approved):
Thanks for the addition, @chadwcarlson !

Issue #2213, opened 2022-03-02
:bug: Broken links


CollierCZ:
The following pages had broken links to check and possibly fix:

- https://docs.platform.sh/increase-observability/integrate-observability/blackfire.html
  - [ ] 403 (following redirect https://support.blackfire.platform.sh/hc/en-us): https://support.blackfire.io
- https://docs.platform.sh/guides/typo3/deploy/next-steps.html
  - [ ] 429 (following redirect https://github.com/login?return_to=https%3A%2F%2Fgithub.com%2Fplatformsh%2Fplatformsh-docs%2Fedit%2Fmain%2Fdocs%2Fsrc%2Fguides%2Ftypo3%2Fdeploy%2Fnext-steps.md): https://github.com/platformsh/platformsh-docs/edit/main/docs/src/guides/typo3/deploy/next-steps.md
- https://docs.platform.sh/create-apps/web/static.html
  - [ ] 429 (following redirect https://github.com/login?return_to=https%3A%2F%2Fgithub.com%2Fplatformsh%2Fplatformsh-docs%2Fedit%2Fmain%2Fdocs%2Fsrc%2Fcreate-apps%2Fweb%2Fstatic.md): https://github.com/platformsh/platformsh-docs/edit/main/docs/src/create-apps/web/static.md
- https://docs.platform.sh/dedicated-gen-2/architecture.html
  - [ ] 429 (following redirect https://github.com/login?return_to=https%3A%2F%2Fgithub.com%2Fplatformsh%2Fplatformsh-docs%2Fedit%2Fmain%2Fdocs%2Fsrc%2Fdedicated-gen-2%2Farchitecture%2F_index.md): https://github.com/platformsh/platformsh-docs/edit/main/docs/src/dedicated-gen-2/architecture/_index.md
- https://docs.platform.sh/guides/ibexa/deploy.html
  - [ ] 429 (following redirect https://github.com/login?return_to=https%3A%2F%2Fgithub.com%2Fplatformsh%2Fplatformsh-docs%2Fedit%2Fmain%2Fdocs%2Fsrc%2Fguides%2Fibexa%2Fdeploy.md): https://github.com/platformsh/platformsh-docs/edit/main/docs/src/guides/ibexa/deploy.md
- https://docs.platform.sh/guides/symfony.html
[...]

PR #2246, opened 2022-03-28
:bug: Fix broken link

CollierCZ:
## Why

From #2213

## What's changed

Fixed second instance of broken link that was missed last time.


trolologuy (approved):

PR #2249, opened 2022-03-29
:memo: Add PLATFORM_CACHE_DIR variable

CollierCZ:
## Why

The `PLATFORM_CACHE_DIR` environment variable wasn't in our list of variables, though it was elsewhere.

## What's changed

Added it to the list.

### DIFF
--- a/docs/src/development/variables/use-variables.md
+++ b/docs/src/development/variables/use-variables.md
@@ -346,8 +346,9 @@ You can't set or update them directly.
 The most important of these variables is the relationship information in `PLATFORM_RELATIONSHIPS`,
 which tells the app how to connect to databases and other services defined in `services.yaml`.
 
-The following table presents the available variables
-and whether they're available during builds and at runtime.
+The following table presents all available variables
+and whether they're available at build time (during [build hooks](../../administration/../configuration/app/hooks/hooks-comparison.md#build-hook))
+and at runtime.
 
 | Variable name               | Build | Runtime | Description |
 | --------------------------- | ----- | ------- | ----------- |


docs/src/development/variables/use-variables.md
```diff
@@ -355,6 +355,7 @@ and whether they're available during builds and at runtime.
 | `PLATFORM_APPLICATION`      | Yes   | Yes     | A base64-encoded JSON object that describes the app. It maps certain attributes from your [app configuration](../../configuration/app/_index.md), some with more structure. See [notes](#platform_application). |
 | `PLATFORM_APPLICATION_NAME` | Yes   | Yes     | The app name as set in your [app configuration](../../configuration/app/_index.md). |
 | `PLATFORM_BRANCH`           | No    | Yes     | The name of the Git branch. |
+| `PLATFORM_CACHE_DIR`        | Yes   | No      | The directory where files are cached from one build to the next. The directory is share among all branches, so the same cache is used for all environments. |
```
trolologuy, line 358:
```suggestion
| `PLATFORM_CACHE_DIR`        | Yes   | No      | The directory where files are cached from one build to the next. The directory is shared among all branches, so the same cache is used for all environments. |
```

trolologuy: Would it make sense to mention the [4GB Limit in the table only](https://pr-2249-m2nrk7q-652soceglkw4u.eu-3.platformsh.site/configuration/app/troubleshoot-disks.html#no-space-left-on-device)?
Since the table works as a reference and probably will get more frequent updates, it might be better than on the "no space left on device" troubleshoot part?

trolologuy: Would it also make sense to make the `Build` headline of the tables point to [the build hook page](https://pr-2249-m2nrk7q-652soceglkw4u.eu-3.platformsh.site/configuration/app/hooks/hooks-comparison.html#build-hook)

trolologuy (changes requested):
Small typo to fix + a few thoughts.

docs/src/development/variables/use-variables.md
CollierCZ: - I don't think mentioning the limit is appropriate here because that is a more general limit, not limited to this specific instance. This table is a reference for variables, not for build and other containers, so i don't think it'd get updated more frequently for the relevant ideas.
- I don't think the heading of tables is a good place for adding context. I added some before, even though I think it might be something of a distraction for some people. Could be helpful, though.

CollierCZ: Typo is already fixed :slightly_smiling_face:

trolologuy (approved):
Thanks for adding the additional context, despite not being in scope, I think it's helpful ❤️

PR #2254, opened 2022-03-30
:memo: Add system for review comments

CollierCZ:
## Why

To make review comments more organized and clearer about where important issues are.

## What's changed

Added a system based on [conventional comments](https://conventionalcomments.org/) using emoji (similar to [other systems](https://daniel.town/github,/code/review/2017/09/08/emoji-code-review.html)).

It's an open proposal, so feel free to make suggestions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -135,25 +135,30 @@ Otherwise, we make suggestions and work with you to finalize the changes.
 ### Review comment style
 
 To make comments clearer, each comment should have an emoji label describing its purpose.
-The following labels are used:
+The following labels are primarily used:
 
-| Emoji | `:code:`              | Meaning      | Description |
-| ----- | --------------------- | ------------ | ----------- |
-| 👍/😍 | `:+1:`/`:heart_eyes:` | Praise       | Used when something is well done. No further action required. Should be at least one of these for each review. |
-| ⛏️    | `:pick:`              | Nitpick      | For small details based on preferences. Doesn't block merging. |
-| 🛠️    | `:hammer_and_wrench:` | Suggestion   | To indicate an idea for improvement. By default, the suggestion needs to be accepted or otherwise addressed before merging. |
-| 💅    | `:nail_car:`          | Polish       | To indicate an idea for improvement that isn't fixing something wrong, but just pointing out ways to improve quality. Doesn't block merging by default. |
-| 📋    | `:clipboard:`         | To-do         | For small, necessary changes, such as fixing typos. |
-| 🐞    | `:lady_beetle:`       | Issue        | To highlight a specific issue that needs to be fixed. Can be paired with a suggestion if a solution is known. Blocks merging by default. |
-| ❓    | `:question:`          | Question     | For potential concerns that may not be relevant or for areas that aren't completely clear. Requires a response by default. |
-| 💡    | `:bulb:`              | Thought      | To introduce an idea that came up from reviewing. Doesn't block merging by default, but can lead to more discussion. |
+| Emoji | `:code:`              | Meaning      | Can merge  | Description |
+| ----- | --------------------- | ------------ | ---------- | ----------- |
+| 😍    | `:heart_eyes:`        | Praise       | ▶️          | Used when something is well done. No further action required. Should be at least one of these for each review. |
+| ⛏️    | `:pick:`              | Nitpick      | ▶️          | For small details based on preferences. |
+| 🛠️    | `:hammer_and_wrench:` | Suggestion   | 🚫         | To indicate a specific idea for improvement. By default, the suggestion needs to be accepted or otherwise addressed before merging. |
+| 💅    | `:nail_car:`          | Polish       | ▶️          | To indicate a specific idea for improvement that isn't fixing something wrong, but just pointing out ways to improve quality. |
+| 📋    | `:clipboard:`         | To-do        | 🚫         | For small, necessary changes, such as fixing typos. |
+| 🐞    | `:lady_beetle:`       | Issue        | 🚫         | To highlight a specific issue that needs to be fixed. Can be paired with a suggestion if a solution is known. |
+| ❓    | `:question:`          | Question     | 🚫         | For potential concerns that may not be relevant or for areas that aren't completely clear. By default, requires a response. |
+
+The following labels are also possible:
+
+| Emoji | `:code:`              | Meaning      | Can merge  | Description |
+| ----- | --------------------- | ------------ | ---------- | ----------- |
+| 💡    | `:bulb:`              | Thought      | To introduce an idea that came up from reviewing. More general than a suggestion and not focused on the details. Doesn't block merging by default, but can lead to more discussion. |
 | 🧹    | `:broom:`             | Chore        | For small process tasks that need to be done before merging. |
 
 The following decorations can be added for further clarification (to override default of blocking or not):
 
 | Emoji | `:code:`          | Meaning      | Description |
 | ----- | ----------------- | ------------ | ----------- |
-| ⚠️     | `:warning:`       | Blocking     | Used to indicate that the comment blocks merging. |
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -129,6 +130,34 @@ We generally review for:
...
+### Review comment style
+
+To make comments clearer, each comment should have an emoji label describing its purpose.
+The following labels are used:
+
+| Emoji | `:code:`              | Meaning      | Description |
+| ----- | --------------------- | ------------ | ----------- |
+| 👍/😍 | `:+1:`/`:heart_eyes:` | Praise       | Used when something is well done. No further action required. Should be at least one of these for each review. |
+| ⛏️    | `:pick:`              | Nitpick      | For small details based on preferences. Doesn't block merging. |
+| 🛠️    | `:hammer_and_wrench:` | Suggestion   | To indicate an idea for improvement. By default, the suggestion needs to be accepted or otherwise addressed before merging. |
```
trolologuy, line 144: ❓ 💡 
As discussed during the weekly, maybe we could merge that one with the 💡 .
and add stuff like 🚫 or  ⚠️ for the blockers.
As you pointed out, it sounds "negative". 

Alternatively, we might just use the ❓ with any other emoji, to mark it as blocking?
So ❓ 💡  would be a blocking idea that needs discussion.
What do you think?

CONTRIBUTING.md
```diff
@@ -129,6 +130,34 @@ We generally review for:
...
+| 💅    | `:nail_car:`          | Polish       | To indicate an idea for improvement that isn't fixing something wrong, but just pointing out ways to improve quality. Doesn't block merging by default. |
+| 📋    | `:clipboard:`         | To-do         | For small, necessary changes, such as fixing typos. |
```
trolologuy, line 146: ❓ 
Could we merge that one with the 🧹 ? it feels a bit redundant.

CONTRIBUTING.md
```diff
@@ -129,6 +130,34 @@ We generally review for:
...
+
+To make comments clearer, each comment should have an emoji label describing its purpose.
+The following labels are used:
+
+| Emoji | `:code:`              | Meaning      | Description |
+| ----- | --------------------- | ------------ | ----------- |
+| 👍/😍 | `:+1:`/`:heart_eyes:` | Praise       | Used when something is well done. No further action required. Should be at least one of these for each review. |
+| ⛏️    | `:pick:`              | Nitpick      | For small details based on preferences. Doesn't block merging. |
+| 🛠️    | `:hammer_and_wrench:` | Suggestion   | To indicate an idea for improvement. By default, the suggestion needs to be accepted or otherwise addressed before merging. |
+| 💅    | `:nail_car:`          | Polish       | To indicate an idea for improvement that isn't fixing something wrong, but just pointing out ways to improve quality. Doesn't block merging by default. |
```
trolologuy, line 145: ❓ 
What's the difference between the 💡 ? Is it just for look and feel?

trolologuy (dismissed):
Great idea, it's going to be helpful to prioritize the actionable items out of a review.
We could simplify the list (see comments and propositions).
We can also merge it already and iterate through improvements as we go.

CONTRIBUTING.md
CollierCZ: I also think it's important that :bulb: represents a thought rather than an idea. An idea would be classified more as a suggestion.

CONTRIBUTING.md
CollierCZ: Example of a chore (assuming you thought the first draft of #2259 was perfect) would be to rerun the check for the PR so it could pass and merge. (I didn't bother to get the check working, cause I figured I'd have to change it anyway)

CONTRIBUTING.md
```diff
@@ -129,6 +130,39 @@ We generally review for:
...
+| ----- | --------------------- | ------------ | ---------- | ----------- |
+| 💡    | `:bulb:`              | Thought      | To introduce an idea that came up from reviewing. More general than a suggestion and not focused on the details. Doesn't block merging by default, but can lead to more discussion. |
+| 🧹    | `:broom:`             | Chore        | For small process tasks that need to be done before merging. |
+
+The following decorations can be added for further clarification (to override default of blocking or not):
+
+| Emoji | `:code:`          | Meaning      | Description |
+| ----- | ----------------- | ------------ | ----------- |
+| 🚫    | `:no_entry_sign:` | Blocking     | Used to indicate that the comment blocks merging. |
+| ▶️     | `:arrow_forward:` | Non-blocking | Used to indicate that the comment doesn't block merging. |
```
trolologuy, line 162: :hammer_and_wrench: I think using the non-ascii character might help to help clarify what's blocking from what isn't.

```suggestion
| ✅     | `:white_check_mark:` | Non-blocking | Used to indicate that the comment doesn't block merging. |
```

CONTRIBUTING.md
```diff
@@ -129,6 +130,39 @@ We generally review for:
...
+| 💅    | `:nail_car:`          | Polish       | ▶️          | To indicate a specific idea for improvement that isn't fixing something wrong, but just pointing out ways to improve quality. |
+| 📋    | `:clipboard:`         | To-do        | 🚫         | For small, necessary changes, such as fixing typos. |
+| 🐞    | `:lady_beetle:`       | Issue        | 🚫         | To highlight a specific issue that needs to be fixed. Can be paired with a suggestion if a solution is known. |
+| ❓    | `:question:`          | Question     | 🚫         | For potential concerns that may not be relevant or for areas that aren't completely clear. By default, requires a response. |
+
+The following labels are also possible:
+
+| Emoji | `:code:`              | Meaning      | Can merge  | Description |
+| ----- | --------------------- | ------------ | ---------- | ----------- |
+| 💡    | `:bulb:`              | Thought      | To introduce an idea that came up from reviewing. More general than a suggestion and not focused on the details. Doesn't block merging by default, but can lead to more discussion. |
```
trolologuy, line 154: :lady_beetle: The table is missing a column: the "can merge" column isn't present.

trolologuy (changes requested):
✅ The additional answers you provided helped clarify the logic and reasoning behind the table and the logic of the review system.

I think the table as it is currently should (hopefully) be clear enough. 
It's probably more important to officialize the changes in the commenting approach (as you proposed) and iterate with changes if further clarification is needed.

CONTRIBUTING.md
CollierCZ: To me, a check mark looks like it's something that has been resolved and therefore doesn't require any thought at all. I feel like that's different than a non-blocking comment, which can be skipped but isn't necessarily resolved yet. I think there might be something better than an arrow, so if you have other ideas I'd welcome them, but I'd probably avoid a check mark.

I'm not sure I get the non-ASCII part of the comment. The ASCII character would be →, no?

trolologuy: My phrasing was confusing, I assume I meant that the emoji version would be better than the triangle.
As propositions:
-  🆗 since it's ok to proceed 
-  ➡️ Because the flow is not disrupted
-  ⏸ since it's not "pausing" the process

trolologuy (approved):
A few propositions in the comments. 

After some usage of the emojis for the reviews, I think the explanation you provided do make sense, if not we can always further iterate the page to either broaden or narrow down the scope for each of them.

CONTRIBUTING.md
CollierCZ: The triangle _is_ the emoji version, so I'm still not sure what the comment should mean.

The last suggestion is amusing to me since it's the opposite of how I think of it. This isn't pausing the process, so I wouldn't use pause here, but it could make sense for the blocking comment (and is thus less negative there). That way pause = block and play = continue.

CollierCZ: To clarify, the triangle is the emoji for the play button: https://emojipedia.org/play-button/ GitHub just gives it another name.

trolologuy: I like the idea of pause and play 🤔  So if you like it too, we can go with that?

PR #2279, opened 2022-04-08
Update Drupal Solr guide

raupie:


CollierCZ (approved):
Thanks for the improvement, @raupie!

PR #2287, opened 2022-04-14
:white_check_mark: Lint prose on forked PRs

CollierCZ:
## Why

The Vale action only had permissions to annotate PRs from non-forked repos.

## What's changed

Switched to run after the **Get PR info** workflow and override the GitHub SHA.
Based on https://github.com/errata-ai/vale-action/issues/58, this should work similar to our other PR comments


trolologuy (approved):
Let's test that one ✅

PR #2296, opened 2022-04-22
Fix typo in app configuration doc

alexandra-zaharia:
## Why

Fix typo


CollierCZ (approved):
Thanks for the fix, @alexandra-zaharia!

PR #2298, opened 2022-04-23
Use explicit curl in Windows

shyim:
## Why

`curl` is an alias to Invoke-WebRequest on Windows. When you really want curl, you have to add the `.exe`


CollierCZ (approved):
Thanks for the improvement, @shyim!

PR #2303, opened 2022-04-26
📝 Rework email docs

### LINKED ISSUE #1994: Update email docs

https://docs.platform.sh/development/email.html

Rename as "Send email"

Update for:

* Content accuracy
* Includes what it should
* Doesn't include what doesn't fit
* Fits template
* Fits style guide

### Additional context

Check with support if there's anything to add about running into limits (abuse checks)

CollierCZ:
## Why

Closes #1994

## What's changed

Turned the email page into a how-to and tried to simplify it and make it fit the style guide.


CollierCZ:
There probably should be something added about limits and such. Not sure if it would fit in this how-to guide, though. Maybe another page? Not sure how much detail would be helpful though. :thinking:

docs/src/development/email.md
```diff
@@ -92,61 +44,113 @@ file=none
...
 
-To disallow emails, replace `true` with `false`.
+To turn off outgoing email, replace `true` with `false`.
 
 {{< /codetabs >}}
 
 Changing the setting rebuilds the environment.
 
-## SMTP host address
+## 2. (Optional) Add email domain validation
```
trolologuy, line 59: :hammer_and_wrench: :arrow_forward:

Maybe we could:
- rename that section to `Improve deliverability`
- start off with the SPF part and then get to the Elite-only DKIM section?

That would make sure that non-elite customers don't skip the SPF part?

docs/src/development/email.md
```diff
@@ -208,3 +212,11 @@ Guides on using JavaMail:
 
 - [Send email with HTML formatting and attachments](https://mkyong.com/java/java-how-to-send-email/)
 - [JavaMail API](https://javaee.github.io/javamail/)
+
+{{< /codetabs >}}
+
+## Alternative: Set up your own email server
```
trolologuy, line 218: :hammer_and_wrench: :arrow_forward:
Since that section isn't only about the "own" mail server, maybe let's make it more generic?
```suggestion
## Alternative: Set up an external email server
```

trolologuy (approved):

docs/src/development/email.md
CollierCZ: That was how I had it initially, but I thought it made sense to prioritize the premium offer. Also so it didn't seem like people who do DKIM should also do the other option. But I can try with different headings now.

docs/src/development/email.md
CollierCZ: That's a fair point, but I don't think the server has to be external, either. They can run one on the infrastructure, as far as I know, just on a different port. I could try `Use a different email server`, although it's less clear, it is more comprehensive

docs/src/development/email.md
CollierCZ: Huh. That wasn't my impression, and I definitely didn't get that from your previous suggestion. I will switch it to better match that idea.

trolologuy: Sorry, I should've been more precise in my explanation. The flow is much clearer now 😍

trolologuy (approved):
✅ 🚀

PR #2305, opened 2022-04-27
🐛 Use Vale version that recognizes workflow runs

CollierCZ:
## Why

The changes in #2287 haven't been released in a version of the Vale action, so the [workflow run failed](https://github.com/platformsh/platformsh-docs/runs/6179626912?check_suite_focus=true).

## What's changed

Switched to a commit that uses the change that includes the fix #2287 relied on.


trolologuy (approved):
✅

PR #2307, opened 2022-04-28
🐛 Use PR SHA for Vale

CollierCZ:
## Why

I missed the use of the SHA instead of the PR number :facepalm:. It's possible even this won't work, but it might :shrug:

## What's changed

Added the PR SHA for the Vale workflow


trolologuy (approved):
🚢 ✅

PR #2310, opened 2022-04-28
✨ Add shortcut for create project links

### LINKED ISSUE #2260: :sparkles: Improve links to project creation

Everywhere that has links to creating a new project, such as: 
https://docs.platform.sh/development/local/lando.html

- Define a universal way to use the link (making sure people get context in the console, and aren't just dumped into a list they don't understand)
- Define if analytics should be included (channel, language, and so on) and if so, which
- Document the approach chosen
- Add it everywhere

See #2240 for more context (especially the related context on #1995)

### Additional context

_No response_

CollierCZ:
## Why

Closes #2260

## What's changed

Created a shortcut with documented behavior
Added UTM parameters in other links


docs/themes/avocadocs/layouts/_default/_markup/render-link.html
```diff
@@ -4,7 +4,7 @@
   {{- $isRemote := in $remoteSchemes $url.Scheme -}}
   {{- $isFile := strings.HasSuffix .Destination "pdf" }}
   {{- if or $isRemote $isFile -}}
-    <a href="{{ .Destination }}" target="_blank" rel="noopener">{{ .Text | safeHTML }}</a>
+    <a href="{{ htmlUnescape .Destination }}" target="_blank" rel="noopener">{{ .Text | safeHTML }}</a>
```
trolologuy, line 7: :question: What is this file/part doing?

docs/src/development/local/ddev.md
```diff
@@ -0,0 +1,204 @@
+---
+title: Use DDEV for local development
+sidebarTitle: DDEV
+weight: 1
+---
+
+[DDEV](https://ddev.readthedocs.io/en/stable/) is an open-source tool to get local PHP development environments.
+It allows you use Docker in your workflows without needing
```
trolologuy, line 8: :clipboard: This sentence is missing a part. What isn't it needing?

docs/src/development/local/ddev.md
```diff
@@ -0,0 +1,204 @@
...
+It allows you to have fully containerized environments to run everything locally
+without having to install tools (such as the Platform.sh CLI) on your machine.
+It's maintained by a community of developers and is a viable option for most Platform.sh projects.
+
+This guide assumes you have a project already running within Platform.sh.
+If you're starting from scratch, first [create a project from a template](https://console.platform.sh/org/create-project/template).
+
+## Before you begin
+
+Make sure your computer meets the [system requirements for Docksal](https://docs.docksal.io/getting-started/setup/#system-requirements).
```
trolologuy, line 21: :lady_beetle: This page is about DDEV, why are we referring to Docksal and Docksal commands throughout the document?
Was this file added by mistake? I don't see clearly the link between the scope of the PR and this file.

trolologuy (changes requested):
😍  Thank you for the follow-up on the tracking aspects of things! It's great to have aligned behavior with the buttons now!

⛏️  On the following pages, we could add the button as well:
1. https://docs.platform.sh/gettingstarted/introduction/template/create-project.html
2. https://docs.platform.sh/guides/general/region-migration.html#1-create-a-new-project
3. https://docs.platform.sh/guides/strapi/adding-frontends.html#signing-up-for-platformsh -> If we add the button, the steps described will be the same right? (it should propose a free trial I assume)
4. https://docs.platform.sh/tutorials/migrating.html#preparation

❓ We could add the link to the [button creation page](https://docs.platform.sh/guides/general/deploy-button.html#defined-template) in our [markup format](https://github.com/platformsh/platformsh-docs/blob/main/contributing/markup-format.md) and add the `utm` format in the markup page, since we now have a specific way of adding these buttons.

docs/src/development/local/ddev.md
CollierCZ: Yes, the file was added by mistake.

docs/themes/avocadocs/layouts/_default/_markup/render-link.html
CollierCZ: The link gets passed with an ampersand (`&`). This ensures it doesn't get encoded as `&amp;`, as it does otherwise.

CollierCZ:
I think by "button" you mean link? Because I actually took out the button shortcode since it wasn't in use anywhere. Anyway, about the possible places for it (though going through everything is a little out of scope):

1. This page is part of a series where the previous step created a new organization. They don't need a create-project link, since they're prompted to do it just by logging into the console. Plus it's part of get-started, which will change soon anyway.
2. Added. It's possible the UTM parameter will need to change, but for now using the same.
3. As with 1, if you create a new org, you don't need a link to create a project.
4.  As with 2.

The UTM parameters are for tracking from a specific location (the docs) and for our purposes. The button creation page is for buttons elsewhere and maybe for other people. The link won't work elsewhere and the parameters would have to be changed for other locations.

trolologuy (approved):
Thanks for the additional context on the previous comment/review.
Should be good to go, now 😍 

PR #2338, opened 2022-05-10
:bug: Fix Varnish relationship name

trolologuy:
## Why

The relationship name was wrong.

## What's changed

Updated varnish relationship in the registry.
We need to check if the main [varnish page](https://github.com/platformsh/platformsh-docs/blob/main/docs/src/add-services/varnish.md) needs further tweaking, for example, the line 61.


CollierCZ (changes requested):
Lines 61 and 65 in the varnish.md file should definitely change if the relationship name changes in the example.

And while you're in there, please update the line
```
The file name is relative to the `.platform` directory.
```
to include it being within the directory (as noted in the context). It's kind of out of scope, but it's another bug quick fix that's in the same file.

PR #2342, opened 2022-05-10
📝 Recommend Network Storage 2.0 even on Grid

pjcdawkins:
## Why

I'm told that network-storage 2.0 works on Grid. I've tried it, minimally, and this appears to be the case.

It would be painful to have to change versions between Grid and Dedicated Gen 3, so, having the same version on both would be highly preferable.

## What's changed

Versions in `registry.json` and the text describing versions on `/configuration/services/network-storage.html#supported-versions`

### DIFF
--- a/docs/data/registry.json
+++ b/docs/data/registry.json
@@ -559,9 +559,11 @@
     "runtime": false,
     "type": "network-storage",
     "versions": {
-      "deprecated": [],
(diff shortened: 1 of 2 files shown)


CollierCZ (approved):
Thanks for the change, @pjcdawkins

PR #2349, opened 2022-05-11
📝 Rework multi-app docs

CollierCZ:
## Why

Closes #1929
Closes #1925

## What's changed

Clarified app root and used in various places
Restructured multi-app doc

PR #2350, opened 2022-05-11
:memo: Add missing link

trolologuy:
## Why

The link to contact the sales representative was missing.

## What's changed

Added the missing link


CollierCZ:
If this was going to be added, I'd do so in #2349. But I'm pretty sure the sales representative there means a specific account/success manager, rather than a generic message to all of Sales. So I wouldn't encourage the general message. Dedicated projects have dedicated people that are responsible for them.

If the idea is to contact generic support, that'd be different, but with sales, I'd think they should have a specific contact.

Feel free to reopen if you understand it differently.

PR #2351, opened 2022-05-11
Remove unneeded DBGp Proxy section, and mention that the default port is now 9003

derickr:
… is now 9003

- Xdebug 3's default port is 9003, and not 9000
- The DBGp proxy section is not needed, and wouldn't have worked anyway because the host was left empty.

## Why

Closes #{ISSUE_NUMBER}


lolautruche (approved):

CollierCZ (approved):
Thanks for the fixes, @derickr!

PR #2353, opened 2022-05-11
Updated HTTP Cache documentation.

markdorison:
## Why
When referencing the [HTTP cache documentation](https://docs.platform.sh/define-routes/cache.html), I was unaware of the [separate best practices doc](https://docs.platform.sh/bestpractices/http-caching.html) that contains valuable information. I wanted to surface some of that and provide a link on the main page.

## What's changed

Updated HTTP Cache documentation to include reference to HTTP caching best practices documentation.

## Alternative
The contents of the [HTTP Caching Best Practices](https://docs.platform.sh/bestpractices/http-caching.html) page could be moved completely to the [HTTP cache](https://docs.platform.sh/define-routes/cache.html) page.

### DIFF
--- a/docs/src/define-routes/cache.md
+++ b/docs/src/define-routes/cache.md
@@ -17,7 +17,9 @@ When caching is on...
 * cookies will bypass the cache;
 * responses with the `Cache-Control` header set to `Private`, `No-Cache`, or `No-Store` are not cached.
 
-You should _not_ use the Platform.sh HTTP cache if you are not using Varnish or an external CDN. Mixing cache services together will most likely result in stale and unclearable caches. For more details, see [Best Practices: HTTP caching](../bestpractices/http-caching.md).
+You should _not_ use the Platform.sh HTTP cache if you're using Varnish or an external CDN.
+Mixing cache services together most likely results in caches that are stale and can't be cleared.
+For more details, see [best practices on HTTP caching](../bestpractices/http-caching.md).
 
 ## Basic usage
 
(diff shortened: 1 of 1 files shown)


docs/src/define-routes/cache.md
```diff
@@ -17,6 +17,8 @@ When caching is on...
 * cookies will bypass the cache;
 * responses with the `Cache-Control` header set to `Private`, `No-Cache`, or `No-Store` are not cached.
 
+You should _not_ use the Platform.sh HTTP cache if you are not using Varnish or an external CDN. Mixing cache services together will most likely result in stale and unclearable caches. For more details, see [Best Practices: HTTP caching](../bestpractices/http-caching.md).
```
CollierCZ, line 20:
```suggestion
You should _not_ use the Platform.sh HTTP cache if you're using Varnish or an external CDN.
Mixing cache services together most likely results in caches that are stale and can't be cleared.
For more details, see [best practices on HTTP caching](../bestpractices/http-caching.md).
```
Small styling suggestions, plus I think the second "not" in the first sentence shouldn't be there.

CollierCZ (changes requested):
Thanks for the suggestion, @markdorison! Definitely a good idea to link the pages together.

I made a small suggestion to better fit our style.

As for the alternative, it might be best, but would require more effort now. We're planning changes to the information architecture and pages like this one, so I think we'll do something similar when the time for that comes.

CollierCZ (approved):

PR #2356, opened 2022-05-12
📝 Clarify name changes are destructive

### LINKED ISSUE #2331: 📝Clarify that name changes delete service data
We want to clarify for all services that changing their name deletes their data. Probably with a shortcode to make it standard across services.

CollierCZ:
## Why

Closes #2331

## What's changed

Added a note to the endpoint description. Decided against a `note` as it's an intro to the service and might feel too disruptive.

### DIFF
--- a/docs/layouts/shortcodes/endpoint-description.md
+++ b/docs/layouts/shortcodes/endpoint-description.md
@@ -35,8 +35,12 @@ Or the `redis-persistent` type for persistent Redis:
 
 Persistent Redis requires a disk to store data.
 {{ else if eq $type "network-storage" }}
-You can define `<SERVICE_NAME>` as you like, but it should not include underscores (`_`).
+You can define `<SERVICE_NAME>` as you like, but it shouldn't include underscores (`_`).
 {{ end }}
+
+Note that if you later change the name, it's treated as an entirely new service.
+This removes all data from your service.
+Always backup your data before changing the service.
 
 <!-- Extra text to explain service configuration -->
 {{ .Inner }}


docs/layouts/shortcodes/endpoint-description.md
```diff
@@ -35,9 +35,13 @@ Or the `redis-persistent` type for persistent Redis:
 
 Persistent Redis requires a disk to store data.
 {{ else if eq $type "network-storage" }}
-You can define `<SERVICE_NAME>` as you like, but it should not include underscores (`_`).
+You can define `<SERVICE_NAME>` as you like, but it shouldn't include underscores (`_`).
 {{ end }}
 
+Note that if you later change the name, it's treated as an entirely new service.
+This removes all data from your service.
+Always backup your data before changing the service.
```
trolologuy, line 43: :pick: :arrow_forward: 
```suggestion
Always backup your data before changing the service name.
```

docs/layouts/shortcodes/endpoint-description.md
```diff
@@ -35,9 +35,13 @@ Or the `redis-persistent` type for persistent Redis:
 
 Persistent Redis requires a disk to store data.
 {{ else if eq $type "network-storage" }}
-You can define `<SERVICE_NAME>` as you like, but it should not include underscores (`_`).
+You can define `<SERVICE_NAME>` as you like, but it shouldn't include underscores (`_`).
 {{ end }}
```
trolologuy, line 40: :hammer_and_wrench:

Although we do not want to over-use notes, I think there might be a good use case for that, given the potential for loss of data.
```suggestion
{{< note theme="warning">}} ... {{< /note >}}
```
It's probably being overly cautious since customers won't probably be re-checking the documentation before renaming a service, but if they remember there were some warnings on the first setup, it increases the likeliness that they'll re-open the page.

trolologuy (approved):
A few propositions, otherwise it's a clear change.

docs/layouts/shortcodes/endpoint-description.md
CollierCZ: I'd thought about that, but then I thought that it's true for any part of the service, so I left out `name` intentionally as good advice for any change. :shrug:

docs/layouts/shortcodes/endpoint-description.md
CollierCZ: I definitely considered a note, but decided against it. Partly because, as I mentioned, I think it's disruptive of the flow at a time when people are getting started with the service. I think it could create a worse experience -- more fears, less getting things done -- at a time when they are uncertain. I think a small note without too much attention is the right fit during that part of the task.

On the other hand, I've seen some evidence that years of ads being separate from text on websites have trained people to pay *less* attention to text in boxes like that. :thinking:

docs/layouts/shortcodes/endpoint-description.md
trolologuy: Fair point, better to be overly precautious indeed!

PR #2359, opened 2022-05-16
:memo: Update CDN pages

trolologuy:
## Why

Add a note regarding the risks of stale content when using two caching mechanisms.

## What's changed

Fixed typos, added missing information on how to disable internal cache, added links between the pages.

### DIFF
--- a/docs/layouts/shortcodes/disable-cache.md
+++ b/docs/layouts/shortcodes/disable-cache.md
@@ -1,9 +1,9 @@
 
-## Check the routes settings
+## Avoid double-caching
 
-To avoid stale content that can't be cleared, don't use {{ .Get "CDN" }} with HTTP caching.
+To avoid stale content that can't be cleared, don't use {{ .Get "CDN" }} with [HTTP caching](../../src/define-routes/cache.md).
 
-For routes where {{ .Get "CDN" }} is used, disable HTTP caching by setting the [`cache` key](../../src/define-routes/cache.md) to `false`:
+For routes where {{ .Get "CDN" }} is used, disable HTTP caching:
 
 ```yaml {location=".platform/routes.yaml"}
 https://{default}/:
--- a/docs/layouts/shortcodes/disable-cache.md
+++ b/docs/layouts/shortcodes/disable-cache.md
@@ -1,0 +1,14 @@
+
+## Check the routes settings
+
+To avoid stale content that can't be cleared, {{ .Get "CDN" }} shouldn't be used with the Platform.sh caching mechanism.
+
+For the routes where {{ .Get "CDN" }} is used, make sure that the Platform.sh HTTP Cache layer is disabled by setting the [`cache` key](../../src/define-routes/cache.md) to `false`:
+
+```yaml {location=".platform/routes.yaml"}
+https://{default}/:
+    type: upstream
+    ...
+    cache:
+        enabled: false
+```
--- a/docs/src/define-routes/cache.md
+++ b/docs/src/define-routes/cache.md
@@ -26,7 +26,7 @@ For more details, see [best practices on HTTP caching](../bestpractices/http-cac
 
 The HTTP cache is enabled by default, however you may wish to override this behavior.
 
-To configure the HTTP cache, add a `cache` key to your route, as in the default:
+To configure the HTTP cache, add a `cache` key to your route. You may like to start with the defaults:
 
 ```yaml {location=".platform/routes.yaml"}
 https://{default}/:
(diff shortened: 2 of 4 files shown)


docs/src/define-routes/cache.md
```diff
@@ -12,22 +12,22 @@ If a request is can be cached, Platform.sh builds a cache key from several reque
 
 When caching is on...
 
-* you can configure cache behaviour for different location blocks in your `.platform.app.yaml`;
+* you can configure cache behavior for different location blocks in your `.platform.app.yaml`;
```
CollierCZ, line 15: :+1: Good eye :us:

docs/src/define-routes/cache.md
```diff
@@ -12,22 +12,22 @@ If a request is can be cached, Platform.sh builds a cache key from several reque
...
 Mixing cache services together most likely results in caches that are stale and can't be cleared.
 For more details, see [best practices on HTTP caching](../bestpractices/http-caching.md).
 
 ## Basic usage
 
-The HTTP cache is enabled by default, however you may wish to override this behaviour.
+The HTTP cache is enabled by default, however you may wish to override this behavior.
 
-To configure the HTTP cache, add a `cache` key to your route in `.platform/routes.yaml`. You may like to start with the defaults:
+To configure the HTTP cache, add a `cache` key to your route:
```
CollierCZ, line 28: :question: Why'd you take out that it's the default? Is it not? If it is the default and the note isn't there, it seems like the example will do something when it might not.

docs/src/define-routes/cache.md
```diff
@@ -12,22 +12,22 @@ If a request is can be cached, Platform.sh builds a cache key from several reque
...
 
-```yaml
+```yaml {location=".platform/routes.yaml"}
```
CollierCZ, line 30: :+1: Yay consistency

docs/layouts/shortcodes/disable-cache.md
```diff
@@ -0,0 +1,14 @@
+
+## Check the routes settings
```
CollierCZ, line 2: :lady_beetle: This heading should be more specific. Right now it's about settings, which isn't something we want to focus on (settings aren't a user goal, just a means to achieve the goal). A heading focused on why they'd want to do the step would be better.

CollierCZ (changes requested):
A good change for consistency and clarity :+1: 

Some small thins to clear up before merging, but nothing complex.

- [ ] The information is comprehensive.
- [ ] The information is concise.

docs/layouts/shortcodes/disable-cache.md
trolologuy: I hope it's better now? I don't like the repetition of `avoid`, but it should do the trick.
`double-caching` is probably easier to understand as a headline. Or would you go with `Avoid stale content`?.

docs/src/define-routes/cache.md
```diff
@@ -12,22 +12,23 @@ If a request is can be cached, Platform.sh builds a cache key from several reque
...
-You should _not_ use the Platform.sh HTTP cache if you're using Varnish or an external CDN.
+You should _not_ use the Platform.sh HTTP cache if you're using Varnish or an external CDN
+such as [Fastly](../domains/cdn/fastly.md) or [Cloudflare](../domains/cdn/cloudflare.md).
 Mixing cache services together most likely results in caches that are stale and can't be cleared.
 For more details, see [best practices on HTTP caching](../bestpractices/http-caching.md).
 
 ## Basic usage
 
-The HTTP cache is enabled by default, however you may wish to override this behaviour.
+The HTTP cache is enabled by default, however you may wish to override this behavior.
```
trolologuy, line 27: :hammer_and_wrench:
Maybe even better that way? If we provide explanations and details on how to override the defaults... it's fairly clear that the behavior can be overriden.
```suggestion
The HTTP cache is enabled by default.
```

CollierCZ (approved):

docs/src/define-routes/cache.md
CollierCZ: It's not the only sub-optimal phrasing on the page :shrug: Probably better without that phrase, but I wouldn't spend too long thinking either way.

docs/src/define-routes/cache.md
```diff
@@ -12,22 +12,23 @@ If a request is can be cached, Platform.sh builds a cache key from several reque
...
 
-To configure the HTTP cache, add a `cache` key to your route in `.platform/routes.yaml`. You may like to start with the defaults:
+To configure the HTTP cache, add a `cache` key to your route, as in the default:
```
CollierCZ, line 29: :nail_care: I'm not sure about the `in the default` phrasing. It seems slightly less clear than the original was, I think partly because of the `in`. But I wouldn't worry about it too much.

CollierCZ (dismissed):

CollierCZ (approved):

Issue #2355, opened 2022-05-12
📝 Clarify how to use Varnish


CollierCZ:
### Where on docs.platform.sh should be changed?

https://docs.platform.sh/configuration/services/varnish.html
https://docs.platform.sh/bestpractices/http-caching.html

### What exactly should be updated?

People have been confused about how to set up Varnish and how to purge the cache (see context).

Rework the page so it's clear what to do with Platform.sh. Don't get into too many details about Varnish itself, which should be covered by Varnish docs.

Update for:

-  Content accuracy
-  Includes what it should
-  Doesn't include what doesn't fit
-  Fits template
-  Fits style guide

aikaterine:
It is critical here for you to be clear that your solution:

1. Opens up the possobility for any project hosted on the same server to clear the cache (call with external URL), or
2. Requires protecting the purging with ip addresses, and/or token

None of those are ideal and if a project wants the ability to purge uising Varnish (which, lets be honest, most do) then their options at platform.sh are problematic and burdensome (to be nice). Odds are that projects which require push based caching will need to incur additional expense with Cloudflare or Fastly to get what should be provided out of the box with platform. sh. 

Particularly galling is the line:

"Generally speaking, you should use Varnish only if your application requires push-based clearing or relies on Varnish-specific business logic."

You don't actually support that in a safe manner. It is misleading, at best.

CollierCZ:
Thanks for the feedback, @aikaterine. We'd definitely like to stop clear up any places that might be misleading.

The solution proposed internally so far seemed to be using a token. I haven't looked into the details yet myself (since I haven't gotten to the task yet). Do you have particular concerns that would make the token not a good route to take?

letharion:
The issue I see with a token is that it changes the implementation from being completely standard, as in, something you can copy from the docs, to entirely bespoke. That also appears to be the opposite of the desire of this ticket, as the description says:

> Don't get into too many details about Varnish itself, which should be covered by Varnish docs.

But this approach won't be covered by varnish docs. I haven't seen this done elsewhere, so integrations to varnish, in our case a Drupal-module, will need to be adapted in addition to the custom varnish logic be written. Overall it will make everything more complicated, for the "gain" of working around a limitation in the platform services specification.

CollierCZ:
[...]

Issue #2352, opened 2022-05-11
🐛 Fix circular relationship with Varnish


CollierCZ:
### Where on docs.platform.sh should be changed?

https://docs.platform.sh/configuration/services/varnish.html

### What exactly should be updated?

The stats endpoint creates a circular relationship, which breaks things. See context.

The main steps should no longer suggest creating a relationship with the main app and using the stats that way. The stats section can be optional to add another add that isn't user facing that monitors stats, but make it clear to avoid circularity.

PR #2361, opened 2022-05-17
:memo: Fix Varnish relationship name

trolologuy:
## Why

The relationship name was wrong.

## What's changed

Updated varnish relationship in the registry.
We need to check if the main [varnish page](https://github.com/platformsh/platformsh-docs/blob/main/docs/src/add-services/varnish.md) needs further tweaking, for example, the line 61.

### DIFF
--- a/docs/src/add-services/varnish.md
+++ b/docs/src/add-services/varnish.md
@@ -34,7 +34,7 @@ You can define `<RELATIONSHIP_NAME>` as you like.
 `<APP_NAME>` should match the name you gave your app in your [app configuration](../create-apps/app-reference.md).
 
 The `configuration` block must reference a VCL file (`config.vcl` in this example).
-The file name is relative to the `.platform` directory.
+The file name is relative to, and must be inside the `.platform` directory.
 
 {{% /endpoint-description %}}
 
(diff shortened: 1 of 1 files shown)


trolologuy:
I hope the changes are now correct.
Additional changes on that page are required.

Loosely related to the following issues:
- https://github.com/platformsh/platformsh-docs/issues/2352
- https://github.com/platformsh/platformsh-docs/issues/2355

docs/src/add-services/varnish.md
```diff
@@ -56,20 +56,20 @@ The VCL file you provide has three specific requirements over and above the VCL
...
 sub vcl_recv {
-    set req.backend_hint = application.backend();
+    set req.backend_hint = varnishstats.backend();
 }
 ```
 
-Where `application` is the name of the relationship defined in `services.yaml`.
-(If the relationship was named differently, use that name instead.)
+Where `varnishstats` is the name of the relationship defined in `services.yaml`.
+(If the relationship was named differently in `<RELATIONSHIP_NAME>`, use that name instead.)
```
CollierCZ, line 66: On further investigation, I think this change is not correct. The relationship here should be the one Varnish has to the app, which isn't defined as `varnishstats`. The `varnishstats` relationship only shows stats and doesn't do the things in this file.

It might be easier when the stats relationship is gone as part of #2352, so I might just suggest closing this one

CollierCZ (changes requested):

PR #2363, opened 2022-05-17
🛠 Update Regions Environmental impact section

SabriPlatform:
Add a website for real-time emissions and explain Platform.sh' use of static data

## Why

Update doc based on feedback regarding emissions data and external website

## What's changed
A more in-depth explanation of our data and also link to an external website

Add more information regarding emissions from third-party website and real-time info. Also, explain that the data we use is static and in table form.


CollierCZ (approved):
Thanks, @SabriPlatform

PR #2364, opened 2022-05-18
🐛 Remove suggested Varnish circular relationship

### LINKED ISSUE #2352: 🐛 Fix circular relationship with Varnish

CollierCZ:
## Why

Closes #2352

## What's changed

- Removed configuration with stats endpoint
- Tried to quickly clarify the need for a separate app to get stats
- Clarified link checker

### DIFF
--- a/docs/src/add-services/varnish.md
+++ b/docs/src/add-services/varnish.md
@@ -33,8 +33,8 @@ The `relationships` block allows Varnish to talk to your app.
 You can define `<RELATIONSHIP_NAME>` as you like.
 `<APP_NAME>` should match the name you gave your app in your [app configuration](../create-apps/app-reference.md).
 
-The `configuration` block must reference a VCL file (`config.vcl` in this example).
-The file must be inside the `.platform` directory and is defined relative to that directory.
+The `configuration` block must reference a VCL file inside the `.platform` directory.
+The `path` defines the file relative to the `.platform` directory.
 
 {{% /endpoint-description %}}
 
@@ -63,9 +63,16 @@ sub vcl_recv {
 ```
 
 Where `<RELATIONSHIP_NAME>` is the name of the relationship you defined in [step 1](#1-configure-the-service).
+With the [example configuration](#example-configuration), that would be the following:
+
+```bash {location=".platform/config.vcl"}
+sub vcl_recv {
+    set req.backend_hint = application.backend();
+}
+```
 
 If you have multiple applications fronted by the same Varnish instance,
-then you will need to include logic to determine to which application a request is forwarded.
+then you need to include logic to determine to which application a request is forwarded.
 For example:
 
 ```yaml {location=".platform/services.yaml"}
@@ -111,7 +118,7 @@ You also need to disable the router cache as it is now entirely redundant with V
 
 For example:
 
-{{< readFile file="src/registry/images/examples/full/varnish.routes.yaml" highlight="yaml" >}}
+{{< readFile file="src/registry/images/examples/full/varnish.routes.yaml" highlight="yaml" location=".platform/routes.yaml" >}}
 
 That will map all incoming requests to the Varnish service rather than the application.
 Varnish will then, based on the VCL file, forward requests to the application as appropriate.
@@ -131,7 +138,7 @@ Platform.sh supports a number of optional modules you can include in your VCLs, 
 
 To use in your VCL, add an import such as:
 
-```bash
+```bash {location=".platform/config.vcl"}
 import xkey;
 ```
 
@@ -149,11 +156,22 @@ which provides access to some Varnish analysis and debugging tools.
 
 You can't use it from an app fronted by Varnish because of the restriction with [circular relationships](#circular-relationships).
 To access the stats, create a **separate app** with a relationship *to* Varnish, but not *from* it.
-Add the following to your [app configuration](../create-apps/app-reference.md):
+Define an [app configuration](../create-apps/app-reference.md) similar to the following:
 
-{{< readFile file="src/registry/images/examples/full/varnish.app.yaml" highlight="yaml" location=".platform.app.yaml" >}}
+```yaml {location=".platform.app.yaml"}
+name: statsApp
+type: "php:8.1"
 
-You can then access the `varnishstats` relationship over HTTP at the following paths to get diagnostic information:
+build:
+    flavor: none
+
+relationships:
+    varnishstats: "varnish:http+stats"
+```
+
+You choose any valid name and type.
+When the app is deployed, the app can access the Varnish service over HTTP to get diagnostic information.
+The following paths are available:
 
 * `/`: returns the error if generating the VCL failed with an error
 * `/config`: returns the generated VCL
@@ -162,6 +180,7 @@ You can then access the `varnishstats` relationship over HTTP at the following p
 
 To access the Varnish stats endpoint from the command line:
 
-1. Connect to your cluster [using SSH](../development/ssh/_index.md): `platform ssh -p <PROJECT_ID>`,
+1. Connect to your stats app [using SSH](../development/ssh/_index.md): `platform ssh --app statsApp`
+   (replace `statsApp` with the name you gave the app).
 2. Display the [relationships array](../create-apps/app-reference.md#relationships) with `echo $PLATFORM_RELATIONSHIPS | base64 -d | jq .`,
 3. Query Varnish with `curl <HOST>:<PORT>/stats`, replacing `<HOST>` and `<PATH>` with the values from step 2.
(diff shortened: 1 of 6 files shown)


docs/src/add-services/varnish.md
```diff
@@ -146,8 +144,12 @@ This limitation may be lifted in the future.
...
-The Varnish service also offers an `http+stats` endpoint, which provides access to some Varnish analysis and debugging tools.
-To access it, from a dedicated app container add the following to your [app configuration](../create-apps/app-reference.md):
+The Varnish service also offers an `http+stats` endpoint,
+which provides access to some Varnish analysis and debugging tools.
+
+You can't use it from an app fronted by Varnish because of the restriction with [circular relationships](#circular-relationships).
+To access the stats, create a **separate app** with a relationship *to* Varnish, but not *from* it.
+Add the following to your [app configuration](../create-apps/app-reference.md):
 
 {{< readFile file="src/registry/images/examples/full/varnish.app.yaml" highlight="yaml" location=".platform.app.yaml" >}}
```
trolologuy, line 154: :hammer_and_wrench: Why not provide a working example to make the page easier to use?
Something along the lines of:
```
name: varnishstats

type: "php:8.1"

build:
    flavor: none

relationships:
    varnishstats: "varnish:http+stats"
```

docs/src/add-services/varnish.md
```diff
@@ -158,10 +160,8 @@ You can then access the `varnishstats` relationship over HTTP at the following p
...
 
-Because of the circular relationship noted above, this can't be done on the application that Varnish is forwarding to.
-It needs to be run on a separate application container.
+To access the Varnish stats endpoint from the command line:
 
-To access the Varnish endpoint:
-- Connect to your cluster [using ssh](/development/ssh/_index.md) or through the CLI: `platform ssh -p <project id>`,
-- Display the [relationships array](../create-apps/app-reference.md#relationships) with `echo $PLATFORM_RELATIONSHIPS | base64 -d | jq '.'`,
-- Query Varnish with `curl varnishstats.internal:8081/stats`, for example, to access the statistics directly. Be sure to update the request according to the name of the relationship.
+1. Connect to your cluster [using SSH](../development/ssh/_index.md): `platform ssh -p <PROJECT_ID>`,
```
trolologuy, line 165: :pick:
```suggestion
1. Connect to your cluster [using SSH](../development/ssh/_index.md): `platform ssh -p <PROJECT_ID> -A <APPNAME>`,
```

where `<APPNAME>` is the name of the second `.platform.app.yaml`, so from the previous comment that could be `varnishstats`.

docs/src/add-services/varnish.md
```diff
@@ -146,8 +144,12 @@ This limitation may be lifted in the future.
 
 ## Stats endpoint
 
-The Varnish service also offers an `http+stats` endpoint, which provides access to some Varnish analysis and debugging tools.
-To access it, from a dedicated app container add the following to your [app configuration](../create-apps/app-reference.md):
+The Varnish service also offers an `http+stats` endpoint,
+which provides access to some Varnish analysis and debugging tools.
+
+You can't use it from an app fronted by Varnish because of the restriction with [circular relationships](#circular-relationships).
+To access the stats, create a **separate app** with a relationship *to* Varnish, but not *from* it.
```
trolologuy, line 151: :question:
Wouldn't it require a change in the `.platform/routes.yaml` as well?

docs/src/add-services/varnish.md
```diff
@@ -56,20 +56,19 @@ The VCL file you provide has three specific requirements over and above the VCL
...
+```bash {location=".platform/config.vcl"}
 sub vcl_recv {
-    set req.backend_hint = application.backend();
+    set req.backend_hint = <RELATIONSHIP_NAME>.backend();
 }
 ```
 
-Where `application` is the name of the relationship defined in `services.yaml`.
-(If the relationship was named differently, use that name instead.)
+Where `<RELATIONSHIP_NAME>` is the name of the relationship you defined in [step 1](#1-configure-the-service).
```
trolologuy, line 65: :pick:
An copy-pastable example based on the example we provided above could be nice.

docs/src/add-services/varnish.md
```diff
@@ -56,20 +56,19 @@ The VCL file you provide has three specific requirements over and above the VCL
...
 
 If you have multiple applications fronted by the same Varnish instance,
```
trolologuy, line 67: :pick: That could be a subpoint to make the flow easier.

docs/src/add-services/varnish.md
```diff
@@ -146,8 +144,12 @@ This limitation may be lifted in the future.
 
 ## Stats endpoint
```
trolologuy, line 145: :pick: Shouldn't that be an optional step of the flow? (Step 4)

trolologuy (changes requested):
Nice work there! The example is now working 😍 . However, I could not get the [stats endpoint](https://pr-2364-fi72ivy-652soceglkw4u.eu-3.platformsh.site/add-services/varnish.html#stats-endpoint) to work elsewhere than through ssh.
Some parts of the page might be clearer (see comments).

- [ ] The information is clear.

docs/src/add-services/varnish.md
```diff
@@ -34,11 +34,11 @@ You can define `<RELATIONSHIP_NAME>` as you like.
 `<APP_NAME>` should match the name you gave your app in your [app configuration](../create-apps/app-reference.md).
 
 The `configuration` block must reference a VCL file (`config.vcl` in this example).
-The file name is relative to the `.platform` directory.
+The file must be inside the `.platform` directory and is defined relative to that directory.
```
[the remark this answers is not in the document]
CollierCZ: Hmm. I personally find your suggestion more complex. I think because "reference to the file" seems somewhat unclear to me, but I'm not sure. 

I could try moving part of it to the previous sentence and using `path` to make it clear what needs to be relative. See if that helps.

docs/src/add-services/varnish.md
CollierCZ: Probably, but this isn't a rework of the doc for better structure. Should be done in #2355

docs/src/add-services/varnish.md
```diff
@@ -81,8 +80,7 @@ varnish:
             path: config.vcl
 ```
 
-```bash
-# config.vcl
+```bash {location=".platform/config.vcl"}
```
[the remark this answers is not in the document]
CollierCZ: I added it where it was easy and I was already looking at the text. I didn't go through everything, as I was planning that for #2355. I can add it, as it's better, though out of scope.

docs/src/add-services/varnish.md
CollierCZ: I didn't think so. It's a nice add on, but it's not really necessary to run Varnish itself. Plus it's complicated and involves adding another app. I wouldn't push it so strongly.

docs/src/add-services/varnish.md
CollierCZ: If you wanted to make the app available to the outside world, I guess so. Right now the flow is just using SSH, so it doesn't seem necessary. It's up to them how they'd like to do it, though. It's not an ideal workaround, so I'm not sure I'd spend too long polishing it rather than just clearing up the current bug.

docs/src/add-services/varnish.md
CollierCZ: I was focusing on removing the bug, rather than optimizing the flow of building the workaround. I was also hoping to get an answer on when the product would be fixed so the workaround wasn't necessary, so I didn't spend so long on the flow there. I wasn't sure this was a comprehensive or clear list of endpoints.

I can add an example, though.

docs/src/add-services/varnish.md
CollierCZ: Defining the app name is probably better. But we shouldn't be using the abbreviated flags in general (though I didn't find that in the style guide quickly).

CollierCZ (commented):
If the workaround works, I don't think it's necessary to investigate fully how to make it available in other ways in order to fix the bug :shrug:

docs/src/add-services/varnish.md
trolologuy: The following:
> When the app is deployed, the app can access the Varnish service over HTTP to get diagnostic information.
Got me confused.

So that part of the flow could be a bit clearer, but that's maybe something for the bigger rework then.

docs/src/add-services/varnish.md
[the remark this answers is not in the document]
trolologuy: Yes 😻 That's much clearer and straightforward! Thanks!

trolologuy (approved):
Thanks for the changes 😍 🚀

docs/src/add-services/varnish.md
CollierCZ: Will make a note to come back to it. It's not ideal.

PR #2366, opened 2022-05-19
📝 Note the availability of Python 3.10

CollierCZ:
## Why

Python 3.10 is available. See context.

## What's changed

Updated to show that.


CollierCZ:
I guess the language doesn't follow the template. It has 3.9 in the steps, and then 2.7 in the full example :see_no_evil: 

Created #2367

PR #2369, opened 2022-05-20
🐛 Remove incorrect step referencing other DB

CollierCZ:
## Why

The guide for using SQLite had a step for installing PostgreSQL

## What's changed

Removed the step.


trolologuy (approved):
Indeed, that looked wrong. Good catch 👀 😍 

PR #2381, opened 2022-05-24
🐛 Allow more recent Hugo versions

CollierCZ:
## Why

To allow upgrades to newer Hugo versions. They were failing while looking for the HTML links generated by the sidebar.

## What's changed

Made the link renderer not look for relative links when the link ends with `html`.


trolologuy (approved):

PR #2383, opened 2022-05-24
📝 Add notes for legacy regions where appropriate

CollierCZ:
## Why

To make sure the motivation for migrating from legacy regions is clear. See context

## What's changed

Added shortcode in extra places
Adapted shortcode to allow different header levels

### DIFF
--- a/docs/src/create-apps/hooks/hooks-comparison.md
+++ b/docs/src/create-apps/hooks/hooks-comparison.md
@@ -5,11 +5,11 @@ description: What hooks are available in the build and deploy process and how to
 
 The following table presents the main differences among the three available hooks:
 
-| Hook name     | Failures stop build | Variables available | Services available | Run on `worker` instances | Writable directories | Blocks requests | Runs on all redeploys\* |
-| ------------- | ------------------- |------------------- | ------------------ | ------------------------- | -------------------- | --------------- | --------------- |
-| `build`       | Yes                 | Build variables     | No                 | Yes                       | `$PLATFORM_APP_DIR`, `$PLATFORM_CACHE_DIR`, and `/tmp` | No  | No  |
-| `deploy`      | No                  | Runtime variables   | Yes                | No                        | [Mounts](../app-reference.md#mounts)                   | Yes | No  |
-| `post_deploy` | No                  | Runtime variables   | Yes                | No                        | [Mounts](../app-reference.md#mounts)                   | No  | Yes |
+| Hook name     | Failures stop build | Variables available | Services available | Timeout | Run on `worker` instances | Writable directories | Blocks requests | Runs on all redeploys\* |
+| ------------- | ------------------- |-------------------- | ------------------ | ------- | ------------------------- | -------------------- | --------------- | --------------- |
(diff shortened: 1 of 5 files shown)


trolologuy (dismissed):
As per the context, the (probably not exhaustive) list of features that are not available on legacy regions is the following: 
- [ ] Automatically time out the builds after 1h. -> Couldn't find a mention of it in the docs. Might be worth an issue?
- [x] Outbound firewall restrictions.
- [x] Network-storage service.
- [x] Deploy hooks displayed in the activity logs.
- [x] Crons activities and logs in the activity logs.
- [x] Live backups.
- [x] Support for SSH during the deploy hook.
- [x] Grid Metrics.

I checked all of these, the information is provided and clear.

Except the one "missing" point above, and one non-blocking question, that PR is ready to be 🚀 ✅ 

docs/src/increase-observability/logs.md
```diff
@@ -6,6 +6,8 @@ aliases:
   - /development/logs.html
 ---
 
+{{% legacy-regions featureIntro="Deploy hook and cron activity logs" featureShort="these logs" plural=true %}}
```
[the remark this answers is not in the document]
CollierCZ: On this page, I thought it would get lost at the bottom. My idea was it would seem like I could ready only the activity logs section or only the container logs section and skip the rest. I don't think it's worth repeating for each section, so once at the top seemed to make sense to me.

The other places have just one place, so it was easier to fit into the structure there.

docs/src/create-apps/hooks/hooks-comparison.md
```diff
@@ -51,6 +51,13 @@ If a `build` hook fails for any reason, the build is aborted and the deploy does
 Note that this only works for `build` hooks.
 If other hooks fail, the deploy still happens.
 
+### Timeout
+
+Build hooks automatically time out if they run for 1 hour.
+So if you accidentally add an unbroken loop, it gets cut off and you can continue with other activities.
+
+{{% legacy-regions featureIntro="Build hooks timing out" featureShort="this feature" level=4 %}}
```
trolologuy, line 59: :nail_care:

To be aligned with the "deploy hook legacy note", this item should probably be one level higher.
That would also add it to the nav menu.
https://pr-2383-ax73k6a-652soceglkw4u.eu-3.platformsh.site/create-apps/hooks/hooks-comparison.html#supported-regions-1 

```suggestion
{{% legacy-regions featureIntro="Build hooks timing out" featureShort="this feature" level=3 %}}
```

trolologuy (approved):
😍 Thanks for the extra clarification. ✅

docs/src/create-apps/hooks/hooks-comparison.md
CollierCZ: I don't think I agree. The current structure is:
* Build
  * Build timeout
    * Supported regions
 * Deploy 
   * Supported regions

There's another layer of structure in the build one. If it were level 3, it would apply to all of the build hook, but it's only about the timeout, so that's why it's structured under that.

I could see the idea of adding more structure to the deploy section, but I wouldn't remove a level under the build.

PR #2387, opened 2022-05-26
🐛 Fix bugs in search

CollierCZ:
## Why

Search config was being cached
This was leading to outdated API keys
because the script was updated but not the config.

Also, nothing was returned when no results, so it wasn't clear that the search was working

## What's changed

Now always looks for the newest config version.
It's a small file, so the lack of cache shouldn't be a problem.

Also added a note for when no results are found.

### DIFF
--- a/docs/config/_default/params.yaml
+++ b/docs/config/_default/params.yaml
@@ -1,6 +1,6 @@
 
 # Basics
-version: "1.0.1"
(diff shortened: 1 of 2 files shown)


trolologuy (approved):
Great addition / good call for the no-result page. ✅ 

PR #2390, opened 2022-05-27
✅ Update to new Vale version

CollierCZ:
### Why

To fix reporting issues and get back to linting properly. It's a new version that uses [reviewdog](https://github.com/reviewdog/reviewdog) for reporting.

### What's changed

Switched to new version.


CollierCZ:
Some people already use the new version and the suggestion is that it only fails before the files are pushed to the default branch (see [example thread](https://github.community/t/workflow-files-only-picked-up-from-master/16129/55)). So I think we should try to merge it and see. If it breaks all branches, we'll revert. If it doesn't work, we'll be no worse than we are now.

trolologuy (approved):
(answer to comment), Indeed let's try and see if it fails (hopefully not).
Quite an amount of linting issues raised that are still left to tackle.

The changes look good to me, no specific feedback there.

🤞🏼 ✅

PR #2395, opened 2022-05-30
🐛 Get Vale running on PRs

CollierCZ:
## Why

Was failing every check

## What's changed

Changed the reporter to one that should work in our case
The PRs will at least run and annotations should show up, though I'm not sure the exact flow is ideal


trolologuy (approved):
✅ Let's try it out!

PR #2397, opened 2022-05-31
💄 Decrease weight of menu items

CollierCZ:
## Why

The redesign is further off than it may have seemed and people noted the menu is too bold.

## What's changed

It's not worth investigating the fonts exactly, so just removed the weight from lower-level menu items.


trolologuy (approved):
✅  Looks much better now 😍 

PR #2406, opened 2022-06-07
📝 Clarify service versions

CollierCZ:
## Why

Was some confusion in patch versions and how to apply them. See context.

## What's changed

Clarified what versions are available.

### DIFF
--- a/docs/src/add-services/_index.md
+++ b/docs/src/add-services/_index.md
@@ -121,6 +121,13 @@ Add them to the `type` key of the [service configuration](#1-configure-the-servi
 <!-- To update the versions in this table, use docs/data/registry.json -->
 {{< readFile file="src/registry/images/tables/services_supported.md" markdownify="true">}}
 
+### Service versions
+
+These services generally follow [semantic versioning conventions](https://semver.org/).
+You choose which major/minor version to add to your project.
+Patch versions are applied periodically for bug fixes and the like.
+When you deploy your app, you always get the latest available patches.
+
 ## Service timezones
 
 All services have their system timezone set to UTC by default.


docs/src/add-services/_index.md
```diff
@@ -121,6 +121,13 @@ Add them to the `type` key of the [service configuration](#1-configure-the-servi
 <!-- To update the versions in this table, use docs/data/registry.json -->
 {{< readFile file="src/registry/images/tables/services_supported.md" markdownify="true">}}
 
+### Service versions
+
+These services generally follow [semantic versioning conventions](https://semver.org/).
+You choose which major/minor version to add to your project.
+Patch versions are applied periodically for bug fixes and the like.
```
trolologuy, line 128: :hammer_and_wrench:
Although "Periodically" is a better fit, maybe adding the differentiation between what the customer has power over (and what not), might be helpful?
Since patches are applied by us, and they shouldn't be a direct concern for the customers.

```suggestion
Patch versions are applied automatically over time for bug fixes and the like.
```

trolologuy (approved):
Some propositions, but other than that ✅ 

docs/src/add-services/_index.md
CollierCZ: I considered moving away from the original `periodically` wording, but the thing is that the updates aren't applied automatically. The process on our side is manual. I feel like using `automatically` gives the impression that as soon as a patch is released it'll automatically be added to our versions.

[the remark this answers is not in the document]
CollierCZ: If the idea is to be explicit about what users can do, the first part probably shouldn't be passive.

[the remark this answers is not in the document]
CollierCZ: But that's why the previous sentence used "You" -- that's the part they can handle themselves.

trolologuy: Fair point too... hmm, should we add more clarification, or good enough for now?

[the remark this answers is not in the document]
trolologuy: Since the PR adds clarification about the process, the "limitations" and helps clarify what the customer can (and can't change). I'd say we could dismiss the comments for now (unless we come up with a simpler explanation), since it might be a bit over the original scope of that PR?

CollierCZ: I personally think it's clear enough for now. But if you feel differently or have another suggestion, we can look into it more.

PR #2407, opened 2022-06-07
🚚 Rename create apps

CollierCZ:
## Why

It was suggested that "create" apps is a little confusing, since the information isn't really about the development process.

## What's changed

Changed to "Configure" to include existing apps and the like.


trolologuy (approved):

PR #2409, opened 2022-06-07
📝 Update Slack activity script

chadwcarlson:
## Why

Some small gotchas I've found playing with activity scripts


CollierCZ (approved):
Thanks for the added context. I've reformatted it slightly to better fit the style guide for the docs.

PR #2411, opened 2022-06-08
➖ Remove unnecessary deprecated dependency

CollierCZ:
## Why

A package that was deprecated 3 years ago had conflicts with more recent packages. This was breaking builds

## What's changed

Removed the package as it wasn't necessary. Only thing that could potentially be affected is the presentation of the search results, but they seem unaffected (as they should be)

### DIFF
--- a/docs/config/_default/params.yaml
+++ b/docs/config/_default/params.yaml
@@ -1,6 +1,6 @@
 
 # Basics
-version: "1.0.3"
+version: "1.0.4"
 description: Platform.sh User Documentation
 author: Platform.sh
 title: User documentation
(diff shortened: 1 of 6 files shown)


trolologuy (changes requested):
There's a breaking change for the extended list of search results.
HTML code is displayed on the page:

```
Other resources from Platform.sh
<b>Templates | </b><code>php</code>
```
To reproduce:
1. [Start a search](https://pr-2411-ti3tv2q-652soceglkw4u.eu-3.platformsh.site/search.html?q=build%20log)
2. Click on `Show all results`
3. Scroll to the bottom of the page

trolologuy (approved):
Working as expected!
✅

CollierCZ:
The cron task now runs successfully

PR #2413, opened 2022-06-09
✅ Increase buffer size for link checker

CollierCZ:
## Why

To get around one larger document. Relates to #2213

## What's changed

Increased size


trolologuy (approved):
I added the link to the metrics in the additional context. 
The resource usage looks good. I wanted to cross-check between the current plan size L resources and the calculator but I lost access to it.

Is there anything more for me to check or to pay attention to?

CollierCZ:
I wouldn't even have thought that would be a serious concern, so nothing I can think of.

PR #2422, opened 2022-06-15
Fixes grammar and brevity

SKhubaib:
## Why

Line 93: helping: incorrect form of verb - change to 'help'.
Line 105: and instead try: missing punctuation - add comma after instead.
Line 110: offers an explanation of: against your guidelines for brevity - change to 'explains'.
Line 133: chance readers leave the page and not return: missing particle and helping verb - change to 'chance that readers leave the page and don't return'.
Line 187: want be: missing particle - change to 'want to be'.
Line 188: make it clearer: unnecessary pronoun - change to 'make clearer'.
Line 242: making readings think: incorrect noun - change to 'making readers think'.
Line 243: Make use to use `it's` only for: against guidelines for brevity and incorrect preposition: change to 'Use `it's` only for'.
Line 257: these aren't always understood correctly: negation is incorrect - change to 'these are always understood correctly'.
Line 284: too many in one place: missing noun - change to 'too many note in one place'.
Line 319: checks note when: change to 'check notes when'.

## What's changed

Grammar, mistakes, and brevity.

### DIFF
--- a/contributing/content-style.md
+++ b/contributing/content-style.md
@@ -102,7 +102,7 @@ Some guides to help you make sure your writing is inclusive and accessible to ev
 
 To make your content as accessible as possible, the purpose of a link should be clear from its text alone.
 Avoid links with text like `click here`
-and instead, try to include meaning inside the link itself.
+and instead try to include meaning inside the link itself.
 
 Remember to use the [proper format for links](./markup-format.md#links)
 
@@ -185,7 +185,7 @@ see Content Design London's [Readability Guidelines on clear language](https://r
 ### Use direct sentences
 
 You generally want to be as direct as possible to keep the ideas simple.
-By avoiding overly wordy phrases, you help make clearer what needs to be done.
+By avoiding overly wordy phrases, you help make it clearer what needs to be done.
 <!-- vale Platform.condescending = YES -->
 
 For example, try to avoid using sentences starting with `There is/are` or `It is` too often.
@@ -254,7 +254,7 @@ Too many abbreviations can be confusing to readers.
 Make sure to explain any abbreviations that might be unfamiliar the first time they're used.
 
 Avoid abbreviations for Latin terms, such as `i.e.`, `e.g.`, and `etc.`.
-as these are always understood correctly.
+as these aren't always understood correctly.
 
 Use                                                                         | Avoid 
 ----------------------------------------------------------------------------|-------
@@ -281,7 +281,7 @@ To login, run `platform login`.                               | To login, run `p
 
 Notes are pieces of information that stand outside the normal text flow.
 Use them for short ideas that aren't.
-Don't use too many notes in one place or they lose their value.
+Don't use too many in one place or they lose their value.
 
 There are three types of notes:
 
@@ -316,7 +316,7 @@ Some of the rules are enforced with [Vale](https://docs.errata.ai/vale/about), a
 
 Because writing style is subjective and no checking tool is perfect,
 the rules are mostly set to warnings rather than errors.
-So check notes when something might be wrong,
+So automated checks show when something might be wrong,
 but use your common sense and ignore them when appropriate.
 
 One exception is spelling, which returns errors.
--- a/contributing/content-style.md
+++ b/contributing/content-style.md
@@ -90,7 +90,7 @@ allowlist                                                | whitelist
 
 ### Resources for inclusive language
 
-Some guides to helping you make sure your writing is inclusive and accessible to everyone.
+Some guides to help you make sure your writing is inclusive and accessible to everyone.
 
 * [Mailchimp guide to writing for accessibility](https://styleguide.mailchimp.com/writing-for-accessibility/)
 * [Microsoft's ideas on writing for all abilities](https://docs.microsoft.com/en-us/style-guide/accessibility/writing-all-abilities)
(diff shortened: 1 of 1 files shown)


contributing/content-style.md
```diff
@@ -254,7 +254,7 @@ Too many abbreviations can be confusing to readers.
 Make sure to explain any abbreviations that might be unfamiliar the first time they're used.
 
 Avoid abbreviations for Latin terms, such as `i.e.`, `e.g.`, and `etc.`.
-as these aren't always understood correctly.
+as these are always understood correctly.
```
ludoge, line 257: Removes a negation that would be correct

contributing/content-style.md
```diff
@@ -90,7 +90,7 @@ allowlist                                                | whitelist
 
 ### Resources for inclusive language
 
-Some guides to helping you make sure your writing is inclusive and accessible to everyone.
+Some guides to help you make sure your writing is inclusive and accessible to everyone.
```
CollierCZ, line 93: I think the original idea was more like `for helping`, but this works as well.

contributing/content-style.md
```diff
@@ -102,12 +102,12 @@ Some guides to helping you make sure your writing is inclusive and accessible to
 
 To make your content as accessible as possible, the purpose of a link should be clear from its text alone.
 Avoid links with text like `click here`
-and instead try to include meaning inside the link itself.
+and instead, try to include meaning inside the link itself.
```
CollierCZ, line 105: :clipboard: The comma there does not fit with our style. We do not want to emphasize it as an aside.

```suggestion
and instead try to include meaning inside the link itself.
```

contributing/content-style.md
```diff
@@ -184,8 +184,8 @@ see Content Design London's [Readability Guidelines on clear language](https://r
 
 ### Use direct sentences
 
-You generally want be as direct as possible to keep the ideas simple.
-By avoiding overly wordy phrases, you help make it clearer what needs to be done.
+You generally want to be as direct as possible to keep the ideas simple.
+By avoiding overly wordy phrases, you help make clearer what needs to be done.
```
CollierCZ, line 188: :hammer_and_pick: I think it's clearer with an object.
```suggestion
By avoiding overly wordy phrases, you help make it clearer what needs to be done.
```

contributing/content-style.md
CollierCZ: Yes, it was `aren't` on purpose because those terms are not always understood (they are often misunderstood).
```suggestion
as these aren't always understood correctly.
```

contributing/content-style.md
```diff
@@ -281,7 +281,7 @@ To login, run `platform login`.                               | To login, run `p
 
 Notes are pieces of information that stand outside the normal text flow.
 Use them for short ideas that aren't.
-Don't use too many in one place or they lose their value.
+Don't use too many notes in one place or they lose their value.
```
CollierCZ, line 284: :hammer_and_wrench: I like your focus above on taking out unnecessary words. I feel like `notes` here isn't necessary, so I'd keep that trend and remove it here.
 ```suggestion
Don't use too many in one place or they lose their value.
```

File: contributing/content-style.md (hunk 1 above)
Comment (SKhubaib) [reply contributing/content-style.md:257]: Sorry? So this means that the terms like i.e., e.g., and etc. should be avoided?

Or it means that id est, exampli gratia, and etcetera should be used instead of the abbreviated forms?

If this means none of the above two, then do you not think that the negation is wrong?

Please explain. Thanks

File: contributing/content-style.md [hunk 6]
```diff
@@ -316,7 +316,7 @@ Some of the rules are enforced with [Vale](https://docs.errata.ai/vale/about), a
 
 Because writing style is subjective and no checking tool is perfect,
 the rules are mostly set to warnings rather than errors.
-So checks note when something might be wrong,
+So check notes when something might be wrong,
```
CollierCZ, line 319: :lady_beetle: It was actually correct originally (before this change). But I can see how it would be confusing. The issue is that it uses the GitHub term `check` without really introducing it.

Perhaps something like this would help?
```suggestion
So automated checks show when something might be wrong,
```

CollierCZ (changes requested):
Rather embarrassing that the style guide of all places would have so many typos and the like :see_no_evil: Thanks for the improvements!

Some of the suggestions don't work. See if my comments about how to fix those issues make sense or if you can come up with better ideas.

contributing/content-style.md
CollierCZ: All Latin terms should be avoided (both the abbreviations and the full forms).

The terms are *not* always understood, which is why we made the suggestion.

SKhubaib:
Here too, under Use Notes Appropriately, the second sentence ends abruptly. It says, Use them for short ideas that aren't.

Short ideas that aren't WHAT? Please fix.

SKhubaib:
Also, the line 256 is still ambiguous for a common reader. Should it not be something like

Avoid abbreviated Latin terms, such as `i.e.`, `e.g.`, and `etc.`, and their full forms, as these aren't always understood correctly.

A common reader expects the full forms after such as, where abbreviated terms are stated. Also, after etc., there's a period, while the sentence goes on. Thanks

CollierCZ (approved):
Thanks for the changes!

Issue #2378, opened 2022-05-23
✨ Add reporting for collected feedback


CollierCZ:
Once #2377 is completed, we will need a way to view and process the feedback. Create reports to make this easier.

It should be easy to sort by page, positive votes, and negative votes. And a place for all feedback to process (ideally with ability to mark as resolved/won't do)

Ideally could correlate with page view data to aid in prioritization.

PR #2425, opened 2022-06-15
✨ Add a form to collect feedback

### LINKED ISSUE #2377: ✨Add a form to collect feedback
We'd like to collect feedback from docs users on which pages work and which don't. It could be as simple as "Was this page helpful? Yes/No" with an optional feedback form to collect more.

CollierCZ:
## Why

Closes #2377

## What's changed

Added a basic form that uses Alpine.js to send feedback to a MariaDB table


CollierCZ:
@trolologuy looking for a basic sanity check at this point. Whether the direction makes sense.

The very minimalist report is here: https://pr-2425-idyy6zy-652soceglkw4u.eu-3.platformsh.site/feedback/report.html

docs/themes/avocadocs/layouts/partials/feedback-form.html
```diff
@@ -0,0 +1,75 @@
...
+      class="feedback-message"
+    >
+      ❤️ Thanks for letting us know!
+    </div>
+    <div
+      x-show="feedbackSubmitted && feedback === 'negative'"
+      class="feedback-message"
+    >
+      😞 Thanks for the feedback. <br />
+      <a :href="editPageUrl">Suggest a change</a>
```
trolologuy, line 21: :heart_eyes:
I love the flow you did there!!! 😍 
Really a great idea and implementation! The links are useful and should help reduce barriers to entry.

trolologuy (commented):
Great poc! It works perfectly!

Some thoughts:
- We don't block/hinder spamming, is a re-captcha an option? Or limit the number of vote per page to one IP per week? (no idea of the complexity, just a thought).
- Some grouping by dates? Or time ranges? Otherwise, the table is going to be very long... (e.g. from the 1-6th June, the page X had 10 upvotes and 30 downvotes or something)
- The vote option is only available on pages where the TOC is present. Does it make sense to display it at the bottom of the page as well? (and automatically hide it from the sidebar when the scroll is getting to the bottom)?
- For the styling, would it make sense to add the same fine line around the block as for the TOC?

CollierCZ:
1. It's not actually a form, so I don't think automated spammers would use it. Not sure that it'd be a problem. If it starts to be, we could look into it.
2. I figured anything advanced in reporting would go into #2378
3. I think it'd be better to standardize on having the space there all the time. Maybe just at the bottom on thin viewports.
4. If anything, I'd take out the one around the contents. Not add more of them. I've asked if design as time.

chadwcarlson:
This is great!

trolologuy (approved):
As discussed, safe to merge as functionality is working as expected.

PR #2432, opened 2022-06-20
🐛 Get better at search for specific properties

### LINKED ISSUE #2423: 🐛 Allow searching for code strings with separators

The search

Searching for strings with separators like `X-Frame-Options` and `memory_ratio` don't show the results for pages that have those strings directly. Putting quotes around the strings doesn't help.

We'd like people to be able to get info on specific properties and strings, so the search should return these results.

### Additional context

_No response_

CollierCZ:
## Why

Closes #2423

## What's changed

Moved around ranking rules so exactness is higher. Slight problems with some other searches, but far better results for things like `X-Frame-Options`. Seems overall better.

### DIFF
--- a/docs/static/scripts/xss/src/components/SuggestionsPrimary.js
+++ b/docs/static/scripts/xss/src/components/SuggestionsPrimary.js
@@ -11,12 +11,7 @@ const SuggestionsPrimary = ({ hits, title }) => {
     <li key={r.relurl}>
       <h5>
         <a href={r.url}>
-          <b>
-            {r.section}
-            {' '}
-            |
-            {' '}
-          </b>
+          <b dangerouslySetInnerHTML={{ __html: `${DOMPurify.sanitize(r.section)} | ` }} />
           {r.title}
         </a>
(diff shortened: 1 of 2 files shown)


trolologuy (dismissed):
The returned results look indeed better ✅ 

There's an [encoding bug](https://pr-2432-6efhm3a-652soceglkw4u.eu-3.platformsh.site/search.html?q=X-Frame-Options) on the results returned when searching for `X-Frame-Options`.
It might be linked to how [parent elements are parsed](https://pr-2432-6efhm3a-652soceglkw4u.eu-3.platformsh.site/search.html?q=Set%20custom%20headers%20on%20static%20content) but the bug is also on the [live version of the docs](https://docs.platform.sh/search.html?q=Set%20custom%20headers%20on%20static%20content).

CollierCZ:
Yeah, that's been there for a while. Never got around to it. Should be fixed now though.

trolologuy (approved):
Works like a charm 😻  ✅

Issue #2418, opened 2022-06-14
📝 Rework region migration


CollierCZ:
### Where on docs.platform.sh should be changed?

https://docs.platform.sh/projects/region-migration.html

### What exactly should be updated?

Update for:

- Content accuracy
- Includes what it should
- Doesn't include what doesn't fit
- Fits template
- Fits style guide

Ensure reasons for migration are up to date. Including what a legacy region is.

platformbill:
Perhaps https://docs.platform.sh/development/regions.html should also have a "what is a legacy region" type section.

CollierCZ:
Thanks for the note, @platformbill! Added it as part of #2434

PR #2434, opened 2022-06-21
📝 Clarify region availability

### LINKED ISSUE #2426: 📝 Add availability details for the regions

https://docs.platform.sh/development/regions.html

- The region list should only include public regions.
- Regions marked as legacy should have their names changed
- Add a note about how region availability depends on org settings
- Regions that are not publicly available without specific settings should be marked as such in the list (or mark the others as "available to new organizations" or something similar)
- Consider changing the order of the list and other changes to make it clearer what's available and what is a priority.

CollierCZ:
## Why

Closes #2426

## What's changed

- Added legacy labels
- Removed non-public region
- Clarified org settings
- Added data on whether available to new orgs and added column in table
- Sorted tables based on whether available or not

### DIFF
--- a/docs/data/regions.yaml
+++ b/docs/data/regions.yaml
@@ -1,8 +1,9 @@
 europe:
     eu:
-        name: Ireland (eu)
+        name: Ireland (eu-legacy)
         label: eu.platform.sh
         provider: AWS
+        available: false
         outbound_ips: 
             - "54.72.94.105"
             - "54.76.137.67"
@@ -16,6 +17,7 @@ europe:
         name: Ireland  (eu-2)
         label: eu-2.platform.sh
         provider: AWS
+        available: false
         outbound_ips: 
             - "52.208.123.9"
             - "52.214.63.84"
@@ -29,6 +31,7 @@ europe:
         name: Ireland (eu-4)
         label: eu-4.platform.sh
         provider: AWS
+        available: false
         outbound_ips: 
             - "18.200.158.188"
             - "18.200.157.200"
@@ -42,6 +45,7 @@ europe:
         name: Sweden (eu-5)
         label: eu-5.platform.sh
         provider: AWS
+        available: true
         outbound_ips: 
             - "13.48.116.14"
             - "13.51.46.87"
@@ -53,6 +57,7 @@ europe:
         name: Germany (de-2)
         label: de-2.platform.sh
         provider: Google Cloud Platform
+        available: true
         outbound_ips: 
             - "35.246.248.138"
             - "35.246.184.45"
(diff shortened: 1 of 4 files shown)


docs/src/projects/region-migration.md
```diff
@@ -12,16 +12,22 @@ aliases:
 Platform.sh doesn't offer an automated way to migrate a project from one region to another after it's created.
 However, the process to do so manually is fairly straightforward and can be scripted.
 
-## Why migrate between regions?
+## Why migrate between regions
 
 * Different data centers are located in different geographic areas,
   and you may want to keep your site physically close to the bulk of your user base for reduced latency.
```
trolologuy, line 18: :bulb:
Do we want to also mention the environmental aspect of such a migration to newer regions? 

https://docs.platform.sh/development/regions.html#environmental-impact

docs/data/regions.yaml
```diff
@@ -62,23 +67,11 @@ europe:
...
-        inbound_location: gw.fr-1.platformsh.site
-        inbound_ips:
-            - "90.84.47.148"
-            - "90.84.46.222"
-            - "90.84.46.40"
     fr-3:
         name: France (fr-3)
         label: fr-3.platform.sh
         provider: OVHcloud
+        available: true
```
trolologuy, line 74: :bulb:
Shouldn't the carbon intensity be also directly included in the table to help customer choose which region they want to create from / migrate to? 

The carbon intensity of a region is shown during the project creation step, but there's no overview available (only solution is the CLI) which feels cumbersome.

docs/layouts/shortcodes/regions.html
```diff
@@ -8,11 +8,12 @@
         <tr class="region-table-header">
             <th>Region</th>
             <th>Provider</th>
+            <th>Available to new orgs</th>
```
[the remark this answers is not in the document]
trolologuy: I agree with Guguss there 👍🏼

docs/layouts/shortcodes/regions.html
```diff
@@ -24,6 +25,7 @@ <h4><a href="#{{ $regionAnchor }}" id="{{ $regionAnchor }}" aria-label="Anchor">
             </td>
             <!-- Provider -->
             <td>{{ .provider }}</td>
+            <td>{{ if .available }}✅{{ end }}</td>
```
trolologuy, line 28: :bulb:
Do we want to add a ⚠️ for the ones that are not directly available?
Since in some cases, it's still possible for customers, but it can be seen as a reminder that not all features are available + a further incentive to consider migration.

trolologuy (approved):
The changes solve all the points raised in the linked issue and improve clarity of the page a lot 😍 
I added a few (non-blocking) ideas for your consideration for future iterations.

As raised by Guguss, the headline of the table might be changed to improve clarity.

- [x] The information is correct (matches the dashboard).

docs/src/projects/region-migration.md
CollierCZ: This is outside the scope of this task. Part of #2418

docs/data/regions.yaml
CollierCZ: I guess that would be possible, though the number could be dynamic. Seems slightly out of scope here.

docs/layouts/shortcodes/regions.html
CollierCZ: Also, regions that aren't available might still have all the features. There's a difference between `legacy` and not `available`

docs/src/projects/region-migration.md
trolologuy: Okey, should I add a note in there? Or is it part of the rework? (doesn't seem to be explicitly a part of it)

docs/data/regions.yaml
trolologuy: It's indeed out of scope but might be nice to have.

docs/src/projects/region-migration.md
CollierCZ: It's part of the reasons for migration. It's already done.

PR #2443, opened 2022-06-23
:memo: Update language template

trolologuy:
## Why

The config reader is now set up as a shortcode. This should be reflected in the template.

### DIFF
--- a/docs/templates/language-reference.md
+++ b/docs/templates/language-reference.md
@@ -54,29 +54,30 @@ To configure a given service, see the page dedicated to that service.
 {{< codetabs >}}
 
 ---
-title=<SERVICE_EXAMPLE>
+title=<SERVICE>
 file=static/files/fetch/examples/<LANGUAGE_TYPE>/<SERVICE>
-highlight=<LANGUAGE_TYPE>
+highlight=<LANGUAGE_NAME>
 markdownify: false
 ---
 
 <--->
 
 ---
-title=<DIFFERENT_SERVICE_EXAMPLE>
+title=<DIFFERENT_SERVICE>
 file=static/files/fetch/examples/<LANGUAGE_TYPE>/<DIFFERENT_SERVICE>
-highlight=<LANGUAGE_TYPE>
+highlight=<LANGUAGE_NAME>
 markdownify: false
 ---
 
 {{< /codetabs >}}
 
-<!-- If available, include the configuration reader available for the given language,
- by adding the shortcode below and providing the link to the specific flavour of the config reader
-(for example, for the Python library, the config reader link is: [platformshconfig](https://github.com/platformsh/config-reader-python)). -->
-
 {{% config-reader %}}
 [`<CONFIG_READER_NAME>`](<CONFIG_READER_REPOSITORY_URL>)
+{{% /config-reader%}}
+
+For `Node.js`, that would result in the following:
+{{% config-reader %}}
+[platformshconfig](https://github.com/platformsh/config-reader-nodejs).
 {{% /config-reader%}}
 
 ## Any other functions specific to the language
--- a/docs/templates/language-reference.md
+++ b/docs/templates/language-reference.md
@@ -45,38 +45,44 @@ Depending on the language, a brief description of the recommended way to handle 
 
 ## Connect to services
 
-If available, include examples for connecting in an app
+If available, include examples for connecting to different services in an app
 (taken from https://examples.docs.platform.sh/).
+
+The following examples show how to use <LANGUAGE_NAME> to access various [services](/add-services/_index.md).
+To configure a given service, see the page dedicated to that service.
 
 {{< codetabs >}}
 
 ---
-title=<EXAMPLE_1>
+title=<SERVICE_EXAMPLE>
 file=static/files/fetch/examples/<LANGUAGE_TYPE>/<SERVICE>
-highlight=js
+highlight=<LANGUAGE_TYPE>
+markdownify: false
 ---
 
 <--->
 
 ---
-title=<EXAMPLE_2>
-file=static/files/fetch/examples/<LANGUAGE_TYPE>/<SERVICE>
-highlight=js
+title=<DIFFERENT_SERVICE_EXAMPLE>
+file=static/files/fetch/examples/<LANGUAGE_TYPE>/<DIFFERENT_SERVICE>
+highlight=<LANGUAGE_TYPE>
+markdownify: false
 ---
 
 {{< /codetabs >}}
 
-<!-- If available, include the configuration reader available for the given language
-(for example, for [Python](https://github.com/platformsh/config-reader-python)). -->
+<!-- If available, include the configuration reader available for the given language,
+ by adding the shortcode below and providing the link to the specific flavour of the config reader
+(for example, for the Python library, the config reader link is: [platformshconfig](https://github.com/platformsh/config-reader-python)). -->
 
 {{% config-reader %}}
-[`<NAME_OF_THE_CONFIG_READER_FOR_THAT_LANGUAGE>`](<LINK_TO_THE_REPO>)
+[`<CONFIG_READER_NAME>`](<CONFIG_READER_REPOSITORY_URL>)
 {{% /config-reader%}}
 
 ## Any other functions specific to the language
 
 Structure like [how to steps](./how-to.md#1-do-this-step-first).
-Split each function 
+Split each function.
 
 ## Project templates
 
--- a/docs/templates/language-reference.md
+++ b/docs/templates/language-reference.md
@@ -63,8 +63,8 @@ markdownify: false
 <--->
 
 ---
-title=<DIFFERENT_SERVICE>
-file=static/files/fetch/examples/<LANGUAGE_TYPE>/<DIFFERENT_SERVICE>
+title=<DIFFERENT_SERVICE_NAME>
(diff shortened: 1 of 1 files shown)


docs/templates/language-reference.md
```diff
@@ -49,10 +49,30 @@ Depending on the language, a brief description of the recommended way to handle
 If available, include examples for connecting in an app
 (taken from https://examples.docs.platform.sh/).
 
-### Configuration reader
+{{< codetabs >}}
 
-If available, include the configuration reader available for the given language
-(for example, for [Python](https://github.com/platformsh/config-reader-python)).
+---
+title=<EXAMPLE_1>
```
CollierCZ, line 55: :lady_beetle: This isn't really descriptive as a label. The number isn't as helpful as it could be.

Also, it obscures what should be placed there, which is not just an example.

docs/templates/language-reference.md
```diff
@@ -49,10 +49,30 @@ Depending on the language, a brief description of the recommended way to handle
...
+file=static/files/fetch/examples/<LANGUAGE_TYPE>/<SERVICE>
```
CollierCZ, line 56: :lady_beetle: I'd be consistent here. If "TYPE" is to indicate it should be the machine name, then it should be for both.

docs/templates/language-reference.md
```diff
@@ -49,10 +49,30 @@ Depending on the language, a brief description of the recommended way to handle
...
+highlight=js
```
CollierCZ, line 57: :lady_beetle: This isn't true for all languages.

docs/templates/language-reference.md
```diff
@@ -49,10 +49,30 @@ Depending on the language, a brief description of the recommended way to handle
...
+---
+
+<--->
+
+---
+title=<EXAMPLE_2>
+file=static/files/fetch/examples/<LANGUAGE_TYPE>/<SERVICE>
+highlight=js
```
CollierCZ, line 65: As above

docs/templates/language-reference.md
```diff
@@ -49,10 +49,30 @@ Depending on the language, a brief description of the recommended way to handle
...
-### Configuration reader
+{{< codetabs >}}
 
-If available, include the configuration reader available for the given language
-(for example, for [Python](https://github.com/platformsh/config-reader-python)).
+---
+title=<EXAMPLE_1>
+file=static/files/fetch/examples/<LANGUAGE_TYPE>/<SERVICE>
+highlight=js
+---
```
CollierCZ, line 58: :question: Is this missing `markdownify: false`?

docs/templates/language-reference.md
```diff
@@ -49,10 +49,30 @@ Depending on the language, a brief description of the recommended way to handle
...
+title=<EXAMPLE_2>
+file=static/files/fetch/examples/<LANGUAGE_TYPE>/<SERVICE>
+highlight=js
+---
+
+{{< /codetabs >}}
+
+<!-- If available, include the configuration reader available for the given language
+(for example, for [Python](https://github.com/platformsh/config-reader-python)). -->
```
CollierCZ, line 71: :lady_beetle: This example worked with the previous wording, but doesn't fit the pattern of how with do examples with variables.

docs/templates/language-reference.md
```diff
@@ -49,10 +49,30 @@ Depending on the language, a brief description of the recommended way to handle
...
+
+{{% config-reader %}}
+[`<NAME_OF_THE_CONFIG_READER_FOR_THAT_LANGUAGE>`](<LINK_TO_THE_REPO>)
```
CollierCZ, line 74: :pick: The first name is too long (more like part of a sentence than a variable name) and the second not specific enough.

CollierCZ (changes requested):
Good to get this in there. Some small places for improvement.

docs/templates/language-reference.md
```diff
@@ -44,20 +44,45 @@ Ideally, also with a full working example that shows it in context.
 
 Depending on the language, a brief description of the recommended way to handle dependencies in the language.
 
-## Connect to services
+## Connecting to services
```
CollierCZ, line 47: :lady_beetle: This change does not fit the style guide: https://developers.google.com/style/headings?hl=en#heading-and-title-text

docs/templates/language-reference.md
```diff
@@ -44,20 +44,45 @@ Ideally, also with a full working example that shows it in context.
...
 
-If available, include examples for connecting in an app
+If available, include examples for connecting to different services in an app
 (taken from https://examples.docs.platform.sh/).
 
-### Configuration reader
+The following examples show how to use <LANGUAGE_NAME> to access various services. To configure a given service, see the page dedicated to that service.
```
CollierCZ, line 52: :hammer_and_wrench: 
```suggestion
The following examples show how to use <LANGUAGE_NAME> to access various [services](/add-services/_index.md).
To configure a given service, see the page dedicated to that service.
```

:lady_beetle: I would note that this addition means the text in the template is more of a mixture of instructions and text that could be included in the final version and it's not immediately clear which is which.

docs/templates/language-reference.md
CollierCZ: This wasn't resolved. Note that `<SERVICE>` in line 57 is different than `<SERVICE>` in line 58

docs/templates/language-reference.md
```diff
@@ -44,20 +44,45 @@ Ideally, also with a full working example that shows it in context.
...
+file=static/files/fetch/examples/<LANGUAGE_TYPE>/<SERVICE>
+highlight=<LANGUAGE_TYPE>
+markdownify: false
+---
+
+{{< /codetabs >}}
+
+<!-- If available, include the configuration reader available for the given language,
+ by adding the shortcode below and providing the link to the specific flavour of the config reader
+(for example, for the Python library, the config reader link is: [platformshconfig](https://github.com/platformsh/config-reader-python)). -->
```
CollierCZ, line 76: :lady_beetle: This is not what I meant. Our general pattern with variables is to show the command with the variables and then show an example with actual values _after_ that command. Adding all of the additional text here doesn't seem to add as much clarity.

docs/templates/language-reference.md
```diff
@@ -44,20 +44,45 @@ Ideally, also with a full working example that shows it in context.
...
+
+{{% config-reader %}}
+[`<CONFIG_READER_REPO_NAME>`](<LINK_TO_THE_SPECIFIC_REPO>)
```
CollierCZ, line 79: :hammer_and_wrench: 
```suggestion
[`<CONFIG_READER_NAME>`](<CONFIG_READER_REPOSITORY_URL>)
```

CollierCZ (changes requested):

docs/templates/language-reference.md
trolologuy: Hmm maybe it would be clearer when removing the type altogether?
Or would it be lacking the differences with capitalization + the slightly different name for the highlight?

docs/templates/language-reference.md
trolologuy: So I guess it's safe to remove that part altogether? Since the example should be clear enough.

docs/templates/language-reference.md
CollierCZ: I would personally be inclined for the suggestion below to distinguish between the human-readable name and the machine-readable name.

docs/templates/language-reference.md
```diff
@@ -45,18 +45,45 @@ Depending on the language, a brief description of the recommended way to handle
...
+The following examples show how to use <LANGUAGE_NAME> to access various [services](/add-services/_index.md).
+To configure a given service, see the page dedicated to that service.
 
-If available, include the configuration reader available for the given language
-(for example, for [Python](https://github.com/platformsh/config-reader-python)).
+{{< codetabs >}}
+
+---
+title=<SERVICE>
+file=static/files/fetch/examples/<LANGUAGE_TYPE>/<SERVICE>
```
CollierCZ, line 58: :hammer_and_wrench: 
```suggestion
title=<SERVICE_NAME>
file=static/files/fetch/examples/<LANGUAGE_TYPE>/<SERVICE_TYPE>
```

docs/templates/language-reference.md
```diff
@@ -45,18 +45,45 @@ Depending on the language, a brief description of the recommended way to handle
...
+markdownify: false
+---
+
+{{< /codetabs >}}
+
+{{% config-reader %}}
+[`<CONFIG_READER_NAME>`](<CONFIG_READER_REPOSITORY_URL>)
+{{% /config-reader%}}
+
+For `Node.js`, that would result in the following:
```
CollierCZ, line 78: :pick: 
```suggestion
For Node.js, that would result in the following:
```

Node.js isn't the codename.

CollierCZ (changes requested):

docs/templates/language-reference.md
```diff
@@ -45,18 +45,45 @@ Depending on the language, a brief description of the recommended way to handle
...
+{{< /codetabs >}}
+
+{{% config-reader %}}
+[`<CONFIG_READER_NAME>`](<CONFIG_READER_REPOSITORY_URL>)
+{{% /config-reader%}}
+
+For Node.js, that would result in the following:
+{{% config-reader %}}
+[platformshconfig](https://github.com/platformsh/config-reader-nodejs).
+{{% /config-reader%}}
```
CollierCZ, line 81: :nail_care: Looking at the unresolved comments, I maybe liked this better when you had it as a comment.

CollierCZ (approved):

PR #2446, opened 2022-06-28
📝 Clarify versions for Dedication Gen 3

CollierCZ:
## Why

It wasn't always clear what was available for Dedicated Gen 3. Some lists were out of date and it wasn't included in languages at all.

## What's changed

Went through all services and languages to see if they were supported. Updated registry when services were supported. Changed table heading for all languages because the Grid and Gen 3 should be the same.


trolologuy (dismissed):
✅  A few open questions, but nothing blocking.

❓  What is the purpose of the empty `docs/layouts/shortcodes/image-versions-language.md` file?

💡 Do we want the same table aggregated layout for `languages` and `services` where it makes sense? Or would it be more confusing?
For example here: https://pr-2446-bmsktki-652soceglkw4u.eu-3.platformsh.site/add-services/rabbitmq.html

❓ 💡  (I think we talked about that, but I'm not sure any more, so ignore if duplicate) 
For the future, would it make sense to have an initiative to improve the tracking of versions supported? Or some dashboard? (to avoid mistakes and make sure doc update is part of the process).

CollierCZ:
* The file shouldn't have been there.
* I think consistency is better in terms of services have a consistent layout and languages do too. Services can vary (it's explicitly [part of Gen 3 descriptions](https://docs.platform.sh/dedicated-gen-3/overview.html#available-services)), so it makes sense to have separate entries there. I think it's better to be clearer there.
* It's going to be an initiative, but not by us. By the Images team.

trolologuy (approved):
✅

PR #2448, opened 2022-06-29
✨ Build environments through a workflow

CollierCZ:
## Why

To not build environments for all PRs. See context

## What's changed

Created a workflow to activate branch environments and updated the info workflow to adjust to that


0xDSousa:
@CollierCZ, so switching this to a workflow will require explicit approval from a team member to start a build?

Flow will now be that PRs are created but matching environment will be decided later, correct?

0xDSousa (approved):
I'm satisfied that it improves upon our current security for these repos.

PR #2460, opened 2022-07-08
📝 Add note on incomming new jitter value

remilejeune2:
# Why

Add note to warn about the new jitter value roll out.

# What's changed

The jitter value will changed from 5 minutes to 20 minutes in incomming updates.


CollierCZ:
Handled in #2465

PR #2462, opened 2022-07-11
Fix unneeded quotes for web.locations root

lolautruche:
## Why

Contents of `.platform.app.yaml` is presented wrongly regarding `web.locations` root:

```yaml
web:
  locations: 
    '/': 
      root: "\"web\""
      passthru: "/index.php"
```

Instead of:

```yaml
web:
  locations: 
    '/': 
      root: "web"
      passthru: "/index.php"
```

This PR fixes it in the docs.

### DIFF
--- a/docs/data/registry.json
+++ b/docs/data/registry.json
@@ -951,7 +951,7 @@
           "socket_family": "unix"
         },
         "commands": {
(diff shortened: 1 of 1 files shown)


lolautruche:
I don't know about #2011, but I added the change

PR #2463, opened 2022-07-11
🐛 Get Poetry flowing again

CollierCZ:
## Why

Poetry is failing, breaking builds.

## What's changed

* Updated Poetry version
* Made install of Poetry global, rather than user
* Kept user-based installs for packages to keep them working


trolologuy (approved):
🚀  ✅ 

PR #2466, opened 2022-07-13
Fix Relay install docs

tillkruss:
The `install-relay.sh` script requires the `v` prefix to generate the correct URL.

Instead of prefixing it in the script itself, changing the argument allows for `v0.4.3` or `dev` to be passed.

### DIFF
--- a/docs/src/languages/php/redis.md
+++ b/docs/src/languages/php/redis.md
@@ -54,7 +54,7 @@ hooks:
     build: |
         set -e
         # Install Relay v0.4.3:
-        curl -fsS https://raw.githubusercontent.com/platformsh/snippets/main/src/install-relay.sh | { bash /dev/fd/3 0.4.3 ; } 3<&0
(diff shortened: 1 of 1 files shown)


tillkruss:
Ping @thomasdiluccio.

thomasdiluccio (approved):
I'll update the script so it can handle versions without the v

docs/src/languages/php/redis.md
```diff
@@ -71,7 +71,7 @@ That's only for pre-built extensions.
 1. Download the Relay/PhpRedis source code.
 2. Check out the version specified in the build hook.
 3. Compile the extension.
-4. Copy the resulting `redis.so` file to [your app root](../../create-apps/app-reference.md#root-directory).
+4. Copy the resulting `relay.so`/`redis.so` file to [your app root](../../create-apps/app-reference.md#root-directory).
```
CollierCZ, line 74: Thought I committed this change :facepalm: Thanks, @tillkruss

PR #2475, opened 2022-07-27
📝 Restore fr-1 region

CollierCZ:
## Why

We've had multiple requests for region info for the fr-1 region, although it's not available.

## What's changed

Added the region back


trolologuy (approved):
✅ 

PR #2484, opened 2022-08-02
📝 Add Java 17

CollierCZ:
## Why

Java is available in version 17

## What's changed

Noted so


trolologuy (approved):
✅ 

PR #2485, opened 2022-08-04
🚨 Improve search scraping

CollierCZ:
## Why

Scrapy was out of date and the cron job was returning errors about parsing images as text ([example](https://console.platform.sh/psh-internal-dev/652soceglkw4u/main/log/iip4s7fxx5v2m)).

Closes #2478

## What's changed

Updated Scrapy and added a middleware to ignore requests for images.

There's still an error with the lack of a title on this page: https://platform.sh/regions/german/fur-retail/
But that's not an important one on our side (and I reported it).

The script can be tested directly by:

```bash
$ platform ssh
...
$ bash index_external_sites.sh 
```

### DIFF
--- a/search/docs/middlewares.py
+++ b/search/docs/middlewares.py
@@ -6,6 +6,7 @@
 # https://docs.scrapy.org/en/latest/topics/spider-middleware.html
 
 from scrapy import signals
+from scrapy.exceptions import IgnoreRequest
 
 
 class DocsSpiderMiddleware(object):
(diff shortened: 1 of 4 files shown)


trolologuy (approved):
✅ 

Looks good to me. Are there any specific use-cases to test the search against?
The searches I did return the expected results.

CollierCZ:
I think it's just whether the website and such are indexed

PR #2489, opened 2022-08-05
💄 Keep tables inside content wrapper

CollierCZ:
## Why

At smaller widths, the table of contents overlapped with tables that took up more space than the content wrapper (the width of content other than tables).

## What's changed

Made tables stay inside the wrapper and code inside tables back to wrapping when appropriate.

Some pages to compare at smaller widths to see if the change is a positive one:
| Old | New|
|----| ----|
| https://docs.platform.sh/add-services/vault.html#policies | https://toc-overlap-inet3ea-652soceglkw4u.eu-3.platformsh.site/add-services/vault.html#policies |
| https://docs.platform.sh/add-services.html#service-options | https://toc-overlap-inet3ea-652soceglkw4u.eu-3.platformsh.site/add-services.html#service-options |
| https://docs.platform.sh/development/regions.html#europe | https://toc-overlap-inet3ea-652soceglkw4u.eu-3.platformsh.site/development/regions.html#europe |

### DIFF
--- a/docs/config/_default/params.yaml
+++ b/docs/config/_default/params.yaml
@@ -1,6 +1,6 @@
 
 # Basics
-version: "1.0.8"
+version: "1.0.9"
 description: Platform.sh User Documentation
 author: Platform.sh
 title: User documentation
--- a/docs/static/styles/user-customizations.css
+++ b/docs/static/styles/user-customizations.css
@@ -404,3 +404,7 @@ kbd {
   padding: 1em;
   margin-bottom: 1em;
 }
+
+table {
+  table-layout: fixed; /* Keep tables inside the content wrapper */
+}
--- a/docs/themes/avocadocs/static/styles/customizations.css
+++ b/docs/themes/avocadocs/static/styles/customizations.css
@@ -101,10 +101,6 @@ code {
   display: inline-block;
 }
 
-table code {
-  white-space: nowrap;
-}
-
 .chroma {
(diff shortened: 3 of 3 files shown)


trolologuy:
⛏️ 
Indeed, it works as expected, and in most cases it's better, but it makes some parts (headline or even content) of the table harder to read (and shrinks the width sometimes too much).
Tested with the [ResponsivelyApp](https://responsively.app/
Screenshot 2022-08-05 at 13 57 37
)

Maybe making the content have a deeper z-index and be side/scrollable might allow keeping the readability whilst still fixing the issue? Or what do you think?

BTW thanks for adding the table with the links to test in the description, much appreciated ;)

CollierCZ:
Can you get those states in an actual browser? Because I tried in Firefox and Chromium and that's not what I see. I tend to test in only my browsers though.

Also, some of those are for widths that we don't optimize for at all currently. The Fold will have many other such problems.

trolologuy:
Yeah, I did test on brave with different screen sizes and I could reproduce the same as on the app:

Screenshot 2022-08-08 at 07 15 47
 
But indeed, "Mobile first" is probably not our main concern, especially since the end product is not meant for mobile, reading the docs on the go has a limited value to it as well.

trolologuy (approved):
Since the behaviour is as expected and mobile first or smaller resolutions are not our main concern, safe to ✅ 

PR #2490, opened 2022-08-05
♻️ Standardize Console and CLI as administration

CollierCZ:
## Why

The console was referred to in various ways. So it's probably best to standardize as a proper noun.

And also to think of it as a valid option on the same level as the CLI.

See also Context.

## What's changed

- Standardized as Console.
- Moved CLI under administration.


trolologuy (approved):
✅ 

PR #2494, opened 2022-08-11
📝 Note live backups available in Console

CollierCZ:
## Why

Live backups will soon have a checkbox in Console.

## What's changed

Updated docs to make the option clear, including a little more clarity in previous steps.


trolologuy (approved):
✅  Great PR, clear instructions, good explanation. Ready to 🚢 

💡 I don't remember if we had a discussion about it, but maybe for the console steps, in addition to the icons, it could be nice to fake the buttons? To have the same color scheme, ... or would it be too confusing? It felt weird to repeat click on "backup" here, "backup" there, "backup" here, ...
This example might be just the exception that confirms the rule, or might be a good way to make the explanations more intuitive overall (better QoL for the customer) by having a better match between the docs and the console UI without having to deal with screenshots and the like (kinda low-ish hanging fruit). But it's really just a detail, and definitely not a blocker for this PR.

CollierCZ:
If we shared a framework, it'd be easy to replicate the buttons. But we'd have to do it manually and then keep track of changes and such, so it wouldn't be as easy as just the icons. I don't think the value to users would be worth the effort at the moment, but we can revisit it later.

PR #2495, opened 2022-08-11
🐛 Remove feedback from sidebar

CollierCZ:
## Why

It was never meant to be there. And feedback shouldn't be plural. Note sure where the s is coming from.

## What's changed

Made it be ignored for the sidebar

### DIFF
--- a/docs/src/feedback/_index.md
+++ b/docs/src/feedback/_index.md
@@ -1,0 +1,3 @@
+---
+sidebarIgnore: true
+---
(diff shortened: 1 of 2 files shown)


trolologuy (approved):
Indeed, the [new version](https://no-feed-in-side-imayt6y-652soceglkw4u.eu-3.platformsh.site/feedback.html) looks better than [the previous one](https://docs.platform.sh/feedback.html).

✅ 

PR #2500, opened 2022-08-16
🐛 Fix location for PHP access log format settings

CollierCZ:
## Why

We gave the wrong location of the access log format. See context.

## What's changed

Fixed the location and response.


trolologuy (approved):
✅ 

PR #2502, opened 2022-08-16
🐛 Fix misplaced flag in command

1t53c5v03:
changed php command so -S actually uses localhost:8001 instead of -d

## Why

Closes #{ISSUE_NUMBER}


CollierCZ:
Thanks for the fix, @1t53c5v03

PR #2503, opened 2022-08-18
💄Make regions table more responsive

CollierCZ:
## Why

The [regions tables](https://docs.platform.sh/development/regions.html#europe) are really hard to use at small widths (even between 900 and 1200 px with the ToC and content wrapper).

## What's changed

At smaller widths, changed layout to a stack rather than horizontal and added tags to replace the headings.


CollierCZ:
Regions table: https://responsive-regions-jt66ria-652soceglkw4u.eu-3.platformsh.site/development/regions.html#europe

trolologuy (approved):
✅ 
For small resolutions, I really like the additional "tag" for Region, ... it adds a lot of clarity.
Layout is also much clearer, and it's easier to get the content.

🤔  Is there a specific reason why the ordering was changed though?

CollierCZ:
If I had to guess, I'd say it's because the sorting beyond available or not was somewhat random. I added a commit that should sort by the URL alphabetically so it should be standard going forward. Let me know if you think some other sorting would make sense.

trolologuy (approved):
It's "weird" for the fr-1 region to be "lost" amongst the ireland regions, but that's a detail.
Would it make more sense to list all region by order?
- eu
- eu-2
- eu-4
- ...
- fr-1
- fr-3

The current form where  "Available to new orgs" takes precedence over the name/url of the region is also perfectly meaningful. So either way works for me.
✅

CollierCZ:
We had it before as just alphabetical, but given that people specifically requested greater insight into what regions are available, I think it makes sense to keep it sorted that way.

PR #2505, opened 2022-08-18
Minor text issue in NodeJS doc

ramit-mitra:
## Why
- `If you have a package.json file in your code, the default build flavor is run` sounds a little confusing
- Instead, the doc should say "the default build flavor is `npm`" instead of "the default build flavor is `run`" as it's much more aligned to the expected build behaviour 

## What's changed
- docs/src/languages/nodejs/_index.md
- Impacted URL - https://docs.platform.sh/languages/nodejs.html#dependencies

### DIFF
--- a/docs/src/languages/nodejs/_index.md
+++ b/docs/src/languages/nodejs/_index.md
@@ -120,7 +120,7 @@ web:
 ## Dependencies
 
 By default, Platform.sh assumes you're using npm as a package manager.
-If you have a `package.json` file in your code, the default [build flavor is run](../../create-apps/app-reference.md#build):
+If you have a `package.json` file in your code, the default [build flavor is npm](../../create-apps/app-reference.md#build):
 
 ```bash
 npm prune --userconfig .npmrc && npm install --userconfig .npmrc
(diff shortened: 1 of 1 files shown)


docs/src/languages/nodejs/_index.md
```diff
@@ -120,7 +120,7 @@ web:
 ## Dependencies
 
 By default, Platform.sh assumes you're using npm as a package manager.
-If you have a `package.json` file in your code, the default [build flavor is run](../../create-apps/app-reference.md#build):
+If you have a `package.json` file in your code, the default [build flavor is npm](../../create-apps/app-reference.md#build):
```
CollierCZ, line 123: :hammer_and_wrench: 
```suggestion
If your code has a `package.json`, the following command is run as part of the default [build flavor](../../create-apps/app-reference.md#build):
```

CollierCZ (changes requested):
Hmm. I can see how it can be confusing. And thanks for pointing it out! But the build flavor doesn't default to something else if there's no `package.json`, so I feel like the suggestion might be confusing in a different way.

I guess the idea is that the command is run as part of the default build flavor. So maybe if it were changed like the suggestion? Not completely set on that or anything, just thinking it through.

ramit-mitra:
@CollierCZ  I agree with your observation.

PR #2511, opened 2022-08-23
📝 Rework default environment guide

### LINKED ISSUE #2375: 📝 Rework default branch guide

https://docs.platform.sh/environments/default-environment.html

The guide is written from the perspective of a new project. Rewrite so it also is clear for projects with data and settings in them. Things to consider there:

* Add warnings about environment variables and settings
* Maybe add a note to remind to lower the DNS TTL before starting the process
* Maybe not need to repeat `-p <PROJECT_ID>` so often, or at least at end for easier copying and pasting when not needed

In general, update for:

- Content accuracy
- Includes what it should
- Fits template
- Fits style guide

### Additional context

_No response_

### LINKED ISSUE #2504: Uninherited variables when renaming the default branch

- https://docs.platform.sh/environments/default-environment.html#1-create-the-main-environment
- https://docs.platform.sh/environments/default-environment.html#1-create-the-main-branch

If a production-only environment variable is set (`--inheritable=false`), and the steps in this guide are followed, the "new" production environment will potentially not receive important credentials during the switch (as an example, `FASTLY_API_TOKEN`). 

So as an update, I can think of two options
- A warning: place a note in each of these sections, letting users know of this gotcha.
- A script: sort of a force inherit script, that loops through production environment variables, setting them in the `main` environment. 

### Additional context

_No response_

CollierCZ:
## Why

Closes #2375
Closes #2504

## What's changed

Changed the guide content to fit the template.
Because a lot of the content was repeated between sections, switched to codetabs.
Because that would involve a lot of switching between codetabs, made the option to have codetabs switch on a page basis.


docs/src/environments/default-environment.md
```diff
@@ -1,284 +1,183 @@
...
-To start, create a new environment off `master` called `main`:
-
-```bash
-$ platform environment:branch main master --title=Main -p <PROJECT_ID> --force
-```
-
-{{< note >}}
+- Switch the default environment during non-peak hours.
+- Keep your data by taking a [backup of the `old` environment](../administration/backup-and-restore.md)
+- Reduce your DNS time-to-live (TTL) to a minimum.
```
trolologuy, line 22: :question: Why do we recommend lowering the TTL DNS settings? 
I assume that the `edge_hostname` needs to be updated on the DNS level (didn't check the whole process, but can do if needed), if so, that step is missing from the page.

trolologuy (changes requested):

docs/src/environments/default-environment.md
CollierCZ: That's a good point. I guess it wasn't there in the previous one so I skipped it. Probably needs to be for live environments though. Added a step that I hope covers it.

trolologuy: Wonderful, thank you! It should be clear enough now.

trolologuy (approved):
✅  Great PR, I love how the codetab button now changes content throughout the page!

- [x] The information is clear.- [ ] All links work, including to section headers.
- [x] Any relevant issues raised by the linting have been addressed.
- [x] The page fits the content model defined through the templates.

PR #2512, opened 2022-08-23
Fix typo

trsteel88:
Fix composer lock filename typo


CollierCZ (approved):
Thanks for the fix, @trsteel88!

PR #2514, opened 2022-08-23
Update redis.md

### LINKED ISSUE #2513: Clarify Redis+WordPress sample code is dependent on Config reader

https://docs.platform.sh/guides/wordpress/redis.html

The code snippets included in the [Configuration section](https://docs.platform.sh/guides/wordpress/redis.html#configuration) assumes the user is using the Platform.sh Configuration Reader (PHP) and has named the object `$config`. Users who are not using the config reader, or have named it differently and copy & paste the code snippet as-is will receive an error and the redis service will not be accessible in the WordPress instance.

### Additional context

_No response_

gilzow:
Closes #2513

## Why

Closes #2513 

## What's changed
Adds note and link that the included code snippets assume user is using the Platform.sh Config Reader (PHP)


CollierCZ (approved):
Thanks for the improvement, @gilzow!

PR #2517, opened 2022-08-24
📝 Document limits to disk downsizing

### LINKED ISSUE #2473: 📝 Clarify limits to downsizing

https://docs.platform.sh/create-apps/app-reference.html
https://docs.platform.sh/add-services.html#disk
Maybe elsewhere

For potential downsizing of disks, two things should be explicit:

- downsizing a volume means backups from before the downsize are incompatible and can no longer be used
- a downsize fails if there is more data on the volume than the desired size

### Additional context

Applies to anything with a disk, so check if other places need a link

CollierCZ:
## Why

Closes #2473

## What's changed

- Updated legacy-regions shortcode to use smaller headings
- Added a shortcode to repeat disk warning for services and apps
- Used the shortcode twice and linked to it one other time

### DIFF
--- a/docs/layouts/shortcodes/disk-downsize.md
+++ b/docs/layouts/shortcodes/disk-downsize.md
@@ -1,0 +1,7 @@
+{{ $type := .Get "type" }}
+You can decrease the size of an existing disk for a{{ if eq $type "app" }}n{{end}} {{ $type }}.
+If you do so, be aware that:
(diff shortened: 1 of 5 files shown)


trolologuy (approved):
✅ Great change. The steps are clear and brief. The short code is used in all places where it's needed.

Issue #2186, opened 2022-02-21
🔬 List 🐛  and 💡 for the UI/UX


trolologuy:
### Where on docs.platform.sh should be changed?

https://docs.platform.sh

### What exactly should be updated?

(edit the issue to add new items)

- Clicking an anchor often hides the title of said anchor outside of the displayed screen area.
- `code blocks` styling (especially in a table) can be confused for a link.
- Codetabs styling might be clearer from a UX standpoint.

CollierCZ:
I wouldn't display the date of the last change. I don't think automated changes add much value and often is misleading (when I move a page I might have to update a page with a link to it and so that page would be shown as updated even when its content hasn't been verified or anything). And we have too few people working on it to make changes frequently enough that it would make sense manually.

CollierCZ:
> * Clicking on a first-level menu should open the overview page ...
> * Clearly highlight if an element in the navbar has subelements ...

These are going to be resolved in upcoming changes to the information architecture. We already have some potential designs for it.

I'd add that the ToC overlaps with content at certain widths.

CollierCZ:
I went through the list and reorganized a bit (removing some items already done).

Also this bit is already possible:

> When a user is on the bottom of a [section page](https://docs.platform.sh/development/cli.html#autocomplete-commands) that has other children, these could be [listed](https://docs.platform.sh/development/cli/api-tokens.html) at the bottom of the page.

You just need to change the `layout` in the frontmatter. Either remove it so it goes back to the default or set it to `list`.

trolologuy:
Allow nesting of templates when they are related.
For example for Drupal, that would give:
- [Drupal 8](https://github.com/platformsh-templates/drupal8)
- [Drupal 9](https://github.com/platformsh-templates/drupal9)
- [Drupal 8 (multisite variant)](https://github.com/platformsh-templates/drupal8-multisite)
- [Opigno](https://github.com/platformsh-templates/drupal8-opigno)
- [GovCMS8](https://github.com/platformsh-templates/drupal8-govcms8)
Instead of having a super lengthy list of Templates.

CollierCZ:
Removed some of the items done in #2785 and added the notes from #2785

PR #2519, opened 2022-08-26
📝 Deprecate old Elixir versions

CollierCZ:
## Why

Older versions aren't supported anymore. See Context.

## What's changed

Moved 1.9 and 1.10 to `deprecated`.

### DIFF
--- a/docs/data/registry.json
+++ b/docs/data/registry.json
@@ -189,10 +189,11 @@
     "runtime": true,
     "type": "elixir",
     "versions": {
-      "deprecated": [],
-      "supported": [
+      "deprecated": [
         "1.9",
-        "1.10",
+        "1.10"
+      ],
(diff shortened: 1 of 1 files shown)


trolologuy (approved):
✅ 

⛏️ The spacing of bullet points in the list of supported versions exception (eu and us regions) is [not displayed as expected](https://docs.platform.sh/languages/elixir.html#supported-versions). There is too much padding. The same applies [for services](https://docs.platform.sh/add-services/headless-chrome.html#supported-versions). 

CollierCZ:
That's actually how I expect it as it's been that way for a long time. I feel like you may have mentioned something similar on a previous PR, but can't find it so maybe not. Can add it to #2186 if you want us to come back to it at some point.

PR #2522, opened 2022-08-31
🐛 Fix improper version in app example

CollierCZ:
## Why

The minimal version of an app config showed a nonexisting version for Node.js.

## What's changed

Updated it to the latest version.


trolologuy (approved):
✅ 

PR #2525, opened 2022-08-31
Mark geoip as available for PHP 8.0 and 8.1

ederlf:
## Why

Geoip is available for PHP 8.x in our images, so we should update the docs to reflect it.

## What's changed

Mark the geoip plugins as available for PHP 8.x.


CollierCZ (approved):

CollierCZ:
Thanks for the PR, @ederlf!

PR #2526, opened 2022-09-01
🐛 Fix broken link

CollierCZ:
## Why

Relates to #2213 

## What's changed

Changed a link from latest to the latest specific version
Considered removing the whole page. Will come back to that as some things may be useful and would need to investigate.


trolologuy (approved):

PR #2539, opened 2022-09-12
:pencil2: Fix typo - add missing word

nicogommen:
## Why

Closes #{ISSUE_NUMBER}


CollierCZ (approved):
Thanks for the fix, @nicogommen! And with the right emoji in the title, too :heart_eyes:

PR #2543, opened 2022-09-14
Update support solr versions for dedicated legacy

raupie:
We now have 8.11 Solr for Dedicated Generation 2


CollierCZ (approved):
Thanks!

PR #2549, opened 2022-09-16
Update _index.md

bligneri:
Added a section on how to use the CLI in a script in non interavtive mode using the PLATFORMSH_CLI_NO_INTERACTION=1; environment variable

## Why

Closes #{ISSUE_NUMBER}


CollierCZ (approved):
Thanks for the suggestion, @bligneri! I moved it down to be an example of customizing the CLI and changed it a bit to fit the general style of the docs. It seems like a useful example, so hopefully people will get to use it now.

PR #2555, opened 2022-09-19
Fixing encoding issues

### LINKED ISSUE #2550: 🐛 Properly encode HTML in search results

https://docs.platform.sh/search.html
And all search results

Shouldn't get results like `Let&rsquote;s Encrypt`. Should be properly encoded as HTML. See Context.

ramit-mitra:
## Why

Closes #2550

## What's changed
- Using `DOMPurify` and `dangerouslySetInnerHTML` to properly render encoded strings
- Minor formatting fixes


CollierCZ (changes requested):
Thanks for the PR, @ramit-mitra! Appreciate it. And it seems like a clean and simple solution to the issue.

In this case, we're adding an additional dependency to solve something that's being currently solved elsewhere with a different dependency (the lines just above with `dangerouslySetInnerHTML` and `DOMPurify`. I think it'd be best to have only a single approach/dependency to the issue.

ramit-mitra:
Hi @CollierCZ thanks for the feedback. I have updated the changes. Please review. Thank you.

PR #2557, opened 2022-09-20
📝 Define standards for examples

### LINKED ISSUE #2398: 📝 Define standards for examples
We'd like to be consistent in how we present examples so they're clear and easy to see they're examples and what patterns exist. Defining patterns for Dedicated and Grid and such.

This task should just define the standard, not necessarily implement it everywhere.

CollierCZ:
## Why

Closes #2398

## What's changed

Added a shortcode to be able to define variables (the exact styling can then be updated later across all instances)
Added standards for what to use in examples


docs/static/styles/user-customizations.css
```diff
@@ -468,4 +468,12 @@ kbd {
...
-}
\ No newline at end of file
+}
+
+/* Styling for variables inside code blocks and such */
+var {
+  color: var(--brand-color-primary-blue);
+  text-decoration: underline var(--brand-color-primary-blue) dotted 2px;
+  font-style: italic;
+  cursor: help;
```
trolologuy, line 478: 😻 Love the display of the help text!!!
Great idea and great implementation!

:nail_care: 
Would making it bold further help to differentiate it from the rest?

💡 Is it possible to show the overlay without delay? (did not check if that's something that can be overridden or not).

💡 Is blue the colour of choice there? Since it's hard to separate it from all the links we already have on docs pages.

contributing/content-style.md
```diff
@@ -205,6 +206,30 @@ You should follow four rules | There are four rules that should be observed
 Your requests are redirected | The system redirects your requests
 This is caused by            | The cause of this issue has to do with
 
+## Use clear examples
```
trolologuy, line 209: ⛏️  It might be nice to add a note about the "variables in code" part, since IIRC sometimes we provide code (with variables) followed by examples.

trolologuy (approved):
✅ Great addition 🥇 Love the shortcode and it's display.
A few ideas here and there but nothing blocking.

💡 Do we want to check that everything that could be a variable has the right shortcode?

contributing/content-style.md
CollierCZ: There's a link to the section in the Markup Style Guide. I figured the way it was done was more of a markup question rather than a content question.

docs/static/styles/user-customizations.css
CollierCZ: I think the implementation of the tooltip is from the browser, so not sure about changing the delay. 

With that and the other style questions, I think we'll revisit when we do a more comprehensive redesign. It's not ideal, but probably enough for now.

PR #2558, opened 2022-09-20
🚚 Move names to variables

CollierCZ:
## Why

We were repeating things like plan and region names at various places in the docs.
These names and offerings could change in the future. Having them in multiple places makes it harder to update them.

## What's changed

Moved plan size names to variables that can be updated in one place. Removed explicit mentions of region names.


fabpot (approved):
LGTM

PR #2560, opened 2022-09-20
🐛 Fix info on PR workflow

CollierCZ:
## Why

Git wasn't checked out on the workflow for security reasons (and it wasn't necessary for most steps). But it's needed for the step to get the differences in files.

Also, the CLI [complained] about the project not being set for a PR from a fork.

## What's changed

Check out Git for PRs from default repo and approved PRs

Added a step to building from forks to set the remote.


trolologuy (approved):
✅  Let's test it!
Should I open a test-fork once this one is merged and add the `build-fork` label to it?

CollierCZ:
There's already a PR that needs to be built, so no need. That's how I found the error :slightly_smiling_face:

PR #2564, opened 2022-09-21
💚Process improvements

CollierCZ:
## Why
Forked PRs are still problematic:
- The URL is in a different object in JSON
- The authentication is weird

## What's changed

- Put in a switch for forks for the URL
- Tried a different approach for activating a Git branch to try not to need to authenticate through Platform, though I might have it exactly backward as to where the problem is...


trolologuy:
> Hmm, clearly need to reexamine the changes again

Yeah the commenting part is doing weird stuff

trolologuy (approved):
✅  Let's try it out.

PR #2565, opened 2022-09-21
✅ Use token with workflows permission

CollierCZ:
## Why

Pushing with Git isn't possible without higher permissions.

## What's changed

Trying with a Personal Access Token that will expire in 90 days.


trolologuy (approved):
✅ Third time's the charm (hopefully) 🌞

PR #2566, opened 2022-09-21
✅ Only activate inactive environments

CollierCZ:
## Why

Ran into an error when the environment was already active

## What's changed

Put in a condition to only activate inactive environments


trolologuy (approved):
✅ 🚢 :shipit:

PR #2568, opened 2022-09-21
Add MongoDB Enterprise 4.0, 4.2 and 4.4 to supported versions

remilejeune2:
## Why

4.0, 4.2 and 4.4 versions were not documented.

## What's changed

4.0 version goes directly in the deprecated section. It's only there to provide a migration path from 3.6 to 5.0.
I added the deprecated part on the Enterprise edition section.


CollierCZ (approved):
Thanks for the PR, @remilejeune2!

PR #2569, opened 2022-09-22
💚 Fixes to process

CollierCZ:
## Why

Some bugs in the new process

## What's changed

- Fixed conditional syntax for checkout (was running on forks when it shouldn't)
- Stopped running activate env on forks (now handled by `build-from-fork` workflow)
- Update base for Git diff comparison (to compare only changes in the PR not to new changes in `main`)


trolologuy (approved):
✅  🤞🏼

PR #2570, opened 2022-09-22
🐛 Replace hardcoded value with variable

CollierCZ:
## Why

Accidentally left in a specific PR number from testing :facepalm:

## What's changed

Replaced with variable


trolologuy (approved):
✅

I should have catched this one 😓

PR #2571, opened 2022-09-22
🐛 Get the URL from the right place

CollierCZ:
## Why

Was wrong about the payload of the events from forks, as I can see once the fork is working well.

## What's changed

Switched back to the same payload for all


trolologuy (approved):
✅

PR #2572, opened 2022-09-22
📝 Update images and sort

CollierCZ:
## Why

InfluxDB image wasn't up to date.
Also realized we're presenting the oldest versions first in the available list.

## What's changed

Updated list and sorted with newest versions first.


trolologuy (changes requested):
😍 good call for the order, it makes much more sense to have examples with the latest available versions!

A few details here and there, other than that everything works as expected ✅ 

❓ What's surprising is the high disparity of versions between Grid, Dedicated and Dedicated Generation 3 for [solr](https://image-updates-7h3bk3i-652soceglkw4u.eu-3.platformsh.site/add-services/solr.html) both in supported and deprecated compared to the rest. Probably expected, but looked weird, I did not investigate that further.

trolologuy (approved):
✅

PR #2579, opened 2022-09-23
🚚 Move npm workspace to docs

CollierCZ:
## Why

The remaining lock files in various directories (like the `xss` directory) were confusing as npm was installing files differently locally than what was in those files. Given that Platform.sh works in just the `docs` directory, it seemed best to move all necessary context there.

## What's changed

- Moved the workspace from the root to the `docs` directory.
- Updated some dependencies


trolologuy (approved):
✅ The local and remote build work, the steps have been updated.

PR #2589, opened 2022-09-29
🔥 Deprecate old images

CollierCZ:
## Why

Versions were past end of life:

- [Redis](https://docs.redis.com/latest/rs/installing-upgrading/product-lifecycle/)
- [Solr](https://solr.apache.org/downloads.html)

## What's changed

Moved old versions to deprecated


docs/data/registry.json
```diff
@@ -944,7 +941,10 @@
         "6.0",
         "5.0"
       ],
-      "deprecated": ["3.2"]
+      "deprecated": [
```
trolologuy, line 944: :question:
Why is version 5.0 not considered deprecated on enterprise?

From the [release cycle of redis](https://redis.io/docs/about/releases/#support), I think we should consider 
- 6.0 + 6.2 (critical bugs and major security issues)
-  7.x 
as supported and the rest as EOL.
So maybe we should add 6.0 as supported?

docs/data/registry.json
```diff
@@ -944,7 +941,10 @@
...
+        "5.0",
+        "3.2"
+      ]
```
trolologuy, line 947: :lady_beetle:
Dedicated Generation 3 hasn't changed, whereas it should have the same split as grid.

docs/data/registry.json
```diff
@@ -1037,23 +1037,25 @@
...
-        "7.7"
+        "8.11"
       ]
     },
     "versions-dedicated": {
       "supported": [
-        "8.11",
+        "8.11"
+      ],
+      "deprecated":[
```
trolologuy, line 1058: :lady_beetle:
8.6 and 6.6 for Dedicated Generation 3 should be moved to deprecated.
since [<8.11 - All older versions are End Of Life (EOL)](https://solr.apache.org/downloads.html)

trolologuy (changes requested):
🐛 In the [order of versions for the region's availability](https://image-updates-7h3bk3i-652soceglkw4u.eu-3.platformsh.site/add-services/redis.html).

🐛 Dedicated Generation 3 is not consistent with Grid.

- [ ] Final built pages are working as expected.
- [ ] The information is correct.

docs/data/registry.json
CollierCZ: 5.0 was deprecated, as seen in the next line.

As I [linked in the main post of the PR](https://docs.redis.com/latest/rs/installing-upgrading/product-lifecycle/#endoflife-schedule), 6.0 already reached its EOL (in May). Will add it here as well.

trolologuy (changes requested):
:bug There's still an issue with the display:
```
The following versions aren’t available in the EU-1 and US-1 regions:
- 6.0
- 7.0
- 6.2
Consider [region migration](https://image-updates-7h3bk3i-652soceglkw4u.eu-3.platformsh.site/guides/general/region-migration.html) if your project is in those regions.
```

Is the scope of that PR only redis and solr? 
If yes, the following is not a blocker.

🐛  We also display:
- [too many versions of RabbitMQ](https://www.rabbitmq.com/versions.html), `3.9` is the only version that should be displayed.
- [PostgreSQL](https://endoflife.software/applications/databases/postgresql), where `9.6 ` should be removed.
- [MariaDB](https://mariadb.org/tag/eol/) where `10.0`, `10.1` and `10.2` should be removed
- [Kafka](https://kafka.apache.org/downloads) and InfluxDB are a bit unclear

trolologuy (approved):
✅ Looks better now, thank you 🙏🏼 😻

PR #2591, opened 2022-09-30
🐛 Make shortcodes in codetabs work

CollierCZ:
## Why

Shortcodes in codetabs weren't working ideally.

## What's changed

Fixed them for the icons and updated advice for variables, where HTML seems necessary (at least within a reasonable amount of time for finding a solution)

### DIFF
--- a/docs/src/administration/backup-and-restore.md
+++ b/docs/src/administration/backup-and-restore.md
@@ -64,13 +64,10 @@ file=none
 highlight=false
 ---
 
-<!--This is in HTML to get the icon not to break the list. -->
-<ol>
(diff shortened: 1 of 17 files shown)


docs/src/administration/backup-and-restore.md
```diff
@@ -225,11 +219,12 @@ highlight=false
 ---
 
 1. Run the following command:
```
trolologuy, line 221: :nail_care:
There are extra return carriages around the line.

trolologuy (approved):
✅  Besides the extra return carriages, it works as expected (and is easier to maintain and read, nice 😎 !!!)

PR #2592, opened 2022-10-01
🐛 Fix shortcode for create project link

CollierCZ:
## Why

The shortcode was being processed incorrectly, see [example](https://docs.platform.sh/tutorials/migrating.html#preparation).

## What's changed

Changed it to HTML so the comments and such are processed as such.


trolologuy (approved):
✅  bug is fixed

PR #2593, opened 2022-10-03
🐛 Fix broken links

CollierCZ:
## Why

Relates to #2213

## What's changed

Fixed one broken link and removed an outdated section in another place with a broken links.
That whole doc should be updated, but removed the one link and made the final instruction clearer (since that's where the real details are).


trolologuy (approved):
✅ Ready to 🚢 

PR #2595, opened 2022-10-03
📝 Add example of no build flavor

CollierCZ:
## Why

Some found it confusing how to run their own Composer commands since it's a two-step process and each step is separately documented.

Also, the build flavor can be changed with variables. See Context

## What's changed

Added an example.


trolologuy (approved):
✅ Clear instructions, nothing to add there.

PR #2596, opened 2022-10-03
Show that n on platform is compatible with multiple types of node version declarations

tylerssn:
## Why

This PR shows users that `tj/n` on Platform.sh remains compatible with multiple types of Node version declarations.

## What's changed

Added the Node.js version specification syntax for each file that is compatible with `tj/n`


tylerssn:
Thanks for accepting the proposal!

PR #2597, opened 2022-10-04
📝 Add note on necessary resources

CollierCZ:
## Why

It wasn't perfectly clear that resources were needed for workers

## What's changed

Added a note (similar to what's in the [app reference](https://docs.platform.sh/create-apps/app-reference.html#workers).


AnouckColson (approved):
Looks good to me! 😍

PR #2600, opened 2022-10-05
🐛 Fix partial call

CollierCZ:
## Why

A partial was called improperly resulting it is appearing on the [published site](https://docs.platform.sh/create-apps/app-reference.html#sizes-in-development-environments).

## What's changed

Fixed the call


AnouckColson:
😍 All good!

AnouckColson (approved):

PR #2602, opened 2022-10-05
✨ Add premium badges

CollierCZ:
## Why

We have features that are premium and want to highlight that in the UI.

## What's changed

Moved from the registry parser table for supported versions to a shortcode that is easier to edit in a single repo.
Added badges for premium features in the navigation and the supported services table.


AnouckColson (approved):
😍 I don't see anything that needs changing!

PR #2605, opened 2022-10-06
Do not propose to use a deprecated version

OskarStark:


CollierCZ (approved):
Thanks for the improvement, @OskarStark Hope you don't mind I took the opportunity to also fix the suggestion to use a deprecated PHP version lower on the page. :see_no_evil:

PR #2611, opened 2022-10-10
👷 Improve automated processes

CollierCZ:
## Why

- PRs from forks that were built ended up in a branch that wasn't automatically deleted. Also the workflow to get info on the build failed without a good message, so it wasn't clear why it failed.
- It was possible to merge PRs labelled as upcoming versions. This made it more dangerous to get reviews on upcoming changes as accidental merges might happen.

## What's changed

- Added steps to delete branches when they exist and fail the workflow with a reasonable message
- Added a workflow to fail if the PR is labelled `upcoming-version`. This way reviews can happen before the change and a change in label should later be enough to merge.

### DIFF
--- a/.github/workflows/check-labels.yaml
+++ b/.github/workflows/check-labels.yaml
@@ -1,0 +1,20 @@
+name: Check labels
+on:
+  pull_request_target:
+    branches: [main]
+    types: [labeled,opened,reopened,synchronize]
+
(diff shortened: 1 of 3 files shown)


trolologuy (approved):
✅ The code and the approach look meaningful and it's better to have good safeguards.
Let's try it out.

PR #2622, opened 2022-10-18
🔍️ Make .environment section more visible

CollierCZ:
## Why

Searching for `.environment` brought up pages for specific frameworks, but nothing generic.

## What's changed

Added a keyword.

The remaining issue is that periods are removed. So a search for `.environment` actually searches for `environment`. This makes the results not ideal if people are looking for environment-related ideas. But that's already an issue and I didn't find a way to solve it quickly, so this way at least gets the generic result higher.


AnouckColson (approved):
😍 Looks good to me!

PR #2623, opened 2022-10-18
🏗️ Move Gen 2 below Gen 3

CollierCZ:
## Why

Dedicated Gen 2 is listed above Gen 3, although we're trying to move forward with Gen 3

## What's changed

Moved Gen 2 down


AnouckColson (approved):
😍 Perfect!

PR #2610, opened 2022-10-10
📝 Rework pricing doc

CollierCZ:
## Why

Closes #2574

## What's changed

- Moved pricing to administration and moved order around a bit
- Reworked the page to be smoother and have no prices

PR #2626, opened 2022-10-20
🐛 Fix broken header links

CollierCZ:
## Why

As part of #2610, the headers on the pricing page were changed and one link pointed to the old header. Also somehow a header got missed in #2614.

Relates to #2213 

## What's changed

Changed the headers linked to.


AnouckColson (approved):
All good!

PR #2631, opened 2022-10-24
👷 Process improvements

CollierCZ:
## Why

1. The check for upcoming versions wasn't working
2. The build from the first time the `build-fork` label was added didn't work (subsequent removal and adding did work)
3. Environments from forks weren't being deleted
4. GitHub is [deprecating some syntax](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) we were using

## What's changed

1. Fixed the name of the label
2. Fix the check to look for other statuses that might be there in the first minute or so of a new environment
3. Used the CLI to delete environments when necessary
4. Used [recommended syntax](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter)

### DIFF
--- a/.github/workflows/build-from-fork.yaml
+++ b/.github/workflows/build-from-fork.yaml
@@ -29,6 +29,7 @@ jobs:
           git push --force -u origin pr-${{ github.event.number }}
 
           # If the environment isn't active, activate it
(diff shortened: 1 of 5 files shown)


trolologuy (approved):
The syntax looks updated, let's try it out. ✅

PR #2624, opened 2022-10-18
:memo: Rework configure environments page

AnouckColson:
## Why

Closes #2576 

## What's changed

Reorganized some content and reworked the page for improved consistency + to fit general style.

PR #2641, opened 2022-10-26
🐛 Fix broken link

CollierCZ:
## Why

There was a new link from #2624 that wasn't taken into account in #2627

## What's changed

Fixed the link


AnouckColson (approved):
👍

PR #2645, opened 2022-10-27
🐛 Fix broken links

CollierCZ:
## Why

Relates to #2213

One broken link left from #2624 
One outdated section from #2627

## What's changed

Fixed the header link and removed the note about backups that no longer applies


trolologuy (approved):
✅  Updated links point to the right sections.

🐛 On that branch, search is broken.
🐛 GitHub automated comment also.
But that's probably due to the history of the branch (had to be reactivated several times, ...) so probably not worth digging a lot.

PR #2646, opened 2022-10-27
💚 Process fixes

CollierCZ:
## Why

Accidentally was trying to delete the environment for the Git branch, rather than the created built environment

Also was using a file as an argument, rather than reading the file for the pull request number

## What's changed

Set it to the right environment and stopped waiting for result to avoid timeouts

Read the file to get the correct PR number


CollierCZ:
That comment shows the fix for the PR number works. It just took the last successful run for this test

trolologuy (approved):
✅ As usual, let's try it out :)

PR #2651, opened 2022-10-27
Improve DDEV documentation

rfay:
## Why

The DDEV documentation, especially about pull, was out of date. It's no longer necessary to edit an example file, etc.

## What's changed

General updates, but mostly how to install tokens and other environment and how to pull.


CollierCZ (approved):
Thanks for the change!

PR #2652, opened 2022-10-28
Fix: Note

OskarStark:
Not sure this is meant for version 13, but telling "using postgis is not supported", while in the second sentence "see how to upgrade to 12 using postgis" looks weird to me.

After my change it would make sense, that upgrading to 13 using postgis is not supported, while upgrading to 12 is.


CollierCZ:
I've gotten confirmation that it's still the upgrade from 11 to 12 that isn't directly supported. It involves a change in a major version. I'll try to work on clarifying the page.

CollierCZ (approved):

CollierCZ (approved):

PR #2667, opened 2022-11-07
🐛 Update instructions for properties

CollierCZ:
## Why

The instructions for properties still referenced a cURL request even after the CLI command was updated. (I overlooked that in the update.)

## What's changed

Changed it to reference flags with an example.


AnouckColson (approved):
👍 Perfect

PR #2668, opened 2022-11-07
📝 Add clarity to metrics

### LINKED ISSUE #2637: Improve the explanation for the relationship between container limits (in particular burst) to overall host resource usage for Dedicated Gen 3

https://docs.platform.sh/increase-observability/metrics.html#default-thresholds

The main source of confusion seems to be that on Dedicated Gen 3 environments, metrics are reported per container and resource limits are set per container, even when the customer has the entire host dedicated to them. This can lead to situations where the containers are operating at maximum (or even exceeding) capacity, according to the defined limits.
Customers may wonder why their application is performing in burst mode (thus exceeding their allowed resources) while the host itself still has plenty of resources available. 
We should explain that host resources are shared across all containers and that each container can have their limits customized via the `.platform.app.yaml` file; when the default limits do not work for the particular use case, customer can modify the limits for the affected containers, knowing that this will also change limits for other containers (the underlying host resources will not change, so whatever you grant to one container will have to be removed from another container).
When multiple containers are in burst mode, the total allocation for all the containers should be considered to see if the project would need to big upsized.

See also Context.

### LINKED ISSUE #2647: Document Metrics in the Services tab for Grid

Update the https://docs.platform.sh/increase-observability/metrics/grid.html page to include that we also show metrics within the Services tab, not just on the Metrics page.
This change is specific to Grid section only for now.

Example:
If you go to https://console.platform.sh/olga-linek-olga-linek-olga-linek/p53yqxghmploy/master/services/app, you will see that for every App + Service we show Metrics.

Please add a screenshot of the Metrics within the Service tab + a couple of sentences around this is sufficient.

Proposal:
" Within the Services tab, when clicking on any of the Application & Services, you will also see the latest usage and allocation CPU, RAM and Disk metrics for the container. You can click on the `View metrics` link, which will take you directly to the metrics page for the selected environment." 

### Additional context

Screenshot of what we can see is attached below:
Screenshot 2022-10-27 at 11 56 13

Note: 
This is currently only visible for the Grid as these metrics are specifically per Service, Application and Worker only - which is what we show on the Services tab. For DG3 we have metrics per container per host and for DG2 we have metrics per host.

CollierCZ:
## Why

Closes #2647
Closes #2637


docs/src/increase-observability/metrics/grid.md
```diff
@@ -14,6 +14,9 @@ Infrastructure metrics report CPU, RAM, and disk space for app and worker contai
 and CPU and disk space for service containers.
 These metrics are available for all of your Grid environments.
 
+Summaries of the metrics for each container separately can be seen in the Console under **Services** for a given environment.
```
AnouckColson, line 17: 🛠️ 

```suggestion
You can see summaries of the metrics for each container separately in the Console under **Services** for a given environment.
```

AnouckColson (changes requested):
Just a minor change from passive to active voice, otherwise it's all good 👍

AnouckColson (approved):
👍 🚀

PR #2679, opened 2022-11-09
The token needs admin access to the repository

zargex:
To set the webhooks, it has to be able to change the settings in the repository.

## Why
Write access to the repository  is not enough to set webhooks because those are under the repository settings.
https://confluence.atlassian.com/bitbucketserver/http-access-tokens-939515499.html

## What's changed
"write access" to "admin access"


CollierCZ (approved):
Thanks for the fix, @zargex !

PR #2680, opened 2022-11-10
📝 Clarify HTTP access covers entire environment

CollierCZ:
## Why
It wasn't clear that HTTP access control covered an entire environment and couldn't be set by URL. See Context.

## What's changed

Clarified that.


AnouckColson (approved):
👍

PR #2682, opened 2022-11-10
Document direct connection to MongoDB service via app container

rhubinak:


CollierCZ (approved):
Thanks for the addition, @rhubinak !

PR #2686, opened 2022-11-11
🧑‍💻 Improve developer experience

CollierCZ:
## Why

There were some nagging issues that made it harder to contribute. Like having to use a shortcode AND a page property to use Mermaid diagrams. And variables weren't working in codetabs so had to use ugly HTML. Wanted to simplify things.

Looked into linting for Markdown style, but it would require picking a style and increase the burden on contributors, so will maybe revisit in the future.

## What's changed

- 📝 Added guidance and enforcement on neutral tone
- ♻️ Made it simpler to add Mermaid diagrams
- ♻️ Made adding code block locations easier
- ✨ Got variables working inside other shortcodes


AnouckColson (approved):
LGTM 🚀

PR #2695, opened 2022-11-16
✏️ Add missing trailing slash

CollierCZ:
## Why

A slash was missing at the end of the example so it wouldn't match the regex.

## What's changed

Added the slash


AnouckColson (approved):
🚀

PR #2696, opened 2022-11-16
📝 Make it clear logs in files can't be forwarded

CollierCZ:
## Why

It wasn't clear that logs in files can't be forwarded. See Context.

## What's changed

Added a note about that.


AnouckColson (approved):
👍 Approved!

PR #2702, opened 2022-11-21
:memo: Rework PHP-FPM sizing

### LINKED ISSUE #2633: 📝 Update PHP-FPM sizing guide

https://docs.platform.sh/languages/php/fpm.html

Update for:

- Content accuracy
- Includes what it should
- Fits template
- Fits style guide

### Additional context

_No response_

AnouckColson:
## Why

Closes #2633 

## What's changed

Reworked the page according to our concept template. Added a glossary and reorganized information while striving for less complex sentences.


docs/src/languages/php/fpm.md
```diff
@@ -3,70 +3,107 @@ title: "PHP-FPM sizing"
...
+and look at the PHP access log:
 
 ```bash
 less /var/log/php.access.log
 ```
 
-In the fifth column, you'll see the peak memory usage that occurred while each request was handled. The peak usage will probably vary between requests, but to avoid the severe performance costs that come from swapping, your size hint should be somewhere between the average and worst case memory usages that you observe.
+In the fifth column of the output, 
+you can see the peak memory usage that occurred when each request was handled. 
+The optimal request memory is somewhere between the average and worst case memory usages listed.
```
CollierCZ, line 94: :lady_beetle: `somewhere between` seems rather vague. I'm not sure how useful it really is.

docs/src/languages/php/fpm.md
```diff
@@ -3,70 +3,107 @@ title: "PHP-FPM sizing"
...
 
 ```bash
 tail -n5000 /var/log/php.access.log | awk '{print $6}' | sort -n | uniq -c
 ```
 
-This will print out a table of how many requests used how much memory, in KB, for the last 5000 requests that reached PHP-FPM.
-(On an especially busy site you may need to increase that number).
-As an example, consider the following output:
+Note that the above command takes into account the last 5000 requests that reached PHP-FPM.
+You can adjust this number depending on the amount of traffic on your site.
```
CollierCZ, line 103: :+1: :hammer_and_wrench: This is more clear that the original, for sure. We do try to avoid directional language like `above`.
```suggestion
This command takes into account the last 5,000 requests that reached PHP-FPM.
You can adjust this number depending on the amount of traffic on your site.
```

docs/src/languages/php/fpm.md
```diff
@@ -3,70 +3,107 @@ title: "PHP-FPM sizing"
...
+
+In the output, you can see how many requests used how much memory, in KB.
+For example:
 
 ```text
       1
```
CollierCZ, line 109: :question: Why is there one request without a memory log? I think it should be removed, but I don't understand it.

docs/src/languages/php/fpm.md
```diff
@@ -83,21 +120,29 @@ As an example, consider the following output:
...
 
-This indicates that the majority of requests (4800) used 2048 KB of memory.
-In this case that's likely application caching at work.
-Most requests used up to around 10 MB of memory, while a few used as much as 18 MB and a very few (6 requests) peaked at 131 MB.
-(In this example those are probably cache clears.)
+In the above output:
+
+- The majority of requests (4800) peaked at 2048 KB of memory, 
+  likely thanks to application caching.
+- Most other requests (3206) used up to around 10 MB of memory.
```
CollierCZ, line 127: :+1: Clearer as a list, for sure.

:clipboard: https://developers.google.com/style/numbers#commas-and-decimal-points-in-numbers
```suggestion
- The majority of requests (4,800) peaked at 2,048 KB of memory, 
  likely thanks to app caching.
- Most other requests (3,206) used up to around 10 MB of memory.
```

docs/src/languages/php/fpm.md
```diff
@@ -83,21 +120,29 @@ As an example, consider the following output:
...
+- 758 requests used more than 10 MB and up to around 18 MB of memory.
```
CollierCZ, line 128: :lady_beetle: The numbers `3,206` and `758` aren't very clear from the example. I'd probably simplify the example to make it easier to see the example distinctions. Like there's one at 22.5 MB and where does that fit? Getting rid of some might help.

docs/src/languages/php/fpm.md
```diff
@@ -83,21 +120,29 @@ As an example, consider the following output:
...
+- Only 6 requests peaked at 131 MB of memory, 
+  likely due to cache clears.
 
-A conservative approach would suggest an average request memory of 16 MB should be sufficient.
-A more aggressive stance would suggest 10 MB.
-The more aggressive approach would potentially allow for more concurrent requests at the risk of some requests needing to use swap memory, thus slowing them down.
+In such a situation, you might want to remain cautious and set an average request memory of 16 MB.
+A more aggressive approach might be for you to set an average request memory of 10 MB instead.
```
CollierCZ, line 133: :nail_care: 
```suggestion
A more aggressive approach might be for you to set an average request memory of 10 MB.
```

docs/src/languages/php/fpm.md
```diff
@@ -83,21 +120,29 @@ As an example, consider the following output:
...
+Note that, with the more aggressive approach, there is a risk of allowing more concurrent requests 
+and needing to use swap memory.
```
CollierCZ, line 135: :pick: To try to for more directness and fewer words
```suggestion
With the more aggressive approach, you a risk of allowing more concurrent requests and needing to use swap memory. 
```

Although the idea of `swap memory` isn't clear to everyone and may be more confusing than helpful. :shrug:

docs/src/languages/php/fpm.md
```diff
@@ -83,21 +120,29 @@ As an example, consider the following output:
...
+For further help in estimating your optimal average request memory, 
+you can use the [log analyzer tool for Platform.sh](https://github.com/pixelant/platformsh-analytics) 
+from the web agency [Pixelant](https://www.pixelant.net/). 
+This tool offers a better visualization of access logs to determine how much memory requests use on average.
+It also offers additional insights into the operation of your site. 
+These can help further optimize your configuration 
+and provide guidance on when to increase your plan size.
+Note that this tool is maintained by a third party, not by Platform.sh.
 
+For information on how to set an average request memory, see how to [adjust default assumptions](./fpm.md#adjust-default-assumptions).
```
CollierCZ, line 147: :clipboard: 
```suggestion
For information on how to set the average request memory, see how to [adjust the default assumptions](#adjust-default-assumptions).
```

CollierCZ (changes requested):
I think where you went stronger into changing things, the doc is clearer. I do think you could dive in more strongly in other places as the parts that are the least ideal are mostly leftovers from the previous version. So I'd try to adjust things even more.

The issue with workers and workers is different and confusing. But that's going to be true regardless. :shrug:

CollierCZ (changes requested):
This is so much clearer to me now. :heart_eyes: Just a world of difference. I left in some questions about maybe putting back a little info, but it's definitely better to do that than have too much, like before. They're open questions, as I do think it's good as is.

docs/src/languages/php/fpm.md
```diff
@@ -1,103 +1,114 @@
...
- * The memory available for the container, which depends on the size of the container (`S`, `M`, `L`),
- * The memory that an average request is expected to require,
- * The memory that should be reserved for things that aren't specific to a request (memory for `nginx`, the op-code cache, some OS page cache, etc.)
+You need:
 
-The number of workers is calculated as:
+- An up-and-running web app in PHP, complete with [PHP-FPM](https://www.php.net/manual/en/install.fpm.php)
+- The [Platform.sh CLI](../../administration/cli/_index.md)
 
-![FPM](/images/php/phpfpmworkers.png "0.3")
```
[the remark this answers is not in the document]
AnouckColson: I think a formula like this can sometimes be better than words and bring a lot of clarity. But in this case, I removed it because it actually made everything more confusing to me, and I'm still not sure I completely understand what each element in it means. Granted, I'm terrible at maths, but I figured if it made things more obscure for me, then it might very well be the case for other people too.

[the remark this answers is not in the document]
CollierCZ: I looked into it some more and the `max` and `2` bits mean that it sets a minimum value of 2 (so if the other part of the formula comes out to less than 2, the result is 2). That information is missing elsewhere, but is definitely not ideally presented in the image (let alone only being in image form).

The other three parts of the formula are the three bullet points above. So it should be possible to make them clearer. I think seeing that RequestMemory is on the bottom of the equation is helpful to seeing quickly that it's the inverse -- so increasing it decreases the number of workers.

I also discovered that the result is probably rounded up when it's a fraction. Not sure how much that info helps, but it might.

So I'd replace it with just the three items and include the minimum elsewhere in the text. At the very least, including the minimum.

CollierCZ (changes requested):
Just two small things. The formula doesn't have to be an image. And if you can think of a clear way to make the relationships clear otherwise, it doesn't have to be a formula. But it'd be nice to have it there somewhere.

docs/src/languages/php/fpm.md
```diff
@@ -1,103 +1,122 @@
...
+  depending on [app size](../../create-apps/app-reference.md#sizes).
+- The request memory: the amount of memory an average PHP request is expected to require.
+- The reserved memory: the amount of memory you need to reserve for tasks that aren't related to requests. 
 
-## The heuristic
+The calculation is performed as follows: ![FPM](/images/php/PHP-FPM-Workers-Calculation.png "0.2")
 
-The heuristic is based on three input parameters:
+Note that when the resulting number is a decimal, 
+it is rounded up to set the maximum number of workers.
```
CollierCZ, line 21: :pick:  As noted by the linting, but definitely not a blocker
```suggestion
it's rounded up to set the maximum number of workers.
```

docs/src/languages/php/fpm.md
```diff
@@ -1,103 +1,122 @@
...
-![FPM](/images/php/phpfpmworkers.png "0.3")
+You need:
 
+- An up-and-running web app in PHP, complete with [PHP-FPM](https://www.php.net/manual/en/install.fpm.php)
+- The [Platform.sh CLI](../../administration/cli/_index.md)
 
-## Defaults
+Note that the memory settings mentioned on this page are different from the [`memory_limit` PHP setting](./_index.md). 
+The `memory_limit` setting is the maximum amount of memory a single PHP process can use 
+before it is automatically terminated.
```
CollierCZ, line 35: :pick:  As noted by the linting, but definitely not a blocker
```suggestion
before it's automatically terminated.
```

docs/src/languages/php/fpm.md
```diff
@@ -1,103 +1,122 @@
...
-These estimates are used only for determining the number of PHP-FPM workers to start.
+By default, Platform.sh automatically sets a maximum number of PHP-FPM workers for your app. 
+This number is calculated based on three parameters:
+- The container memory: the amount of memory you can allot for PHP processing 
+  depending on [app size](../../create-apps/app-reference.md#sizes).
+- The request memory: the amount of memory an average PHP request is expected to require.
+- The reserved memory: the amount of memory you need to reserve for tasks that aren't related to requests. 
 
-## The heuristic
+The calculation is performed as follows: ![FPM](/images/php/PHP-FPM-Workers-Calculation.png "0.2")
```
CollierCZ, line 18: :hammer_and_wrench: I'm somewhat allergic to the phrasing `is performed` as it usually seems to me to involve extra words. One of the two suggestions might fit?

Also added meaningful alt text.
```suggestion
The number is calculated as follows: ![The sum of container memory minus reserved memory divided by request memory](/images/php/PHP-FPM-Workers-Calculation.png "0.2")
```

CollierCZ:
```suggestion
This is calculated as follows: ![The sum of container memory minus reserved memory divided by request memory](/images/php/PHP-FPM-Workers-Calculation.png "0.2")
```

CollierCZ (changes requested):
Two very minor things, one definitely not a blocker and the other that maybe shouldn't be, but feels somewhat like it is for me :see_no_evil: The alt text probably is a blocker, so at least there's that to save me :slightly_smiling_face:

docs/src/languages/php/fpm.md
AnouckColson: Didn't mean to trigger you! 😁 But I get it!

CollierCZ (approved):
:rocket:

PR #2705, opened 2022-11-22
👷 Use Homebrew for CLI in GitHub workflows

CollierCZ:
## Why

We weren't using the latest version of the CLI.
Also the process for deleting branches from forks still isn't great.

## What's changed

Added a reusable workflow and used that for the CLI.
Tried to improve the delete branch flow.

### DIFF
--- a/.github/actions/set-up-cli/action.yaml
+++ b/.github/actions/set-up-cli/action.yaml
@@ -1,0 +1,13 @@
+name: Set up Platform.sh CLI
+description: Installs the Platform.sh CLI from Homebrew.
+
+runs:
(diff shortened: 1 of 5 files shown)


CollierCZ:
Can see it working here: https://github.com/platformsh/platformsh-docs/actions/runs/3525083555/jobs/5911348693

AnouckColson (commented):
Had to rely on the GitHub docs as I'm new to GitHub actions. Based on the information I found, I didn't notice anything amiss in the files, but would recommend a second review by @trolologuy just to double-check before going live.

PR #2710, opened 2022-11-23
Fix integration command to use token instead of license_key

raupie:
## Why

```
platform integration:add --type newrelic --url https://log-api.newrelic.com/log/v1 --license_key <REDACTED>

                                              
  [RuntimeException]                          
  The "--license_key" option does not exist.  
```

## What's changed

Use `--token` instead of `--license_key`


CollierCZ:
Are you sure it doesn't exist, @raupie? It's listed as one of the options with `platform help integration:add` specifically for New Relic.

raupie:
Actually it needs to be a dash and not an underscore...

PR #2712, opened 2022-11-24
📝 Switch OpenSearch to major version only

CollierCZ:
## Why

OpenSearch images are switching to major versioning. See Context.

## What's changed

Updated version and added a note.


AnouckColson (approved):
🚀  Perfect

PR #2714, opened 2022-11-24
♿️ Improve accessibility of codetabs

CollierCZ:
## Why

Wasn't always easy to tell that codetabs were there, interactive, and containing content.

## What's changed

Increased contrast more in line with product design system


AnouckColson (approved):
👍 Definitely looks more visible to me.

PR #2715, opened 2022-11-28
📝 Simplify examples for copying and pasting

CollierCZ:
## Why

Too many variables were making examples hard to parse and copy and paste.

## What's changed

Removed the ones that were clear without it and clarified others that could be done quickly.


AnouckColson (approved):
All good in my opinion 👍

PR #2716, opened 2022-11-28
📝 Add guidance on quoting commands

CollierCZ:
## Why

Commands don't work ideally across shells, which was causing some issues on zsh shells.

## What's changed

Added guidance to prefer quotes in commands


lolautruche (approved):

AnouckColson (approved):
🚀

PR #2717, opened 2022-11-28
📝 Clarify where workers are open

CollierCZ:
## Why
It was somewhat unclear how workers were open or not. See Context.

## What's changed

Tried to clarify

### DIFF
--- a/docs/src/create-apps/workers.md
+++ b/docs/src/create-apps/workers.md
@@ -3,7 +3,7 @@ title: "Work with workers"
 description: Interact with your worker instances to handle background tasks for your apps.
 ---
 
-Workers are instances of your code that can't interact with the outside world.
+Workers are instances of your code that aren't open to connections with other apps or services or the outside world.
 They're good for handling background tasks.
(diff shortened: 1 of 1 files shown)


docs/src/create-apps/workers.md
```diff
@@ -3,7 +3,7 @@ title: "Work with workers"
 description: Interact with your worker instances to handle background tasks for your apps.
 ---
 
-Workers are instances of your code that can't interact with the outside world.
+Workers are instances of your code that aren't open to connections with other apps or services or the outside world.
```
Jamesking56, line 6: You can still connect outbound from worker -> external app, I think you mean that they cannot connect inbound external app -> worker

```suggestion
Workers are instances of your code that aren't open to connections from other apps or services or the outside world.
```

CollierCZ: I think that's what I had in my head, but my fingers didn't seem to agree.

AnouckColson (approved):
LGTM!

PR #2719, opened 2022-11-29
🐛 Process shortcodes in note titles

CollierCZ:
## Why

Shortcodes in note titles weren't being processed, example: https://docs.platform.sh/languages/php/xdebug.html#open-a-tunnel

## What's changed

Changed so they would be


AnouckColson (approved):
👍 🚀

PR #2720, opened 2022-11-29
📝 Deprecate PHP 7.4

CollierCZ:
## Why

PHP 7.4 reached the end of its life: https://www.php.net/supported-versions.php

## What's changed

Marked as deprecated


AnouckColson (approved):
🚀

PR #2721, opened 2022-11-29
📝 Clarify automated manual backups

CollierCZ:
## Why

It wasn't clear whether automating manual backups counted as automated or not. See Context.

## What's changed

Tried to clarify.

### DIFF
--- a/docs/src/environments/backup.md
+++ b/docs/src/environments/backup.md
@@ -156,6 +156,16 @@ highlight=false
 
 {{< /codetabs >}}
 
+### Automate manual backups
+
+You can also automate the process of creating manual backups through [cron jobs](../create-apps/app-reference.md#crons).
+The cron job should use the CLI command to backup the environment.
+It requires you to [set up the CLI on the environment with an API token](../administration/cli/api-tokens.md#on-a-platformsh-environment).
+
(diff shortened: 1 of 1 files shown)


docs/src/environments/backup.md
```diff
@@ -156,6 +156,16 @@ highlight=false
 
 {{< /codetabs >}}
 
+### Automate manual backups
+
+You can also automate the process of creating manual backups through [cron jobs](../create-apps/app-reference.md#crons).
+The cron job should use the CLI command to backup the environment.
```
AnouckColson, line 162: 🛠️ To avoid the use of `should` and fix the typo?

```suggestion
In this case, the cron job uses the CLI command to back up the environment.
```

AnouckColson (changes requested):
A small suggestion, otherwise perfect!

docs/src/environments/backup.md
CollierCZ: I don't think `in this case` works ideally here, because it's not clarifying cron jobs in general to this specific example. How about without that start:
```suggestion
The cron job uses the CLI command to back up the environment.
```

AnouckColson (approved):
👍 🚀

PR #2722, opened 2022-11-29
rename gitlab to github in integration:add

boulaffasae:
## Why

Hi ! I think you added the wrong type in the following command

```
platform integration:add --type=gitlab
```

## What's changed

I think what was changed is clear


CollierCZ (approved):
Thanks for the fix, @boulaffasae!

PR #2727, opened 2022-12-02
📝 Clarify use of tunnel:single

### LINKED ISSUE #2701: 📝 Clarify use of tunnel:single

https://docs.platform.sh/development/ssh.html#connect-to-services
Maybe elsewhere, like https://docs.platform.sh/add-services.html#2-connect-to-an-ssh-t
Consider on individual services page if they use `tunnel:open` for a single service.

Make it clear you can connect to a single service with `platform tunnel:single` so you don't have to open all services to get the one that you want. Consider including options like the port, though don't get too complicated.

### Additional context

_No response_

CollierCZ:
## Why

Closes #2701

## What's changed

Clarified instructions on service pages to connect only to the one tunnel.
Made generic service instructions clearer, splitting app and tunnel instructions.

### DIFF
--- a/docs/src/add-services/_index.md
+++ b/docs/src/add-services/_index.md
@@ -140,41 +140,64 @@ For some services, you can change the timezone for the running service
 
 ## Connect to a service
 
-Once a service is running and exposed as a relationship,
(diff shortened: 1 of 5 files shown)


AnouckColson (approved):
Approved! 👍

PR #2737, opened 2022-12-06
👷 Use CLI directly for geting environment info

CollierCZ:
## Why

Routing through the API caused problems for environments with slashes in the name (see #2733).

## What's changed

Used CLI commands instead of curls through the API to get environment info.


AnouckColson (approved):
👍

PR #2738, opened 2022-12-07
🐛 Fix remaining codetabs syntax

CollierCZ:
## Why

The change in #2732 missed some codetabs, example: https://docs.platform.sh/languages/php.html#changing-the-flavor

## What's changed

Changed some missing ones to the right syntax


trolologuy (approved):
✅ 

PR #2739, opened 2022-12-07
💚 Fix environment URL in GitHub comments

CollierCZ:
## Why

With the switch to the Platform command, the workaround for the environment URL was causing duplication and errors

## What's changed

Fixed the errro


AnouckColson (approved):
👍 🚀

PR #2742, opened 2022-12-08
💚 Fix linting of line length with tables

CollierCZ:
## Why

Was improperly showing linting warnings on table length

## What's changed

Added the right parser


trolologuy (approved):
✅  Tested for both CLI and plugin. Works as expected.
Might need to re-add folders to the workspace + restart VSC for the plugin to work.

PR #2743, opened 2022-12-08
📝 Clarify MySQL permissions

CollierCZ:
## Why

Some permissions were missing or unclear in the MySQL docs. See Context.

## What's changed

Added a section specifically on permissions
Updated the list to be more complete
Rounded out the list of included permissions

### DIFF
--- a/docs/src/add-services/mysql/_index.md
+++ b/docs/src/add-services/mysql/_index.md
@@ -202,24 +202,49 @@ If your database relationship has a password, pass the `-p` switch and enter the
 mysql -p -h mysql.internal -P 3306 -u user main
 ```
 
-## Multiple databases
+## Define permissions
 
-With version `10.0` or later, you can define multiple databases and multiple users with different permissions.
+With version `10.0` or later, you can define multiple users with different permissions for your database.
 To do so, define multiple endpoints in your [service configuration](#configuration-options).
 
 For each endpoint you add, you can define the following properties:
 
 | Name             | Type                     | Required | Description |
 | ---------------- | ------------------------ | -------- | ----------- |
-| `default_schema` | `string`                 |          | Which of the schemas defined above to default to. If not specified, the `path` property of the relationship is `null` and so tools such as the Platform CLI can't access the relationship. |
+| `default_schema` | `string`                 |          | Which of the defined schemas to default to. If not specified, the `path` property of the relationship is `null` and so tools such as the Platform CLI can't access the relationship. |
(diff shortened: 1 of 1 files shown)


docs/src/add-services/mysql/_index.md
```diff
@@ -201,24 +201,49 @@ If your database relationship has a password, pass the `-p` switch and enter the
...
 | `privileges`     | A permissions dictionary |          | For each of the defined schemas, what permissions the given endpoint has. |
 
 Possible permissions:
 
-* `ro`: Only SELECT queries are allowed.
-* `rw`: SELECT queries and INSERT/UPDATE/DELETE queries are allowed.
-* `admin`: All queries are allowed including data definition language (DDL) queries (such as `CREATE TABLE`, `DROP TABLE`).
+| Name        | Type          | Description                                         |
+| ----------- | ------------- | --------------------------------------------------- |
+| Read-only   | `ro`          | Can select, create temporary tables, and see views. |
```
trolologuy, line 220: 😍  A table is indeed much clearer here!

docs/src/add-services/mysql/_index.md
```diff
@@ -201,24 +201,49 @@ If your database relationship has a password, pass the `-p` switch and enter the
...
+```
+
+If either `schemas` or `endpoints` are defined, no default is applied and you have to specify the full configuration.
+
+## Multiple databases
+
+With version `10.0` or later, you can define multiple databases.
+To do so, define multiple `schemas` in your [service configuration](#configuration-options).
 
+You can also specify multiple `endpoints` for [permissions](#define-permissions).
```
trolologuy, line 246: :bulb: :arrow_forward:
Would it make sense to reintroduce the concept of "users" in that context, since this is what endpoints are.
Or is the link back to permissions clear enough?

:hammer_and_wrench:
Maybe that could be enough?
```suggestion
You can also specify multiple user `endpoints` for [permissions](#define-permissions).
```

trolologuy (approved):
✅  Looks good! Small proposition, other than that I guess it's safe to accept, even if `replication` might or might not miss some rights (as you pointed out in the additional context).

- [x] Any relevant issues raised by the linting have been addressed.

docs/src/add-services/mysql/_index.md
CollierCZ: Here, I'd say it doesn't add so much value as concept doesn't help me see what needs to be defined here. I feel like endpoints are actually endpoints and it's not necessarily important if they're users or not. Sometimes can be, but for that the link would be enough.

PR #2757, opened 2022-12-16
👷 Switch tools for testing internal links

CollierCZ:
## Why

Our tool wasn't checking if headers for internal links existed

## What's changed

Tried creating a Bash script, but it got too complicated to maintain for long.
Switched to [htmltest](https://github.com/wjdp/htmltest).
This means we don't have a Git hook, but basic pages existing or not can be tested by building with `hugo`.


CollierCZ:
Example job where it failed on a missing header: https://github.com/platformsh/platformsh-docs/actions/runs/3715766686/jobs/6301316068

AnouckColson (approved):
💯

PR #2746, opened 2022-12-09
:memo: Rework CDN docs

AnouckColson:
## Why

Closes #2670 

## What's changed

Rework CDN pages

PR #2762, opened 2022-12-20
:memo: Rework apex domains

### LINKED ISSUE #2672: 📝 Rework apex domains info

https://docs.platform.sh/domains/steps/dns.html

Left over from #2447

Update for:

- Content accuracy
- Includes what it should
- Fits template
- Fits style guide

### Additional context

_No response_

AnouckColson:
## Why

Closes #2672 

## What's changed

Rework apex domains info


docs/src/domains/steps/dns.md
```diff
@@ -1,170 +1,139 @@
...
+- During an upgrade or maintenance operation, routers can be taken offline while changes are applied.
+- During a region upscale or downscale, routers can be added or removed.
+
+To prevent downtime in such situations,
+create `CNAME` records through your DNS provider.
+ 
+`CNAME` records map a domain to another.
+Set up your `CNAME` records to point at your project's [target edge hostname](../../domains/steps/_index.md#2-get-the-target-for-your-project).
+Each Platform.sh edge hostname comes with a DNS record
+that resolves the IP addresses of the routers in your project's region.
```
AnouckColson, line 28: Don't really like this sentence, but couldn't find a way to make it clearer so definitely open to suggestions.

docs/src/domains/steps/dns.md
```diff
@@ -1,170 +1,139 @@
...
-
-<!-- vale Platform.condescending = NO -->
-* `CNAME` flattening at [CloudFlare](https://developers.cloudflare.com/dns/additional-options/cname-flattening)
-* `ANAME` records at [easyDNS](https://easydns.com/features/aname-root-domain-alias/),
-  [DNS Made Easy](https://support.dnsmadeeasy.com/support/solutions/articles/47001001412-aname-records),
-  and [Name.com](https://www.name.com/support/articles/115010493967-adding-an-aname-alias-record)
-* `ALIAS` records at [DNSimple](https://support.dnsimple.com/articles/alias-record/)
-  and [ClouDNS](https://www.cloudns.net/wiki/article/18/)
-<!-- vale Platform.condescending = YES -->
-
```
AnouckColson, line 89: I went back and forth on whether I should keep all the links or not. Open to opinions.

AnouckColson:
Not in love with this first draft but a pair of fresh eyes would help.

docs/src/domains/steps/dns.md
```diff
@@ -1,170 +1,139 @@
...
+ 
+1.  Make the `www.` version of your site the default (canonical) version
+    and configure your app and routes to [use the `www` subdomain as upstream](../../define-routes/_index.md).
+ 
+2.  Follow the instructions on [how to set up a custom domain](./_index.md).
+    When you come to configuring your DNS provider, instead of a `CNAME` record,
+    add an `A` record pointing from your domain to the redirection service.
+ 
+3.  Ensure that your redirects use the same protocol:
    `http://example.com` redirects to `http://www.example.com`, not to `https://www.example.com`.
```
CollierCZ, line 105: :question: Is it necessary to have `http` redirects for the redirection service? Because generally, `http` is automatically redirected to `https` and so `http` doesn't need to be included in the routes definition. If it does for redirects, that should be a separate step. If it's not necessary, I wouldn't use the example of `http`, since many people won't be defining that.

docs/src/domains/steps/dns.md
```diff
@@ -1,170 +1,139 @@
...
+consider using `A` records, but only as a last resort.
+ 
+When you use `A` records, [if a router's IP address changes](#cname-records),
+you need to update your `A` records manually.
+Until you do, your site can appear offline because requests are lost.
+ 
+To use `A` records, follow these steps:
+ 
+1.  To get the IP addresses of your project's production environment, run the following command:
+    ```bash
```
CollierCZ, line 128: :clipboard: 
````suggestion

        ```bash
````

**Review by CollierCZ — CHANGES_REQUESTED** [review]
I think you did well at making it shorter and getting it to follow our standards. I had a few questions, including the one about the use of codetabs that we should probably discuss more?

File: docs/src/domains/steps/dns.md (hunk 4 above)
Comment (AnouckColson) [reply docs/src/domains/steps/dns.md:104]: I applied this suggestion but ESLint then showed an error in my file (could be that my settings are off again? 🤔 ). How many spaces are we truly expecting here?

File: docs/src/domains/steps/dns.md [hunk 5]
```diff
@@ -1,170 +1,111 @@
...
+ 
+## Workarounds for apex domains
+
+Although `CNAME` records can't point to apex domains,
+most DNS providers offer workarounds.
+Contact your DNS provider and choose the option that works for you.
+
+Note that workarounds are also available for use with a CDN.
+For example, [Fastly](../cdn/fastly.md) offers [Anycast options](https://docs.fastly.com/en/guides/using-fastly-with-apex-domains) 
+and [Cloudflare](../cdn/cloudflare.md) offers [`CNAME` flattening options](https://developers.cloudflare.com/dns/additional-options/cname-flattening/).
```
CollierCZ, line 48: :hammer_and_wrench: :arrow_forward: With #2746, I'd be in favor of keeping the exact explanation of what the workaround is in one place. I know it's unlikely they'll change the name or anything, so maybe I shouldn't worry so much about repeating content. But I'd probably go with something like this:

```suggestion
If you use a CDN, you likely already have a workaround in place,
such as with [Fastly](../cdn/fastly.md#3-handle-apex-domains)
and [Cloudflare](../cdn/cloudflare.md#3-handle-apex-domains).
```

Not a blocking idea though.

CollierCZ (approved):
Two small suggestions and one formatting note, but none of them are blocking if you want to publish.

CollierCZ (approved):

PR #2764, opened 2022-12-21
📝 Hard deprecate Python 2.7

CollierCZ:
## Why

Python 2.7 will no longer receive any updates at all. See Context.

## What's changed

Added a note to that effect.


AnouckColson (approved):
👍

PR #2768, opened 2023-01-02
💄 Simplify theme

CollierCZ:
## Why

The theme implementation was complicated and had outdated dependencies. And lots of duplication.
Want to make it easier to redo the styling with upcoming design changes.

## What's changed

Tried to streamline things.
Now using [TailwindCSS](https://tailwindcss.com/) and Alpine.js exclusively (outside the search, which still uses React for now).
Moved visual appearance things into the theme.
Not all visuals will be the same. Final decisions on appearance will come with the rework of the design. This just makes that work easier.
But if anything isn't working at all, it would be good to fix.


AnouckColson (changes requested):
I first looked at the code and could only find one typo (see suggestion).
Then I browsed the site locally with the new theme to see if it applied as expected and found 2 minor things to adjust:

-  When added to a bulleted list, small icons (**Settings** icon for example) aren't in line with the text (an example can be found at [Administration > Console > Configure environments > Actions on environments](https://docs.platform.sh/administration/web/configure-environment.html#actions-on-environments).
- The new theme makes it seem like there is an issue with the formatting of the headers in [this table](https://docs.platform.sh/create-apps/app-reference.html#types) (compared to old theme where you could tell the formatting was intentional). Not a big deal but worth mentioning I think.

Apart from that, I didn't see anything broken or wonky. That seems like a lot of work, well done!  🦾

CollierCZ:
Thanks for finding those issues! I'd probably say that there was and still is an issue with code in the header of tables, like you pointed out. A final design will come later. I inverted the background to make it stand out more, if you think that's enough for now.

AnouckColson (approved):
Looks good to me now 👍

PR #2763, opened 2022-12-21
🚚 Deprecate Hugo aliases

CollierCZ:
## Why

Hugo aliases use meta tags with `refresh` set to true.
For SEO, it's been suggested we should return 301 codes instead.

## What's changed

Changes aliases into redirects and added guidance on why.

PR #2773, opened 2023-01-03
🐛 Fix redirect to right page

CollierCZ:
## Why

One redirect from #2763 was to the wrong place. And there was still guidance on how to use Hugo aliases.

## What's changed

Fixed the redirect and removed the guidance.


AnouckColson (approved):
👍

PR #2775, opened 2023-01-04
🐛 Fix syntax of route example

CollierCZ:
## Why

The [default route definition example](https://docs.platform.sh/define-routes.html#default-route-definition) was broken

## What's changed

Tried to fix it.


AnouckColson (approved):
👍 🚀

PR #2796, opened 2023-01-12
docs: fix backquote in mysql / go sample code

daeMOn63:
## Why

The backtick in MySQL / go sample code got hex encoded somehow, this restores them so the code can be copy pasted.

## What's changed

`\x60` occurrences got replaced by ``` ` ```

### DIFF
--- a/docs/static/files/fetch/examples/golang/mysql
+++ b/docs/static/files/fetch/examples/golang/mysql
@@ -3,6 +3,7 @@ package examples
 import (
 	"database/sql"
 	"fmt"
+
 	_ "github.com/go-sql-driver/mysql"
 	psh "github.com/platformsh/config-reader-go/v2"
 	sqldsn "github.com/platformsh/config-reader-go/v2/sqldsn"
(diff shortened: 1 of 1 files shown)


CollierCZ:
Thanks so much for noticing and submitting a fix, @daeMOn63! Unfortunately, it's a little more complicated. The actual example comes from this site: https://examples.docs.platform.sh/golang/
That site is built from (among other things): this file: https://github.com/platformsh/language-examples/blob/master/golang/examples/mysql.go

The file you changed is our backup in case the example site is down. But the file from the example site has the error and will always overwrite our backup. So we need to fix the problem at the source. :slightly_smiling_face:

daeMOn63:
Thanks! Indeed it's a bit more tricky to fix it there, golang doesn't like to nest backtick in multiline strings. Could probably work around it rewriting the multiline blocks with something like:
```
s = "line1\n" +
    "line2\n" +
    "line3"
```
in place of the current backticks, but that may hurt readability. Lmk if this is something you'd be fine with, I can PR it.

CollierCZ:
The readability might be slightly lower, but it's definitely better than `\x60` :grinning: I would be fine with that way. Thanks!

CollierCZ:
With https://github.com/platformsh/language-examples/pull/83, should be updated soon in the docs automatically. Thanks again!

PR #2798, opened 2023-01-12
📝 Document flexible resources

CollierCZ:
## Why

We'd like to document the ability to adjust memory based on CPU. But it shouldn't be part of the main documentation as it will disappear with more fully fledged flexible plans.

## What's changed

Added a description in an accessible page not in the main navigation.

### DIFF
--- a/docs/src/create-apps/flexible-resources.md
+++ b/docs/src/create-apps/flexible-resources.md
@@ -1,0 +1,61 @@
+---
+title: Flexible resources
+sidebarIgnore: true
+---
+
+Sometimes you have apps that are more memory-intensive.
+In such cases, you can adjust the amount of memory using the `resources` key in your [app configuration](./_index.md).
+All memory must stay within the limits set by your plan.
+
+The resources key has the following possible options:
+
+| Name           | Type      | Default | Minimum | Description                                                               |
+| -------------- | --------- | ------- | ------- | ------------------------------------------------------------------------- |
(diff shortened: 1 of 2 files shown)


docs/src/create-apps/flexible-resources.md
```diff
@@ -0,0 +1,61 @@
+---
+title: Flexible resources
+sidebarIgnore: true
+---
+
+Sometimes you have apps that are more memory-intensive.
+In such cases, you can adjust the amount of memory using the `resources` key in your [app configuration](./_index.md).
+All memory must stay within the limits set by your plan.
+
+The resources key has the following possible options:
```
AnouckColson, line 10: 📋 

```suggestion
The `resources` key has the following possible options:
```

docs/src/create-apps/flexible-resources.md
```diff
@@ -0,0 +1,61 @@
...
+Setting the `resources` key on one app doesn't affect other apps in your project.
+So you can set it for one app and have your other apps be assigned resources automatically.
+
+Workers based on apps with the `resources` key default to inheriting the set resources.
+To free memory for your main app,
+set the values within the `resources` key to be lower for your workers.
+
+## Performance profiles
+
+The two options for `resources` allows for different performance profiles.
```
AnouckColson, line 53: 📋 

```suggestion
The two options for `resources` allow for different performance profiles.
```

AnouckColson (changes requested):
2 minor changes to make, otherwise 👍

AnouckColson (approved):

PR #2811, opened 2023-01-19
✨ Add shortcode for links in code blocks

CollierCZ:
## Why

Wanted to see if it was possible to add links inside code blocks. This can potentially help with complicated YAML files.

## What's changed

Added a short code and some examples in app configuration. Currently have a `↗` after them to distinguish them as links. Can consult on this with Design during the redesign.

Also moved shortcodes to theme that should have been there before.


trolologuy (dismissed):
✅  Really cool addition! Didn't expect such a quick implementation! It really helps to provide additional context (in addition to the inline comments), which can be really helpful for users.

🤔 Only downsides currently could be:
1. how to check these links?
2. length of the inline shortcode. Is it a good idea to have the link inline? Could that parameter be used for overrides instead (and have common links for the most used keys?)?
3. how do we maintain yaml blocks globally? Shoud we use a dedicated folder (for simpler mass updates) with all examples and include them with shortcodes?

CollierCZ:
1. The links are checked on the generated HTML. That's one of the reasons we switched checkers.
2. I'm not sure what you mean by `inline` here or what the alternative is. It would be possible to set it up to have a list of links to choose from, but I'm not sure if that's not more work upfront that would be helpful. Not sure how often it'll be used.
3. I'm not sure I see the benefit from the folder. Could you explain what your thinking is there? A lot of examples are done through shortcodes already, through the registry parser in the `src/registry/images/examples` directory after running `npm dev`.

trolologuy:
1: Ok, I wasn't sure if that would be the case. ✅ 
2. Yeah the idea would have been to have shorter arguments (based on a list for e.g, yes), to have defaults that'd work for most cases.
3. The idea was just to have everything's that's not generated automatically in one folder, so that we can buld update all code blocks at once if there are changes on how we display them, add links, ...

PR #2816, opened 2023-01-20
🦺 Validate form submissions

### LINKED ISSUE #2815: [Improvement] Feedback form processor

/feedback/

Processing of feedback form data

### Additional context

To further improve the security of the feedback submission processor, we should validate the data we are given before attempting to store the data. 

My javascript is a bit rusty so take the code snippets as pseudo code. 

From [Line 35 of platformsh-docs/docs/layouts/partials/feedback/form.html](https://github.com/platformsh/platformsh-docs/blob/main/docs/layouts/partials/feedback/form.html#L35) we know that the currentURL should start with the domain of the site that we're on. Since we're already [including the config reader](https://github.com/platformsh/platformsh-docs/blob/main/docs/index.js#L4), we should be able to retrieve the primary route using `config.getPrimaryRoute().url`

Before we run the insert at line [38 in platformsh-docs/docs/index.js](https://github.com/platformsh/platformsh-docs/blob/main/docs/index.js#L38) we should be able to check the value of `feedback.url` to make sure it starts with a value we expect.
```javascript
if (!feedback.url.startsWith(config.getPrimaryRoute().url)) {
  //throw error and exit
}
```

From [lines 51-52 of form.html](https://github.com/platformsh/platformsh-docs/blob/main/docs/layouts/partials/feedback/form.html#L51-L52), we know that values sent for `feedback` should be either the value `positive` or `negative`. With that knowledge, we should be able to make sure the value we have for `feedback.feedback` is what we expect before inserting.

```javascript
if (['positive','negative'].includes(feedback.feedback)) {
    console.log('you can proceed')
} else {
    console.log('you shall not pass!')
} 
```

Last up is date. We don't really need to have the date sent to the form processor as we can just instruct MySQL to insert the current date using `sysdate()` or `now()` instead of `feedback.date` when inserting the record. Or, you could skip having to insert it all by altering the table schema to have the default for the `date` field be `DEFAULT CURRENT_TIMESTAMP` so that MySQL will automatically insert the current timestamp when it creates the new record.

CollierCZ:
## Why

Closes #2815

## What's changed

Added validation of submissions and switched date to submission itself.

Also made the response be sent only once (Returning in response). And increased limit for URL for long development URLs like the one for this branch.

### DIFF
--- a/docs/index.js
+++ b/docs/index.js
@@ -35,14 +35,29 @@ app.post("/feedback/submit", async (req, res) => {
   // Create a feedback table if it doesn't exist
   try {
     await connection.query(
-      `CREATE TABLE IF NOT EXISTS Feedback (
-          id INT(6) UNSIGNED AUTO_INCREMENT PRIMARY KEY,
-          date DATETIME NOT NULL,
-          url VARCHAR(300) NOT NULL,
-          feedback VARCHAR(10) NOT NULL
-        )`
+      `SELECT count(*) FROM information_schema.TABLES
+      WHERE (TABLE_SCHEMA = '${credentials.path}')
+      AND (TABLE_NAME = 'Feedback')`,
+      async (error, result) => {
+        if (error) {
+          return res.status(500).send("Error looking for feedback table")
+        }
+        if (result.length === 0) {
+          try {
+            await connection.query(
+              `CREATE TABLE Feedback (
+                id INT(6) UNSIGNED AUTO_INCREMENT PRIMARY KEY,
+                date DATETIME NOT NULL,
+                url VARCHAR(300) NOT NULL,
+                feedback VARCHAR(10) NOT NULL
+              )`
+            )
+          }
+          catch (err) { return res.status(500).send("Error creating a table for feedback") }
+        }
+      }
     );
-  } catch (err) { return res.status(500).send("Error creating a table for feedback") }
+  } catch (err) { return res.status(500).send("Error connecting to database") }
 
   // Insert feedback record
   try {
--- a/docs/index.js
+++ b/docs/index.js
@@ -27,6 +27,9 @@ app.post("/feedback/submit", async (req, res) => {
     return res.status(400).send("The submitted URL isn't valid")
   }
 
+  // Encode URL
+  const urlForFeedback = encodeURI(feedback.url)
+
   // Validate feedback itself
   if ((typeof feedback.feedback !== 'string') || !['positive', 'negative'].includes(feedback.feedback)) {
     return res.status(400).send("The submitted feedback isn't valid")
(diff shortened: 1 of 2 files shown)


CollierCZ:
Can test with commands like this:
```
curl -X POST  -H "Content-Type: application/json" -d "{ \"feedback\": \"negative\", \"url\": \"🗡️ SQL ATTACK! 🗡️\" }" https://validate-feedback-input-4y4hooy-652soceglkw4u.eu-3.platformsh.site/feedback/submit
```

docs/index.js
```diff
@@ -20,24 +20,35 @@ app.post("/feedback/submit", async (req, res) => {
...
+
+  // Validate feedback itself
+  if ((typeof feedback.feedback !== 'string') || !['positive', 'negative'].includes(feedback.feedback)) {
+    return res.status(400).send("The submitted feedback isn't valid")
+  }
 
   // Create a feedback table if it doesn't exist
   try {
     await connection.query(
       `CREATE TABLE IF NOT EXISTS Feedback (
```
gilzow, line 38: Unrelated to the security issues, I would suspect a SELECT here would be much more performant than a `create if exists`. 

(this assumes the schema/database is named `main`. Change as needed)
e.g. 
`select count(TABLE_NAME) from information_schema.tables where table_schema = 'main' and table_name = 'feedback';` 

and then check the result to make sure it is 1.

docs/index.js
```diff
@@ -20,24 +20,35 @@ app.post("/feedback/submit", async (req, res) => {
   });
 
   const feedback = req.body
+  const today = new Date().toISOString().slice(0, 19)
+
+  // Validate URL in feedback
+  if ((typeof feedback.url !== 'string') || !feedback.url.startsWith(config.getPrimaryRoute().url)) {
```
gilzow, line 26: One thing to consider is that this doesn't account for other types of injections 

```
curl -X POST  -H "Content-Type: application/json" \                                                                                        127 х  system Node  01:25:02 PM
-d "{ \"date\":\"2023-01-23T19:22:04\", \"feedback\": \"positive\", \
\"url\": \"https://docs.platform.sh/fooobar</td> <script>alert(1)</script><td> \"}" \
https://docs.platform.sh/feedback/submit
```
BUT it looks like you have something already in place to encode the data or mitigate this type of risk on the output side?

gilzow (approved):
Beyond the one issue i mentioned (as a _consideration_) this looks good.

docs/index.js
CollierCZ: The data is only displayed as an escaped string, so it should be safe, but it doesn't hurt to eliminated some values that would never appear in a real URL.

docs/index.js
```diff
@@ -20,26 +20,37 @@ app.post("/feedback/submit", async (req, res) => {
...
+          url VARCHAR(100) NOT NULL,
+          feedback VARCHAR(10) NOT NULL
+        )`
+      );
+    } catch (err) { res.status(500).send("Error creating a table for feedback") }
+
+    // Insert feedback record
+    try {
+      await connection.query(
+        'INSERT INTO Feedback (date, url, feedback) VALUES (?,?,?)', [today, feedback.url, feedback.feedback]);
```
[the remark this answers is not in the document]
0xDSousa: Seems like I forgot to hit the submit review button 😮‍💨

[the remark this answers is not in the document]
CollierCZ: Hope [this works](https://github.com/platformsh/platformsh-docs/pull/2816/files#diff-c562fb846f34de0bd255222a1e312f9f07dee217c353136f78377e1f9d324b37R30-R31)

0xDSousa (approved):
I find this code to be agreeable. Let's get it done.

PR #2817, opened 2023-01-20
📝 Remove OpenSearch from DG3

CollierCZ:
## Why

See Context.

## What's changed

As the title says


trolologuy (approved):
✅ 

Issue #2828, opened 2023-01-26
Library not loaded: openssl@1.1


cobadger-aten:
### Where on docs.platform.sh should be changed?

https://docs.platform.sh/administration/cli.html

### What exactly should be updated?

Provide steps to diagnose and correct the problem in which platformsh-cli returns the error
`dyld[4563]: Library not loaded: /opt/homebrew/opt/openssl@1.1/lib/libssl.1.1.dylib`.

### Additional context

Randy Fay helped me on Platforsh Slack this afternoon diagnosing and correcting this problem on my Mac. He thought it might be helpful to document this in case other users experience the same problem.

Some months ago I got a new Macbook Pro with an M1 chip and I migrated everything over from my non-M1 machine using Apple's migration tool. If I remember correctly, something that I had installed in my non-M1 machine wasn't working at the command line on my new machine, and I tried using Rosetta as a workaround.

It was a successful workaround for months.

For no reason that I could discern, platformsh-cli abruptly stopped working for me: any `platform` command would report the error shown above. No amount of repairing, uninstalling and installing platformsh-cli and openssl@1.1 solved the problem.

rfay:
Can you please update the OP more explaining your experience and how you got there, and what you had to do to fix it?

cobadger-aten:
@rfay Updated.

PR #2829, opened 2023-01-26
Update docs to correct misconfigured Homebrew, per Randy Fay

cobadger-aten:
## Why

Documents process of correcting mis-configured Homebrew on some Macs with M1 chips. (Issue [#2828)](https://github.com/platformsh/platformsh-docs/issues/2828)

### DIFF
--- a/docs/layouts/shortcodes/cli-installation.html
+++ b/docs/layouts/shortcodes/cli-installation.html
@@ -20,3 +20,17 @@ $ scoop install platform
 {{ partial "codetabs/tabs" ( dict "Inner" $Inner "Page" .Page ) }}
 
 <p>For manual installation, you can also <a href="https://github.com/platformsh/cli/releases/latest">download the latest binaries</a>.</p>
+<p>A word about Macbooks with the M1 chip: It may be that you imported your brew
+  formulae from a pre-M1 Mac, in which case your installation could be in the
+  wrong place.</p>
+<pre>
+1. Make a list of your installed formulae: `brew list --formulae >~/Downloads/formulae.txt`
+2. Unlink all of those formulae: `brew unlink $(brew list --formulae)`
+3. Clear out your current Homebrew: `sudo rm -r /usr/local/Homebrew`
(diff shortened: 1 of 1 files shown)


rfay (commented):
You're certainly not alone in needing this information, but it may be that it needs to go somewhere else. I sure wish homebrew wouldn't even allow this to happen.

cobadger-aten:
This is the page I found when searching for solutions to the problem I was having. Might be good to at least have a link to this documentation from this page if the documentation ends up somewhere else.

rfay:
We could probably put up a simple "fixyourhomebrew.org" that had instructions and a script. I wonder if something like this exists. I still wonder why homebrew doesn't solve this particular problem by balking. There are other problems (like the thing in the path isn't homebrew at all) that are maybe harder to deal with.

CollierCZ:
Given that this is for a specific architecture, I'd prefer not to have it in the middle of the main documentation. For now, I have a pending article on the [community site](https://community.platform.sh). When that's published, I'll add the link so people who it's relevant to can see it. Does that work for you?

rfay:
IMO it's better to cover the problem in detail in another place then link to it, so this is a good strategy.

rfay (approved):
This approach looks good and the article looks good.

Should the article or another one also cover:
* Troubleshooting "wrong platform tool in $PATH" which often happens when people first use the go cli?
* Troubleshooting $PATH in mac M1 and Linux homebrew, where people have not understood that they need to add /opt/homebrew/bin or /home/linuxbrew/.linuxbrew/bin to their path? 

Not saying it needs to go in this article, but we'll continue to have these just because people don't have depth of understanding of PATH or homebrew.

cobadger-aten:
> Given that this is for a specific architecture, I'd prefer not to have it in the middle of the main documentation. For now, I have a pending article on the [community site](https://community.platform.sh). When that's published, I'll add the link so people who it's relevant to can see it. Does that work for you?

@CollierCZ sure! thanks.

AnouckColson (approved):

CollierCZ:
I added something about the path from the DDEV PR.

@cobadger-aten , thanks for pointing out this issue and writing instructions on how to fix it!

cobadger-aten:
@CollierCZ You're welcome! It was really helpful for @rfay to get my Macbook sorted out.

PR #2850, opened 2023-02-02
🔥 Remove outdated rules

CollierCZ:
## Why

Some protection rules were rather out of date. See Context.

## What's changed

Removed them.


0xDSousa:
The information is not updated and most the links we were using are now dead/moved. Customer's aren't really concerned about a 2018 vulnerability, they expect it to be patched.

AnouckColson (approved):

PR #2865, opened 2023-02-07
🐛 Fix example for custom PHP extension

CollierCZ:
## Why

The YAML syntax for a custom PHP extension was wrong. It had a new line like a dictionary but syntax like a string. 

## What's changed

changed it to a dictionary.


AnouckColson (approved):

apotek (commented):
It's a good fix. I'm surprised the `yaml` highlighter didn't choke on that.

PR #2872, opened 2023-02-08
📝 Increase findability of setting remote

CollierCZ:
## Why

Some people had trouble seeing how to set a remote Git branch. See Context.

## What's changed

Moved the CLI version to first and added a keyword to increase its rank in search results.


AnouckColson (approved):
Looks good to me!

PR #2887, opened 2023-02-10
:bug: Clarify PHP example values

trolologuy:
## Why

The example table was unclear. Added more reasonable examples.
See related thread.

## What's changed

Edited values.


CollierCZ (changes requested):
Looks good, just would prefer to keep curly quotes out of the code itself.

CollierCZ (approved):

PR #2889, opened 2023-02-10
📝 Update for new settings layout

### LINKED ISSUE #2812: 📝 Update for new settings layout

Various places under https://docs.platform.sh/administration.html and maybe elsewhere.

The settings layout and options will soon be updated (see Context). This means some screenshots will be out of date. Some instructions may also have to change. There shouldn't be large changes, just perhaps in a variety of places.

Look through the various changes and determine what needs to change in the docs.

### Additional context

Should be ready by the end of the month

CollierCZ:
## Why

Closes #2812

## What's changed

Updated instructions and screenshots for some new settings layouts.


docs/src/administration/users.md
```diff
@@ -104,9 +104,9 @@ title=In the Console
 - Select the project where you want to add a new user.
 - Click {{< icon settings >}} **Settings**.
 - Click **Access**.
-- Expand the user you want to delete.
-- Click **Delete**.
-- Click **Save**.
+- Click the user you want to delete.
+- Click **Remove user**.
+- Click **Accept**.
```
AnouckColson, line 109: 🛠️ I believe this should be an ordered list.

docs/src/administration/users.md
```diff
@@ -104,9 +104,9 @@ title=In the Console
...
 
 <--->
```
AnouckColson, line 111: 🛠️ For consistency's sake, I would suggest also changing the terminology from _delete_ to _remove_ in the rest of the section as well (**Using the CLI tab** + the 2 sentences at the end).

docs/src/administration/users.md
```diff
@@ -139,9 +139,9 @@ title=In the Console
 - Select the project where you want to add a new user.
 - Click {{< icon settings >}} **Settings**.
 - Click **Access**.
-- Expand the user you want to delete.
-- Click **Delete**.
-- Click **Delete** to confirm.
+- Click the user whose permissions you want to manage.
+- Change the settings.
+- Click **Save**.
```
AnouckColson, line 144: 🛠️ I believe this should be an ordered list as well.

docs/src/development/variables/set-variables.md
```diff
@@ -23,17 +23,42 @@ Application variables are available at both build time and runtime.
 ## Create project variables
 
 Add secrets for all environments in project variables
-using [the Console](../../administration/web/configure-project.md#variables) or the CLI.
+using [the Console](../../administration/web/configure-project.md#variables) or the [CLI](../../administration/cli/_index.md).
 
-For example, to create the project variable `foo` with the value `bar`, run:
+For example, you may need to set a variable to vary between production and other environments, such as API gateway credentials.
+You can set the non-production version as a project variable
+and then set a [variable specific to the production environment](#create-environment-specific-variables) to override it.
```
AnouckColson, line 30: 🛠️ I'm not sure I understand this correctly. What I understand is something along the lines of this:

```suggestion
For example, you may need to use API gateway credentials for your production environment that are different from those used for your non-production environments. 
To do so, you can set the non-production credentials as a project variable
and set another [variable specific to your production environment](#create-environment-specific-variables) to override it.
```
Maybe I still don't get it, and the phrasing still probably isn't perfect, but my point is that I think clarity could be improved here.

docs/src/development/variables/set-variables.md
```diff
@@ -23,17 +23,42 @@ Application variables are available at both build time and runtime.
...
+<--->
++++
+title=In the Console
++++
+
+To add a project variable, follow these steps:
+
+1. Select the project where you want to create a variable.
+2. Click {{< icon settings >}} **Settings**.
+3. Under **Project settings**, **Variables**.
```
AnouckColson, line 54: 🛠️ 

```suggestion
3. Under **Project settings**, click **Variables**.
```

docs/src/development/variables/set-variables.md
```diff
@@ -48,13 +73,15 @@ Variables have several Boolean options you can set in the Console or the CLI:
...
+So if you want the `foo` variable to be visible at build time but hidden during runtime,
+you can set it by running the following command:
 
 ```bash
 platform variable:create --level project --name foo --value bar --visible-build true --visible-runtime false
 ```
 
-You can also change the variable options after you create them (except for sensitive values, which can't be set to non-sensitive). For example, to make the `foo` variable visible at runtime and hidden during the build, run this command:
+You can also change the variable options after you create them (except for sensitive values, which can't be set to non-sensitive).
+For example, to make the `foo` variable visible at runtime and hidden during the build, run this command:
```
AnouckColson, line 84: 💅 

```suggestion
For example, to make the `foo` variable visible at runtime and hidden during the build, run the following command:
```

docs/src/development/variables/set-variables.md
```diff
@@ -67,29 +94,54 @@ you need to [redeploy](../troubleshoot.md#force-a-redeploy) your environments.
...
-with the exception that visibility in the build and runtime can be set only with the CLI (not in the Console).
+Environment variables share all of the [options available for project variables](#variable-options).
 Environment variables have one additional option:
 
 | Option      | CLI flag        | Default | Description |
 | ----------- | --------------- | ------- | ----------- |
 | Inheritable | `--inheritable` | `true`  | Whether the variable is inherited by child environments. |
 
-This option is useful for  setting production-only values such as credentials. For example, to set a PayPal secret value for only the `main` branch and have it not be readable elsewhere, run:
+This option is useful for  setting production-only values such as credentials.
```
AnouckColson, line 143: 📋 

```suggestion
This option is useful for setting production-only values such as credentials.
```

docs/src/development/variables/set-variables.md
```diff
@@ -67,29 +94,54 @@ you need to [redeploy](../troubleshoot.md#force-a-redeploy) your environments.
...
+For example, to set a PayPal secret value for only the `main` branch and have it not be readable elsewhere, run:
```
AnouckColson, line 144: 🛠️ 

```suggestion
For example, to set a PayPal secret value for only the `main` branch and have it not be readable elsewhere, run the following command:
```

AnouckColson (changes requested):
A couple of questions I asked myself when reviewing, possibly because it's Monday morning still 😄 🛌

docs/src/administration/users.md
```diff
@@ -89,11 +89,11 @@ The user has to create an account before they can contribute to the project.
 Once you add a user to a project, they receive an email with instructions.
 For SSH access changes to apply after you add a user to a project, you have to redeploy each environment by either clicking **Redeploy** in the Console or running `platform redeploy`.
 
-### Delete a user from a project
+### Remove a user from a project
 
-To delete a user, you need to be a [Project Admin](#user-roles).
+To remove a user from a project, you need to be a [Project Admin](#user-roles).
```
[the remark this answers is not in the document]
CollierCZ: This was leftover from before, but I can update it to add consistency.

docs/src/administration/users.md
CollierCZ: Definitely should. Didn't look at everything, but that's a quick fix.

docs/src/development/variables/set-variables.md
CollierCZ: This was wording that I didn't change as part of this PR. I can update it a little, but I didn't want to get into the details here too much. I think `another` is a little confusing, because the point was that there are two types of variables and `another` makes them seem basically the same. But can do at least something.

docs/src/development/variables/set-variables.md
CollierCZ: Again, slightly outside what I was doing, but an easy fix and an improvement.

docs/src/administration/users.md
```diff
@@ -87,61 +88,62 @@ platform user:add user1@example.com -r admin
...
 ```bash
 platform user:delete user1@example.com
 ```
 {{< /codetabs >}}
 
-Once you delete a user, they can no longer access the project.
-After you delete a user from a project or an environment type,
-you must [trigger a redeploy](../development/troubleshoot.md#force-a-redeploy) to propagate SSH access changes to each environment.
+Once you remove a user, they can no longer access the project.
+For SSH access changes to apply after you add a remove a user from a project or an environment type,
```
AnouckColson, line 126: 📋 
```suggestion
For SSH access changes to apply after you remove a user from a project or an environment type,
```

docs/src/administration/users.md
```diff
@@ -155,7 +157,8 @@ and the Contributor role for Development environments:
 platform user:update user1@example.com -r production:v,development:c
 ```
 
-After you change a user's role for an environment type, you must trigger a redeploy each environment to propagate access changes. You can redeploy using the CLI command `platform redeploy`.
+After you change a user's role for an environment type, you must trigger a redeploy each environment to propagate access changes.
```
AnouckColson, line 160: 📋 
```suggestion
After you change a user's role for an environment type, you must redeploy each environment to propagate access changes.
```

docs/src/development/variables/set-variables.md
```diff
@@ -23,38 +23,65 @@ Application variables are available at both build time and runtime.
 ## Create project variables
 
 Add secrets for all environments in project variables
-using [the Console](../../administration/web/configure-project.md#variables) or the CLI.
+using [the Console](../../administration/web/configure-project.md#variables) or the [CLI](../../administration/cli/_index.md).
 
-For example, to create the project variable `foo` with the value `bar`, run:
+For example, you may need to vary API credentials between production and other environments.
+To do so, set the non-production credentials as a project variable
+and then override these credentials for the production environment by setting a [variable specific to that environment](#create-environment-specific-variables).
```
AnouckColson, line 30: 😍 That is so much clearer!

AnouckColson (changes requested):
A couple of minor typos and I believe we're all set! 👍

AnouckColson (dismissed):
🚀 👍

PR #2891, opened 2023-02-10
📝 Rework RabbitMQ overview

### LINKED ISSUE #2856: 📝 Update RabbitMQ

https://docs.platform.sh/add-services/rabbitmq.html

A resource-intensive service. Update for:

- Content accuracy
- Includes what it should
- Fits template
- Fits style guide

### Additional context

_No response_

CollierCZ:
## Why

Closes #2856

## What's changed

Updated the page to fit the style and template


docs/src/add-services/rabbitmq.md
```diff
@@ -66,69 +68,74 @@ highlight=python
...
-   Use the credentials from step 2.
+To open a tunnel, follow these steps.
 
-### From the application container
+1.  SSH into your app container with a flag for local port forwarding:
 
-The application container currently doesn't include any useful utilities to connect to RabbitMQ with. However, you can install your own by adding a client as a dependency in your `.platform.app.yaml` file.
+    ```bash
+    ssh $(platform ssh --pipe) -L 15672:{{< variable "RELATIONSHIP_NAME" >}}.internal:15672
+    ```
```
AnouckColson, line 107: ❓ Several questions on my mind:

- Port 15672 is just an example, right? 
- If so, and if it's meant to be consistent with what's in the example relationship, then why is it 15672 instead of 5672? Is that a mistake?
- I quickly looked at some resources on local port forwarding to gain a rough understanding of what it's about, and what struck me is that there seems to be a source port and a destination port, so shouldn't the two ports in the command be different?
- Is there a way we could use something like {{< variable "LOCAL_PORT" >}} and {{< variable "DESTINATION_PORT" >}} or similar to bring more clarity?

My comments could very well be completely off the mark given my lack of knowledge in this. I don't need you to answer each of my questions if they are irrelevant, I guess my point is to double check that this command is correct and crystal clear for users? If you think it is, then I totally trust your judgement.

AnouckColson (commented):
Some questions came up but they might not warrant any changes. Not sure if my comments are relevant or not, I'm sure you have a better understanding than me!

docs/src/add-services/rabbitmq.md
CollierCZ: The port is the default port for the [management UI](https://www.rabbitmq.com/management.html#usage-ui). As far as I know, you can't customize it. I think it's different than the port for the service itself (which is in the reference), which doesn't involve a UI.

So in this case, it's the source port and that wouldn't change. I suppose it's possible to change the destination port, but it's a standard in RabbitMQ and I'm not sure it'd be necessary. Or maybe if it would be necessary, then I'd imagine people who'd want to do that would know how to do it.

I could add in more explanation of it if you think it would help, but my idea was that it's a command that should work without much effort from users. The number might be clear, I just worried that it's not necessary to know and so trying to explain it might take up more mental effort than it would be worth (as you can see from my poor explanation here taking so much space).

So I think they're good questions. If you think a short explanation would help -- if you think people will ask similar questions themselves -- I can try to add a short explanation. But it's not just an example and I probably wouldn't put in variables in place of the numbers because it makes copying and pasting harder.

AnouckColson (approved):
Based on your explanations (thank you 🙏 ), I think we can push this into production as it is, and only add more context if we get feedback that readers get confused with the command.

PR #2895, opened 2023-02-14
📝 Clarify use case with high-memory plans

CollierCZ:
## Why

The use of high-memory plans wasn't clear. 

## What's changed

Tried to clarify it.


AnouckColson (dismissed):
👍

AnouckColson (approved):

CollierCZ:
We won't publish these examples because they might make things more confusing rather than less (too many why questions involved). But keeping the PR open to have them available in case they're helpful later on.

chadwcarlson:
Closing, as it will be addressed likely with resource management on Upsun.

PR #2896, opened 2023-02-14
✨ Automate more with PHP extensions

CollierCZ:
## Why

The PHP extensions in Grid were in a Markdown table and using them elsewhere was difficult. See Context where this would be useful.

## What's changed

Moved it to data and shortcodes so we can add tables for individual extensions, as in the Xdebug example.


AnouckColson (commented):
I think the spaces are only there for improved readability, so not sure my suggestions bring much value, but here they are for you to decide if you want to apply them. Also, not sure there's an actual rule for parentheses => based on all the examples I've looked at in the repo, the use of spaces before/after the parentheses is very inconsistent. Probably doesn't matter much.

CollierCZ:
The spaces should help with readability. Our current linting is very JavaScript based, so I don't know if it'd be easy to check these kinds of things in the Go-based templates. But your noticing them is helpful.

AnouckColson (approved):

PR #2905, opened 2023-02-17
📝 Show how to sanitize only once

CollierCZ:
## Why

Our examples for [sanitizing databases](https://docs.platform.sh/development/sanitize-db/mariadb.html#sanitize-the-database) ran the sanitization on every deploy, which probably isn't desired. See Context.

## What's changed

Added examples for doing it only on initial deploy.


AnouckColson (approved):
👍

PR #2907, opened 2023-02-17
📝 Document scalability

### LINKED ISSUE #2871: 📝 Document scalability

We should document how scaling works for live sites. This should include [auto-scaling](https://platform.sh/features/auto-scaling/) and also less automated solutions, including distributing resources differently and increasing plan size temporarily (mentioning that only charged for time that's used).

### Additional context

A start on this is in the history of #2852. Need to add at least auto-scaling.

CollierCZ:
## Why

Closes #2871

## What's changed

Added a page on scaling and redirected Dedicated scaling there.. Didn't get into the difference between horizontal and vertical scaling for now.


docs/src/environments/scalability.md
```diff
@@ -0,0 +1,61 @@
...
+These resources are split up between the apps and services you've defined.
+
+You might occasionally get increased load on your site,
+such as during a major product launch or a sales event like Black Friday.
+This pressure might cause problems in your site or even bring it down.
+At such times, you want your site to have enough resources to handle the additional load.
+
+How your site handles the need to scale depends on your plan.
+On higher tiers with the Observability Suite,
+everything is taken care of you automatically.
```
AnouckColson, line 18: 🛠️ ❓ The sentence is either missing the word _for_ before _you_, or the word _you_ is superfluous. Either way, I'm wondering if the sentence shouldn't be in the active voice instead, something like "Platform.sh takes care of everything on your behalf" or similar?

AnouckColson (commented):
One minor comment, otherwise it looks good to me!

AnouckColson (changes requested):
Selected Comment instead of Request changes by mistake, sorry 😞

docs/src/environments/scalability.md
CollierCZ: I had the sentence active at the start, but it felt longer that way and `Platform.sh` didn't feel specific enough to really add a lot of help. I mean that it felt like saying `the system does X`, where `the system` isn't adding a lot of value. If you like it a lot more, it could be `Platform.sh takes care of everything for you`. (The mistaken `you` was left over from when I had the sentence like that before, I think.) I like active, but I also like short and specific, so not sure which works better. :shrug:

AnouckColson (approved):

PR #2921, opened 2023-02-22
📝 Update for CLI installation script

CollierCZ:
## Why

The CLI now has an installation script for direct installs. 

## What's changed

Updated the docs to include this script and avoid complicated workarounds with Homebrew.
Also used in our GitHub Workflows to see if it helps with flaky pipelines.


CollierCZ:
The workflow works, though didn't seem faster: https://github.com/platformsh/platformsh-docs/actions/runs/4241357942/jobs/7371471778

The current failing check is probably because of the change and running against the main branch.

AnouckColson (approved):
All good

PR #2937, opened 2023-03-01
Define Route for multi-app: add web.locations definition example

flovntp:
## Why
When i first try to create a multi-app project, i can't make it till @chadwcarlson helps me debugging my `web.locations` definition into my `.platform.app.yaml` file, adding this `passthru: /api/index.php` settings.
So i guess it worths to be mention here. 

Closes #{ISSUE_NUMBER}


CollierCZ:
I can see how an example might help.
We Do have examples for the app configuration part elsewhere:
- https://docs.platform.sh/create-apps/multi-app.html#unified-app-configuration
- https://docs.platform.sh/create-apps/multi-app.html#routes
Could link to those or adjust them. Using the keys like this PR does isn't the most clear. And ideally the example wouldn't be so PHP-centric, but many of the examples are :shrug:

chadwcarlson (approved):

PR #2941, opened 2023-03-03
:memo: Document custom domains on non-production environments

AnouckColson:
## Why

It is now possible to add custom domains to non-production (staging & development environments).
See context for more information on the feature.

## What's changed

Added a page dedicated to custom domains on non-production environments (explanation of the feature + how-tos).
Made sure the page fits our style guide, is clear and comprehensive.


docs/src/domains/steps/non-production-custom-domains.md
```diff
@@ -0,0 +1,98 @@
...
+You can then access your staging environment through `staging.example.com`
+and still access your production environment through `example.com`.
+
+{{< /note >}}
+
+If you have multiple production environments,
+when you create a non-production custom domain,
+Platform.sh retrieves the `{default}` placeholder in your [routes configuration](../../define-routes/_index.md).
+It then replaces it with the correct production custom domain
+according to what you specify [in the setup command](#add-a-custom-domain-to-a-non-production-environment).
```
nicogommen, line 32: This is the hard part to explain, and after more thoughts, I think we could say something like this:

> If you have multiple custom domains on your production environment,
> when you create a non-production custom domain,
> you don't have to update your [routes configuration](../../define-routes/_index.md), 
> Platform.sh automatically figures out the routing of your non-production environment, based on your production custom domains in your `routes.yaml` file and the non-production custom domains attached to them. 

Let me know what you think.

docs/src/domains/steps/non-production-custom-domains.md
```diff
@@ -0,0 +1,98 @@
+---
+title: Set up a custom domain on your non-production environments
+sidebarTitle: Non-production custom domains
```
nicogommen, line 3:
```suggestion
sidebarTitle: [Beta] Non-production custom domains
```

docs/src/domains/steps/non-production-custom-domains.md
```diff
@@ -0,0 +1,98 @@
...
+Note that you must be a project admin to add non-production custom domains.
+You can add up to 5 non-production custom domains per project.
+
+If you delete a production custom domain,
+all of the attached non-production custom domains are deleted too.
+
+If you downgrade from an Elite or Enterprise plan to a Professional plan,
+all of the custom domains set on non-production environments are automatically removed.
+Downgrading your plan doesn't affect custom domains set on your production environments.
+
```
nicogommen, line 43: Something I just learned today (not sure it's the right place to add this so feel free to move it):

> For Dedicated Gen 2 projects, this feature can only be used on the Grid environments, and not the dedicated ones. It is however possible to add a custom domain on a dedicated environment, but only via filing a support ticket.

nicogommen (changes requested):

docs/src/domains/steps/non-production-custom-domains.md
AnouckColson: That's great. I only split the explanation into several sentences and a bulleted list for improved readability. Also removed the explicit mention of the .yaml file to remain consistent with our docs style guide.

docs/src/domains/steps/non-production-custom-domains.md
AnouckColson: ❓ I added the information, but how does it work for DG3 environments then? Do users need to go through support as well, or is the feature simply not available to them?

docs/src/domains/steps/non-production-custom-domains.md
```diff
@@ -0,0 +1,98 @@
...
+## Add a custom domain to a non-production environment
+
+The creation of non-production custom domains
+from the [Platform.sh Console](../../administration/cli/_index.md) isn't available in the Beta version,
+but will be soon. In the meantime, use the [Platform.sh CLI](/administration/cli/_index.md).
```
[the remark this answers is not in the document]
AnouckColson: I added it to the **Get started** section at the beginning with all the prerequisites/things people should know before they start using the feature.

docs/src/domains/steps/non-production-custom-domains.md
```diff
@@ -0,0 +1,120 @@
+---
+title: Set up a custom domain on your non-production environments
+sidebarTitle: (Beta) Non-production custom domains
+weight: 2
+description: Learn how to set up custom domains on your staging and development environments
+betaFlag: true
+---
+
+You can add a custom domain to a staging or development environment without modifying your [routes configuration](../../define-routes/_index.md).
```
chadwcarlson, line 9:
```suggestion
You can add a custom domain to a non-production environment (`staging` or `development` environment types) without modifying your [routes configuration](../../define-routes/_index.md).
```

chadwcarlson: Maybe worth making the definition explicit right at the start?

docs/src/domains/steps/non-production-custom-domains.md
```diff
@@ -0,0 +1,120 @@
...
+
+- The production custom domains in your existing [routes configuration](../../define-routes/_index.md)
+- The non-production custom domains attached to each of those production custom domains
+
+## Get started with non-production custom domains
+
+You need:
+
+- A production environment with at least one custom domain already set up
+- At least one non-production (staging or development) environment
```
chadwcarlson, line 41: - The environment `Dev` needs to exist for me to run `platform domain:add development.mysite.com --environment Dev --replace mysite.com`?
- Say that environment _did_ exist when I defined this non-production URL, but then the environment was removed. Would re-creating the `Dev` environment after the fact respect the previous custom domain configuration? Or would I need to reconfigure it? A later section implies that the configuration is only removed when 1) the production URL is deleted, or 2) the user explicitly deletes it to define a new one.

docs/src/domains/steps/non-production-custom-domains.md
```diff
@@ -0,0 +1,120 @@
+---
+title: Set up a custom domain on your non-production environments
+sidebarTitle: (Beta) Non-production custom domains
+weight: 2
+description: Learn how to set up custom domains on your staging and development environments
+betaFlag: true
+---
+
+You can add a custom domain to a staging or development environment without modifying your [routes configuration](../../define-routes/_index.md).
+To do so, you need to attach the new non-production custom domain
```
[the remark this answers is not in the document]
nicogommen: Thank you for your input @chadwcarlson.
This is definitely one of the things where we are struggling to find the right wording...
I can see what you are trying to do, but I feel like the second phrase is still a bit confusing.

Maybe something like this:
> A custom non-production domain configured on your project replaces the production domain for the corresponding environment, instead of the domain automatically generated by Platform.sh. If you have multiple production domains, you will have to select which one you are replacing when adding  a custom non-production domain.

Still wordy... and we are back with using "replace" instead of "attach".
So I don't know if this is helping 🙈

docs/src/domains/steps/non-production-custom-domains.md
nicogommen: > The environment Dev needs to exist for me to run platform domain:add development.mysite.com --environment Dev --replace mysite.com?
=> Yes, the environment needs to exist. The CLI returns an error if we try to add a domain to a non-existing environment. I guess that's what we try to say with « _At least one non-production (staging or development) environment_ », but maybe we can be a bit clearer.

> Say that environment did exist when I defined this non-production URL, but then the environment was removed. Would re-creating the Dev environment after the fact respect the previous custom domain configuration? Or would I need to reconfigure it? A later section implies that the configuration is only removed when 1) the production URL is deleted, or 2) the user explicitly deletes it to define a new one.
=> That's a good question, and actually a test that I didn't do. So I tried, and the non-production domain is deleted as well when we delete the environment. If you recreate the environment with the same ID, you have to recreate the non-production domain as well. We should probably add that to the documentation somewhere.

Thanks again for the feedback 🙂

docs/src/domains/steps/non-production-custom-domains.md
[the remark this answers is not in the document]
chadwcarlson: Definitely helps, thanks

chadwcarlson (dismissed):

chadwcarlson (approved):

Issue #2943, opened 2023-03-03
Improve services order in the docs


AnouckColson:
### Where on docs.platform.sh should be changed?

https://docs.platform.sh/add-services.html

### What exactly should be updated?

There seems to be no real logic behind the order of the services pages in the **Add services** section. They are mainly in alphabetical order except for OpenSearch being next to Elasticsearch. 

We need to make the order more logical. See context.

PR #2987, opened 2023-03-15
Enable sectionBefore: grouping Stacks and services

chadwcarlson:
## Why

Closes https://github.com/platformsh/platformsh-docs/issues/2943

## What's changed

In an attempt to address #2943, and solve something similar for frameworks in **Stacks**, this PR

- leverages the `sectionBefore` attribute used in deeper levels
- enables that attribute at both the top and second levels
- Reorders **Stacks** to be 1) grouped by runtime, 2) Alphabetical within that runtime, 3) Reverse alphabetizing within **Stacks**. (Logic: Java has less importance currently than PHP and Python at this point, where it would appear first otherwise.)
- ~Reorders **Add services** around service "types", as is already defined in the titles on their individual pages. Whether these titles are appropriate is open to some debate if there's feedback. Services are then alphabetized within that group.~
- **UPDATE** After internal review/discussion, Services are reordered alphabetically. Only Stacks leverage `sectionBefore`.

Going forward, this enables other groupings within **Guides and reference** if we want to go this route.

### Stacks

CleanShot 2023-03-15 at 14 16 28


chadwcarlson:
Match to use Javascript/Node.js in place of Node.js

AnouckColson (approved):

Issue #3013, opened 2023-03-23
Generalize banner shortcodes


chadwcarlson:
### Where on docs.platform.sh should be changed?

More detail on where changes could be made below

### What exactly should be updated?

We're getting more requests to tweak banner content, but Docs lacks a generalized banner shortcode. Instead, most of that content is hardcoded into multiple shortcodes for specific purposes from their beginnings. 

The goal here would be to generalize this process a bit - sane defaults that make it easier to update/customize content within a banner for a particular page

### Load path for banners:

There are two different types of banners that could be generalized currently, and potentially a third. The first two - Tier availability and Beta features, start with the same load path:

- [theme/baseof.html](https://github.com/platformsh/platformsh-docs/blob/0e576a23cbc2a735a5c39662b0f95901ba5db916/docs/themes/psh-docs/layouts/_default/baseof.html#L41)
- [theme/single.html](https://github.com/platformsh/platformsh-docs/blob/0e576a23cbc2a735a5c39662b0f95901ba5db916/docs/themes/psh-docs/layouts/_default/single.html#L14) or [theme/list.html](https://github.com/platformsh/platformsh-docs/blob/0e576a23cbc2a735a5c39662b0f95901ba5db916/docs/themes/psh-docs/layouts/_default/list.html#L2)
- [theme/page-content.html](https://github.com/platformsh/platformsh-docs/blob/0e576a23cbc2a735a5c39662b0f95901ba5db916/docs/themes/psh-docs/layouts/partials/page-content.html#L8) for Beta flags, and [theme/page-content.html ](https://github.com/platformsh/platformsh-docs/blob/0e576a23cbc2a735a5c39662b0f95901ba5db916/docs/themes/psh-docs/layouts/partials/page-content.html#L2)for Tier availability

#### Split 1 - Beta flags

As in https://docs.platform.sh/domains/steps/custom-non-production-domains.html

Depends on `betaFlag` set in page parameters:

https://github.com/platformsh/platformsh-docs/blob/0e576a23cbc2a735a5c39662b0f95901ba5db916/docs/src/domains/steps/custom-non-production-domains.md?plain=1#L1-L7

- [partials/beta-flag/banner.html](https://github.com/platformsh/platformsh-docs/blob/main/docs/themes/psh-docs/layouts/partials/beta-flag/banner.html#L2) -- If `.Params.betaFlag` is set on the page
[...]

PR #3016, opened 2023-03-23
Generalized banners

chadwcarlson:
## Why

Closes https://github.com/platformsh/platformsh-docs/issues/3013

## What's changed

- All banners go through the same partial: `docs/themes/psh-docs/layouts/banners/banner.html`
- A banner is created on a page with the `banner` front matter parameter, which should include the banner `type`
- All banners have default values for `title` and `body` defined in `docs/data/banners.yaml`, which is displayed if only `banner.type` is defined.
- `title` and `body` can be overwritten for a given page when `type` is defined
- Banners without a defined `type` can be displayed, so long as `title` and `body` are defined
- The Premium feature (currently only used for MongoDB enterprise) is updated, since it depended on the old banner partials
- The `betaFlag` parameter was removed from the Django deploy guide - copypasta

See https://github.com/platformsh/platformsh-docs/issues/3013 for more details

### DIFF
--- a/docs/data/banners.yaml
+++ b/docs/data/banners.yaml
@@ -1,0 +1,34 @@
+# PAGE BANNERS
+#
+# To enable on a page:
+# 
+# banner:
+#   type: beta 
+#   # title: Override the default values below by setting a banner.title
+#   # body: Override the default values below by setting a banner.body
+#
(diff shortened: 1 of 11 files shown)


AnouckColson (approved):
This is awesome, thanks @chadwcarlson!

PR #3030, opened 2023-03-28
🐛Clarify timezones

trolologuy:
## Why

The wording might lead to confusion regarding how the different parts fit into each other.

## What's changed

Moved a sentence to make the relationship hopefully clearer.


trolologuy:
See context for an example of confusion.

I think we might want to move all cron differences, with examples to one page, what do you think, @AnouckColson ?

So for example:
```
timezone: "Europe/Paris"
variables:
  php:
    date.timezone: "Europe/Paris" 
```
+ the project-level one.

AnouckColson (approved):

PR #3080, opened 2023-04-19
:memo: Remove Transparency report page and update intro

AnouckColson:
## Why

Continuation of [#3039](https://github.com/platformsh/platformsh-docs/pull/3039/files). The section on transparency reports will also live in the Trust Center so needs to be removed from docs.

## What's changed

Removed transparency section. Redirect is already in place.


trolologuy (approved):
Looks good to me. The site builds now.

→Didn't check that the redirects work.

PR #3086, opened 2023-04-20
Updating Grid, DG2 and DG3 versions for runtimes and services

### LINKED ISSUE #2945: Update versions in registry file

The registry.json file where all the supported/deprecated versions displayed in the whole docs come from.

We've received feedback that some information on supported versions (PHP, OpenSearch) for DG2 might be outdated (see context). Now might be a good time to run a complete update on the file.

Explaining what we mean by "supported" or "deprecated" is also warranted (see context).

### Additional context

_No response_

remilejeune2:
## Why

Closes #2945 

## What's changed

Update of supported versions on Grid and DG3.


docs/data/registry.json
```diff
@@ -35,6 +35,12 @@
       ]
     }
   },
+  "versions-dedicated-gen-3": {
```
chadwcarlson, line 38: Looks like this is now a top-level key, when it should be within a service/runtime

chadwcarlson: In trying to replicate this, placing `versions-dedicated-gen-3` back within `chrome-headless` fixed the local build. Registry parser was trying to find a `versions-dedicated-gen-3.docs.relationship_name`, which doesn't exist.

chadwcarlson: Top of this file should be changed to:

```
{
  "chrome-headless": {
    "description": "",
    "disk": false,
    "docs": {
      "relationship_name": "chromeheadlessbrowser",
      "service_name": "headlessbrowser",
      "url": "/add-services/headless-chrome.html"
    },
    "endpoint": "http",
    "min_disk_size": null,
    "name": "Headless Chrome",
    "repo_name": "chrome-headless",
    "runtime": false,
    "type": "chrome-headless",
    "versions": {
      "deprecated": [],
      "supported": [
        "95",
        "91",
        "86",
        "84",
        "83",
        "81",
        "80",
        "73"
      ],
      "legacy": [
        "86",
        "84",
        "83",
        "81",
        "80",
        "73"
      ]
    },
    "versions-dedicated-gen-3": {
      "deprecated": [],
      "supported": [
        "95"
      ]
    }
  },
```

chadwcarlson: cc @AnouckColson @remilejeune2

chadwcarlson (commented):
For now, I think we shouldn't make this change. 

This file is pulled from the https://examples.docs.platform.sh project. Going forward, I think it makes sense to dismantle that project and have everything in docs - but for now we should keep the behavior the same everywhere.

docs/static/files/fetch/examples/relationships/influxdb
```diff
@@ -1,11 +1,11 @@
 {
-    "service": "influxdb18",
+    "service": "influxdb23",
```
chadwcarlson, line 2: For now, I think we shouldn't make this change. 

This file is pulled from the https://examples.docs.platform.sh project. Going forward, I think it makes sense to dismantle that project and have everything in docs - but for now we should keep the behavior the same everywhere.

chadwcarlson: Plus, it gets overwritten during the deployment anyway https://bffpr-3086-5xcwfqi-652soceglkw4u.eu-3.platformsh.site/add-services/influxdb.html#relationship-reference

AnouckColson (approved):

PR #3104, opened 2023-04-24
Document Lando workaround

### LINKED ISSUE #3054: 🐛 Issue on the page Use Lando for local development | Platform.sh Docs

https://docs.platform.sh/development/local/lando.html#3-initialize-lando

It says to run this command:
lando init –recipe platformsh –source cwd –platformsh-auth ACCESS TOKEN

but when you put in your access token, it fails. 

If you just try lando init and do the following options:
current directory
recipe: platformsh
enter a platform.sh access token: <puts in API token from the API Tokens panel on platform.sh web console.>

fails without an error. 

They call it an API token, you call it an access token, is this the right spot to pull the access token from?
Click User, go to API tokens, generate new token.

### Additional context

_No response_

AnouckColson:
## Why

Closes #3054 

There was an issue with the `recipe` argument showing as invalid and preventing users from successfully initializing Lando using the documented command.
[Issue was fixed on Lando's side](https://github.com/lando/platformsh/pull/188) but users need to install the latest version of the Platform.sh plugin until this change is reflected on the next Lando general release.

## What's changed

Added a temporary note below the command.

### DIFF
--- a/docs/src/development/local/lando.md
+++ b/docs/src/development/local/lando.md
@@ -57,7 +57,8 @@ Run <code>lando init --recipe platformsh --source cwd --platformsh-auth {{% vari
 
 {{< note >}}
 
-If you get an error stating that the `recipe` argument is invalid,
+If for some reason you get an error using the Platform.sh recipe,
+be sure to 
 [install the latest version of the Platform.sh plugin](https://docs.lando.dev/platformsh/getting-started.html#custom-installation)
 and run the command again.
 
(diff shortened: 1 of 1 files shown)


docs/src/development/local/lando.md
```diff
@@ -55,6 +55,14 @@ Otherwise, access the directory with your project.
 
 Run <code>lando init --recipe platformsh --source cwd --platformsh-auth {{% variable "API_TOKEN" %}}</code> and follow the instructions provided by the interactive prompt.
 
+{{< note >}}
+
+If you get an error stating that the `recipe` argument is invalid,
+[install the latest version of the Platform.sh plugin](https://docs.lando.dev/platformsh/getting-started.html#custom-installation)
+and run the command again.
```
chadwcarlson, line 62:
```suggestion
If for some reason you get an error using the Platform.sh recipe,
be sure to 
[install the latest version of the Platform.sh plugin](https://docs.lando.dev/platformsh/getting-started.html#custom-installation)
and run the command again.
```

chadwcarlson (approved):

PR #3116, opened 2023-04-28
[hotfix] sets the platform cli token at the workflow level instead of at step

gilzow:
sets the platform cli token at the workflow level instead of the step level for every step


chadwcarlson (approved):

chadwcarlson:
Activated manually here https://hotfix-set-cli-token-at-ug2rzai-652soceglkw4u.eu-3.platformsh.site/

PR #3133, opened 2023-05-05
:memo: Add note about versions to service and language pages

AnouckColson:
## Why

To avoid confusion, let's make it clear that users can select the major version but not the minor one.

## What's changed

Created a shortcode and added it to every language and service page in the docs.


docs/layouts/shortcodes/major-minor-versions-note.md
```diff
@@ -0,0 +1,4 @@
+You can select the major version.
```
chadwcarlson, line 1: Maybe we can change this to accept a `configMinor = true|false`?

chadwcarlson (approved):

PR #3134, opened 2023-05-08
Update redirects.md to indicate that query-string based redirect are unsupported.

bligneri:
Add the information that we do not support query-string redirects and that this is to be implemented in the application itself.

## Why

Closes #{ISSUE_NUMBER}


chadwcarlson (approved):

chadwcarlson:
Thanks @bligneri !

Issue #3142, opened 2023-05-10
GA4 changes


chadwcarlson:
### Where on docs.platform.sh should be changed?

- head should include GTM script sections
- body should include GTM noscript section
- Triggered events should hit the data layer

#### Events

- Login
- Free trial
- Page viewed
- Search results

### What exactly should be updated?

See above

PR #3143, opened 2023-05-10
GA4

chadwcarlson:
## Why

Closes https://github.com/platformsh/platformsh-docs/issues/3142

## What's included

- [x] head should include GTM script sections
- [x] body should include GTM noscript section
- [x] Events: Login (need to verify)
- [x] Events: Free trial (need to verify)
- [x] Events: Page viewed
- [x] Events: Search results


chadwcarlson:
"the link_url should be the actual page where the click was done and not the destination."

AnouckColson (approved):

PR #3144, opened 2023-05-10
[Enhancement] Cache retrieved files, use authenticated GH API

gilzow:
This is part 2 of #3118 . 

Moves all calls to the github API (both those to api.github.com, as well as raw.githubusercontent.com) to use an authenticated API call. This will ensure that we have the maximum available requests (5K per hour), and help reduce the possibility of hitting the maximum during builds. 

That data, along with data from examples.docs.platform.sh, is then written to the build cache area, so that subsequent builds can use this cache instead of having to retrieve all the data again.  This should assist in speeding up builds. Unscientific timings show about a 30 second savings in builds once the cache is built.


chadwcarlson:
To merge into this one https://github.com/platformsh/platformsh-docs/pull/3149

chadwcarlson (approved):

PR #3150, opened 2023-05-11
:memo: Update custom non-production domains page

AnouckColson:
## Why

An example needed to be updated to show that any domain can be used for a non-prod environment, and not just a subdomain of the prod. I took this opportunity to rework the whole page as successive updates made it more and more comprehensive but also affected readability.

## What's changed

Reorganized content, in particular the Before you begin list to make the overall page seem lighter. Gave another try at better explaining the foundation concept (replacement notion) to improve clarity. Updated the example.

PR #3151, opened 2023-05-11
Update add custom domain example

nicogommen:
Update add custom domain example to show that you can use any domain, not just a subdomain of the production.

## Why

Closes #{ISSUE_NUMBER}


AnouckColson:
Addressed in #3150

PR #3172, opened 2023-05-17
:memo: Update source integration pages

AnouckColson:
## Why

Closes #3152 

## What's changed

Removed options that weren't mandatory from the setup procedures.
Corrected wrong option for GitLab integration.
Added information about necessary user permissions to be able to add a webhook successfully depending on tool.

PR #3189, opened 2023-05-25
Erratum source integrations

AnouckColson:
## Why

Following #3172 I realized that:

- The shortcode for displaying different user permissions & links depending on the CI/CD tool didn't work as expected (links work well but it always displays Admin rights when it should be Owner or Maintainer on the GitLab page).
- GitLab integration uses `--server-project` instead of `--repository` which I failed to address in #3172.
- The `--base_url` flag, while not a strictly mandatory flag (i.e. it's technically possible to set up an integration without it) _is_ needed if people self host,so I put it back in the doc.

## What's changed

- Fix the user permissions shortcode.
- Create a shortcode for `--server-project` vs. `--repository` flags.
- Put the info on `--base_url` back in.

### DIFF
--- a/docs/layouts/shortcodes/source-integration/validate.md
+++ b/docs/layouts/shortcodes/source-integration/validate.md
@@ -22,7 +22,7 @@ If you see a message that the webhook wasn't added, add one manually.
 {{ else if eq $source "GitLab" }}
   {{ $reqdPerms := "Maintainer or Owner" }}
   {{ $permsLink = "https://docs.gitlab.com/ee/user/permissions.html" }}
-{{ end }} 
+{{ end }}
 
 To configure a webhook on a {{ $source }} repository,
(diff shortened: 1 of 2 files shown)


docs/layouts/shortcodes/source-integration/validate.md
```diff
@@ -22,7 +22,7 @@ If you see a message that the webhook wasn't added, add one manually.
 {{ else if eq $source "GitLab" }}
   {{ $reqdPerms := "Maintainer or Owner" }}
   {{ $permsLink = "https://docs.gitlab.com/ee/user/permissions.html" }}
-{{ end }} 
+{{ end }}
```
AnouckColson, line 25: @chadwcarlson And just above this line the one that doesn't work. Keeps displaying "Admin" on the GitLab page.

chadwcarlson (approved):

PR #3191, opened 2023-05-26
Make Trust Center redirect more obvious

AnouckColson:
## Why

As a follow-up to #3188, make redirect to Trust Center more obvious for users looking for info on pentests for example.


chadwcarlson (commented):
@AnouckColson I feel like the note is good, but grouping a little strangely above the other section links? 

What if we placed it at the top of the page, even before "Platform.sh is compliant with many regulations in favor of the protection of user data,"?

chadwcarlson (approved):

PR #3202, opened 2023-06-06
Update custom non-production domains doc

AnouckColson:
## Why

Console support is being added and the terminology surrounding the feature updated so we needed to update the doc page accordingly.

## What's changed

Added steps to add a custom non-production domain from the Console and updated the terminology.

### DIFF
--- a/docs/src/domains/steps/custom-non-production-domains.md
+++ b/docs/src/domains/steps/custom-non-production-domains.md
@@ -55,19 +55,17 @@ You need:
 
   If you have a {{% names/dedicated-gen-2 %}} project,
   currently you can only add a custom domain to the dedicated environments of your project (production and staging).
-  To do so, contact [Support](https://console.platform.sh/-/users/~/tickets/open).
+  To do so, [contact Support](https://console.platform.sh/-/users/~/tickets/open).
 
   {{< /note >}}
 
   If you use a [Managed Fastly](../cdn/managed-fastly.md) CDN,
   it needs to be configured to operate with custom non-production domains.
-  For more information, contact [Support](https://console.platform.sh/-/users/~/tickets/open).   
+  For more information, [contact Support](https://console.platform.sh/-/users/~/tickets/open).   
 - A production environment with at least one custom domain already set up
 - At least one non-production (staging or development) environment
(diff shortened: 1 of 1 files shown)


chadwcarlson (approved):

docs/src/domains/steps/custom-non-production-domains.md
```diff
@@ -90,68 +88,68 @@ Downgrading your plan doesn't affect custom domains set on your production envir
...
+   ```
 
 <--->
 +++
 title=In the Console
 +++
 
-1. In the Console, open your non-production environment.
-2. Click **URLs** and copy the URL to your site excluding `https://`.
+1. Get the target for your non-production environment. To do so, open the environment in the Console
```
nicogommen, line 124: We can now change this instruction with something easier for the user.
They can go to the new `Domains` tab of the environment settings.
They will find here a `Configure your domain` block, where there is the `CNAME target` to copy.

We should also change it on [this page](https://non-prod-custom-xmpenxa-652soceglkw4u.eu-3.platformsh.site/domains/steps.html#2-get-the-target-for-your-project).

docs/src/domains/steps/custom-non-production-domains.md
nicogommen, line 1: Some other changes that we can prepare for the release:
- Remove the BETA flag in the menu.
- Remove the BETA block at the top of the page, but keep the mention that the feature is only for Enterprise and Elite customers
- Update the version of the CLI required (If you use the Platform.sh CLI, you need version 4.? or later) => I don't have yet the correct version, but adding this bullet-point so we don't forget.

Questions:
- Should we have the Console instructions for the other operations too (list, get, delete)? Maybe list and get don't make much sense, but perhaps delete?

docs/src/domains/steps/custom-non-production-domains.md
AnouckColson: Done

docs/src/domains/steps/custom-non-production-domains.md
AnouckColson: All done

chadwcarlson (approved):

PR #3203, opened 2023-06-06
Update WAF page

AnouckColson:
## Why

A new rate limiting configuration needs to be subtly mentioned in the docs. See context.


jww-sh (approved):

jww-sh (approved):
Looks good

PR #3230, opened 2023-06-18
Refactor scheduled link checker

### LINKED ISSUE #3225: Rework schedule link checks

./.github/workflows/scheduled-check-links.yaml

The current set up shows up as a _workflow_ failure if broken links are detected, even though the workflow actually succeeded (due to the link scanner exiting with a non-zero status and `bash -e` being the default in a workflow). 

In addition, the link scanner in user appears to detect any server response other than 200 as a broken link, even if the links are valid. In the last scan, out of 200ish broken links, **zero** were actually broken. The vast majority were various versions of redirects. 

Suggestion is to move to a different scanner that has far fewer false positives and to restructure the workflow to only show a failure when the workflow itself fails.

### Additional context

_No response_

gilzow:
## Why

Closes #3225 

## What's changed

- Refactors [workflow](https://github.com/platformsh/platformsh-docs/blob/main/.github/workflows/scheduled-check-links.yaml) to use  [Linkchecker](https://linkchecker.github.io/linkchecker/) to check for broken links
- Adds a `workflow_dispatch` to allow for running the linkchecker on-demand
- Adds a [small python script](https://github.com/platformsh/linkchecker-xml2md) for converting the xml output from linkchecker into a markdown table
- if broken links are found, adds an issue with the markdown table as the issue body and a label of `broken-links`

### DIFF
--- a/.github/workflows/scheduled-check-links.yaml
+++ b/.github/workflows/scheduled-check-links.yaml
@@ -3,69 +3,75 @@ name: Regularly scheduled check links
 on:
   schedule:
     - cron: "0 2 * * 1,4"
+  workflow_dispatch:
+    inputs:
+      url:
+        description: 'Site to check for broken links'
+        required: true
+        type: string
+        default: 'https://docs.platform.sh/'
 
 jobs:
   check-links:
     name: Check links
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/setup-go@v2
+      - uses: actions/setup-python@v4
         with:
-          stable: 'true'
+          python-version: '3.10'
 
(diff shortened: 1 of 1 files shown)


AnouckColson (dismissed):

gilzow:
We can also add an `External` column with True/False values in each row, if that would be of benefit

gilzow:
Something to think about is improving the way we manage the "ignore" urls. For now they're all hard-coded into the workflow directly. Seems like maybe we want to change that so it's more easily updated?

chadwcarlson (approved):

PR #3275, opened 2023-07-12
Remove wrong default values on flexible resources page

AnouckColson:
## Why

The docs showed default values for the `base_memory` and `memory_ratio` options (`resources` key) that conflicted with the minimum values advertised within the same table, which generated confusion.

## What's changed

Support confirmed that there are no default values so I removed the `Default` column from the table.


trolologuy (approved):
LGTM 🌞

PR #3309, opened 2023-07-28
Add InfluxDB 2.7 to registry

AnouckColson:
## Why

A new image is available for InfluxDB 2.7

## What's changed

Registry.json amended


docs/src/add-services/influxdb.md
```diff
@@ -75,3 +111,61 @@ To export your data from InfluxDB, follow these steps:
...
+## Upgrade to version 2.3 or later
+
+### From a previous 2.x version
+
+From version 2.3 onwards, the structure of relationships changes.
+
+If you're using a prior 2.x version, your app might currently rely on pulling the `bucket`, `org`, `api_token`,
+or `username` values available in the [`PLATFORM_RELATIONSHIPS` environment variable](../development/variables/use-variables.md#use-platformsh-provided-variables).
+
+If so, to ensure your upgrade is successful, make the following changes to your service configuration:
```
chadwcarlson, line 124:
```suggestion
If so, to ensure your upgrade is successful, make the following changes to your connection logic:
```

docs/src/add-services/influxdb.md
```diff
@@ -75,3 +111,61 @@ To export your data from InfluxDB, follow these steps:
...
+
+- Rename the `user` key to `username`.
+- Move the `org`, `bucket` and  `api_token` keys so they're contained in a dictionary under the `query` key.
+
+So the relationship information (available through the [`PLATFORM_RELATIONSHIPS` environment variable](../development/variables/use-variables.md#use-platformsh-provided-variables) or by running `platform relationships`) would look similar to the following:
```
chadwcarlson, line 129:
```suggestion
If you're relying on any other attributes connecting to InfluxDB, they remain accessible as top-level keys from the [`PLATFORM_RELATIONSHIPS` environment variable](../development/variables/use-variables.md#use-platformsh-provided-variables), aside from those addressed above:

```

docs/src/add-services/influxdb.md
```diff
@@ -75,3 +111,62 @@ To export your data from InfluxDB, follow these steps:
...
     ```
+
+## Upgrade to version 2.3 or later
+
+### From a previous 2.x version
+
+From version 2.3 onwards, the structure of relationships changes.
+
+If you're using a prior 2.x version, your app might currently rely on pulling the `bucket`, `org`, `api_token`,
+or `username` values available in the [`PLATFORM_RELATIONSHIPS` environment variable](../development/variables/use-variables.md#use-platformsh-provided-variables).
```
dephekt, line 122: Prior versions this would be called `user`, so it probably doesn't make sense to call it `username` here as that's what it would be changing to, not what they would be potentially using right now.

dephekt (changes requested):

dephekt (approved):

PR #3318, opened 2023-08-02
Recommend persistent Redis in Drupal guide

AnouckColson:
## Why

Redis ephemeral config came first on https://docs.platform.sh/guides/drupal9/redis.html#1-configure-the-service which could be taken as we recommend ephemeral vs persistent, when in fact it's the opposite.

## What's changed

Added formatted note to Before you begin section to recommend Redis persistent + swapped the order of the configs to highlight Redis persistent first.

### DIFF
--- a/docs/themes/psh-docs/layouts/shortcodes/endpoint-description.md
+++ b/docs/themes/psh-docs/layouts/shortcodes/endpoint-description.md
@@ -77,6 +77,9 @@ You can define `<SERVICE_NAME>` and `<RELATIONSHIP_NAME>` as you like, but it's 
   {{ if eq $type "postgresql" }}
    {{ $extension_name = "pdo_pgsql" }}
   {{ end }}
+  {{ if eq $type "redis-persistent" }}
+   {{ $extension_name = "redis" }}
+  {{ end }}
 For PHP, enable the [extension](/languages/php/extensions.html) for the service:
 
 ```yaml {location=".platform.app.yaml"}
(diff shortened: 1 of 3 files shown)


docs/themes/psh-docs/layouts/shortcodes/endpoint-description.md
```diff
@@ -19,20 +19,22 @@
...
   {{ else }}
   the `{{ $type }}` type{{ end }}:
 
 <!-- Create an example services.yaml file from data in the registry. -->
 {{ partial "examples/servicedefn" $data }}
 
-{{ if eq $type "redis" }}
-Alternatively, use the `redis-persistent` type for persistent Redis:
+{{ if eq $type "redis-persistent" }}
+Note that persistent Redis requires a disk to store data.
```
chadwcarlson, line 30:
```suggestion
Note that persistent Redis requires `disk` to store data.
```

chadwcarlson: Also, since I can't comment on what you haven't changed, I think we need to update [this line](https://github.com/platformsh/platformsh-docs/pull/3318/files#diff-4304e44d631e4b1504a05dc250212f7de83833c75ac42ca8e149988f1ef573feL71) in this file as well.

Called with

```
{{% endpoint-description type="redis-persistent" noApp=true onlyLanguage="php" php=true /%}}
```

`type` is used for service type `redis-persistent` (which is correct) AND the PHP extension name `redis-persistent` (which is incorrect, it should still just be `redis`)

CleanShot 2023-08-03 at 08 38 03
CleanShot 2023-08-03 at 08 38 18


I think we can add another catch around line 80

```
  {{ if eq $type "redis-persistent" }}
   {{ $extension_name = "redis" }}
  {{ end }}
```

chadwcarlson (changes requested):

chadwcarlson (approved):

PR #3321, opened 2023-08-03
Add guidance in Node.js version update procedure using nvm

### LINKED ISSUE #3051: Manage Node.js Versions for Runtime Containers

https://docs.platform.sh/languages/nodejs/node-version.html

The instructions for using the specified Node version (as set via NVM) in the runtime container (script via `.environment`) assume that the `$PLATFORM_APP_DIR/.nvm` directory exists in runtime. However, the preceding instructions suggest building and activating the Node.js version within `$PLATFORM_CACHE_DIR/.nvm` (cache) and then creating a symlink between the application directory and the cache directory: `ln -s $PLATFORM_CACHE_DIR/.nvm $NVM_DIR `.

The symlink is broken because the cache directory does not exist in the runtime container. As such, the Node.js version is not changed in the runtime container.

### Additional context

Perhaps the content of the `$PLATFORM_CACHE_DIR/.nvm` needs to be copied into a local directory instead of using a symlink.

AnouckColson:
## Why

Closes #3051 

## What's changed

Added a note to nvm procedure.

### DIFF
--- a/docs/src/languages/nodejs/node-version.md
+++ b/docs/src/languages/nodejs/node-version.md
@@ -182,6 +182,18 @@ To use `nvm`, follow these steps:
            ln -s $PLATFORM_CACHE_DIR/.nvm $NVM_DIR
    ```
 
+   {{< note >}}
+
+   Instead of using a symlink between your cache and application directories,
+   you might need to copy the content of `$PLATFORM_CACHE_DIR/.nvm` into `$PLATFORM_APP_DIR/.nvm` manually.
+   To do so, run the following command:
+
+   ```bash
+   rsync -av $PLATFORM_CACHE_DIR/.nvm $PLATFORM_APP_DIR/.nvm
+   ```
+
+   {{< /note >}}
+
 4. Use the cache directory and install based on the variables if not present:
 
    ```yaml {location=".platform.app.yaml"}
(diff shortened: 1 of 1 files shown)


AnouckColson:
@chadwcarlson Tbh my memory of our conversation around the original issue is a bit hazy, so I'm not sure I came up with the most helpful note here. Any help to make it more accurate would be appreciated. Also we need to confirm the actual command with @gilzow before merging.

docs/src/languages/nodejs/node-version.md
```diff
@@ -182,6 +182,18 @@ To use `nvm`, follow these steps:
...
    ```
 
+   {{< note >}}
+
+   Instead of using a symlink between your cache and application directories,
+   you might need to copy the content of `$PLATFORM_CACHE_DIR/.nvm` into `$PLATFORM_APP_DIR/.nvm` manually.
+   To do so, run the following command:
+
+   ```bash
+   rsync -av $PLATFORM_CACHE_DIR/.nvm $PLATFORM_APP_DIR/.nvm
```
chadwcarlson, line 192:
```suggestion
   rsync -av $PLATFORM_CACHE_DIR/.nvm $PLATFORM_APP_DIR
```

chadwcarlson (approved):

PR #3331, opened 2023-08-07
Add .platform/services.yaml example to page on workers

### LINKED ISSUE #3306: show or explain services.yaml config in worker docs

https://docs.platform.sh/create-apps/workers.html#customizing-a-worker 

we don't show the services.yaml contents used on this page, so people can understand that the name given to the rabbitmq service is `rabbitqueue`. This can lead users following this page to believe they **need** to use the name `rabbitqueue`  when working with the rabbitmq services vs that is the name given to the rabbitmq service in services.yaml

### Additional context

we could either list in assumptions, or show the example services.yaml so it's clearer.

AnouckColson:
## Why

Closes #3306


gilzow (approved):
Perfect.

PR #3363, opened 2023-08-17
Update python dependencies. Closes #52

gilzow:
## Why

Closes #52 

## What's changed

Updates python dependencies for the search app


flovntp:
@gilzow as there is only the poetry.lock that is changed, i guess that it's valid :P 
and i check on the env, search is working fine.
approved ;)

flovntp (approved):

PR #3399, opened 2023-09-01
Postgres typo fix

chadwcarlson:
## Why

Closes https://github.com/platformsh/platformsh-docs/issues/3400

Vale error introduced in https://github.com/platformsh/platformsh-docs/pull/3391 for "Postgres/PostgreSQL". This is a fix for that to unblock current PR Vale tests.


gilzow (approved):
👍

PR #3412, opened 2023-09-05
Test The premium-feature add-on banner title should be called "Premium Service" rather than "Beta"

flovntp:
## Why

Closes #{ISSUE_NUMBER}


flovntp:
this was a test

PR #3505, opened 2023-10-12
Add info on env: vars

AnouckColson:
## Why

There was confusion on whether `env:` variables were added to `PLATFORM_VARIABLES`.


AnouckColson:
Addressed in #3518

PR #3543, opened 2023-10-24
Several SPF records

trolologuy:
## Why

The doc page isn't warning about existing records. This might break things. Hence the attempt of a note.


sites/platform/src/development/email.md
```diff
@@ -58,12 +58,16 @@ Changing the setting rebuilds the environment.
...
-Add the following `TXT` record to your domain's DNS records:
+If you don't have an SPF record, add the following `TXT` record to your domain's DNS records:
 
 ```txt
 v=spf1 include:sendgrid.net -all
 ```
 
+Having several, conflicting `TXT` records isn't supported due to [rfc4408 section 3.1.2](https://datatracker.ietf.org/doc/html/rfc4408#section-3.1.2).
+
+If you already have an SPF record, please add sendgrid into your existing record.
```
chadwcarlson, line 69:
```suggestion
If you already have an SPF record, please add SendGrid into your existing record.
```

chadwcarlson (approved):

Issue #3532, opened 2023-10-17
Resources: minimum and defaults


chadwcarlson:
### What needs to be documented?

TBD

### Additional context

See requirements in Context field.

AnouckColson:
Addressed in #3579

PR #3564, opened 2023-11-08
Document default resources

AnouckColson:
## Why

Part of #3532 

## What's changed

Added default resources to https://docs.upsun.com/manage-resources.html and updated each getting started guide to reflect this change upon first deployment.

Note: more doc is coming on minimum resources and resource initialization strategies, but this is the first and most urgent PR as the change has already been released earlier than we anticipated.

### DIFF
--- a/sites/friday/src/manage-resources/_index.md
+++ b/sites/friday/src/manage-resources/_index.md
@@ -39,9 +39,28 @@ When you first deploy your {{% vendor/name %}} project, and whenever you add a n
 | --------------------------- | ----------- |
 | CPU                         | 0.5 |
 | RAM                         | Depends on the [container profile](#advanced-container-profiles). |
-| Disk size                   | 512 MB |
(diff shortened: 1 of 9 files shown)


sites/friday/src/manage-resources/_index.md
```diff
@@ -23,23 +23,25 @@ This flexibility allows you to optimize performance and costs.
...
-use `{{% vendor/cli %}} push` instead of `git push`.
+When you first deploy your {{% vendor/name %}} project, and whenever you add a new application or service,
+{{% vendor/name %}} allocates the following default resources to every container:
 
-{{< /note >}}
+| Resource type               | Amount |
+| --------------------------- | ----------- |
+| CPU                         | 0.5 |
+| RAM                         | Depends on the [container profile](#advanced-container-profiles). |
+| Disk size                   | 512 MB |
```
nicogommen, line 42: For the disk size, we should mention that it's only if the application or service requires a disk.

chadwcarlson (approved):

PR #3587, opened 2023-11-14
Add some Upsun docs for Symfony

fabpot:
This is basically what we currently have for Platform.sh, but updated for Upsun.


fabpot:
I've also removed a bunch of shortcodes that didn't make sense to have (as not reused anywhere else).

AnouckColson (approved):

AnouckColson (approved):

PR #3601, opened 2023-11-27
Typo on post-deploy log type

chadwcarlson:
## Why

While there is a `hooks.post_deploy`,  the _log_ associated with that hook is actually `post-deploy.log`. This updates the table for that spelling distinction.


AnouckColson:
Closes #3602

PR #3623, opened 2023-12-07
Remove obsolete php.ini limitation on DG2 clusters

rhubinak:
The inability to use custom php.ini is valid only for a few clusters (~2%) on which the necessary configuration on our side has not been made (yet). Most clusters (~98%) support setting PHP configuration via this file.


chadwcarlson:
@rhubinak could it be worth it to keep this list somewhere else, referencing it with "for a small set of projects, you can't use `php.ini` yada yada" rather than wholly removing?

PR #3630, opened 2023-12-09
:memo: Precise New Relic configuration for PHP

121593:
New Relic APM can be configured through the use of php-prefixed variables. 

This behavior is not really documented, and deserved to be explained in greater details.


121593:
Q: Are [all configuration options](https://docs.newrelic.com/docs/apm/agents/php-agent/configuration/php-agent-configuration/) usable, or is only a subset available ?

chadwcarlson (approved):

PR #3663, opened 2023-12-22
Address simplified relationship availability with note

chadwcarlson:
## Why

It's been reported at least on one image this is not yet rolled out.


gilzow (approved):
approved pending more information next week

PR #3682, opened 2024-01-10
Build resources setting

### LINKED ISSUE #3675: Document `build_resources` setting in prep for GA

Once Upsun gets to GA stage, we are planning on billing users based on how many build minutes they use for each deployment (**including failed deploys**). 

During each build, a specific build container is started and has default resources (CPU/RAM) allocated to it (completely separate from app/service containers that have their own resources).

We need to document the default resources allocated to that build container, and let users know that they can fiddle with its resources via the `build_resources` setting (API support only atm), while highlighting that it will have an impact on their billing.

See async post for details on the related conf and cURL commands.

### Additional context

_No response_

AnouckColson:
## Why

Closes #3675 

## What's changed

Add a page to document build resources as they will be charged when we get to GA.

### DIFF
--- a/contributing/markup-format.md
+++ b/contributing/markup-format.md
@@ -117,7 +117,7 @@ To add extra information or a warning that stands outside the normal text flow, 
 By default, a title of `Note:` is added.
 
 ```markdown
-{{< note >}}
+{{<  >}}
 
 A short note.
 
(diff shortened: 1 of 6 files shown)


AnouckColson:
Putting this on hold. As I was writing it I figured it didn't make much senses to advertize this feature so much time ahead of GA, which Rémi agreed with. So we'll pick this back up once we get closer to GA and users can actually make use of this new setting.

AnouckColson:
Moved to #3782

PR #3688, opened 2024-01-11
Update Upsun Beta access information

AnouckColson:
## Why

Upsun Beta sign up CTAs still mentioned the closed Beta phase so they needed updating.

## What's changed

Updated the info block + CTA button


Mupsi (approved):
LGTM!

PR #3696, opened 2024-01-15
Update docs on mounts

### LINKED ISSUE #3636: Storage multiapp mounts

https://docs.upsun.com/create-apps/app-reference.html#mounts

Upsun will support shared multiapp mounts.
Users will be able to share `storage` mounts between applications using the `service` key and pointing it to the target app.

### Additional context

_No response_

AnouckColson:
## Why

Closes #3636 

## What's changed

- [x] Upsun & PSH: Document default behaviour for `source_path`
- [x] Upsun only:  Document ability to share a `storage` mount between 2 apps
- [x] Update Network Storage page accordingly


sites/platform/src/create-apps/app-reference.md
```diff
@@ -375,31 +377,61 @@ applications:
...
 The accessibility to the web of a mounted directory depends on the [`web.locations` configuration](#web).
 Files can be all public, all private, or with different rules for different paths and file types.
 
 Note that when you back up an environment,
 the mounts on that environment are backed up too.
 
 Also, mounted directories aren't deleted when they're removed from `{{< vendor/configfile "app" >}}`.
-The files still exist on disk until manually removed (or, for `tmp` mounts, until the app container is moved to another host during a maintenance operation).
+The files still exist on disk until manually removed.
+Or, for `tmp` mounts, **until the app container is moved to another host during a maintenance operation.**
```
lolautruche, line 388:
```suggestion
Or, for `tmp` mounts, **until the app container is moved to another host during a maintenance operation**.
```

sites/platform/src/create-apps/app-reference.md
```diff
@@ -375,31 +377,61 @@ applications:
...
 
 {{% version/only "2" %}}
 
-### Mounts, instances, and Network Storage
+### Data sharing through mounts
 
-`storage` mounts can be shared **between different instances of the same app**,
+By default, [`storage` mounts](#mounts) can be shared **between different instances of the same app**,
```
lolautruche, line 394:
```suggestion
By design, [`storage` mounts](#mounts) are shared **between different instances of the same app**,
```
This is actually the case 100% of the time, so adjusting the sentence :)

sites/platform/src/create-apps/app-reference.md
```diff
@@ -375,31 +377,61 @@ applications:
...
+In a [multi-application context](/create-apps/multi-app/_index.md),
+you can even share a `storage` mount **between different applications** in the same project.
+
+For example, say you have a `backend` app and a `frontend` app.
+You want both apps to share data from the same mount.</br>
+Use the following configuration:
+
+```yaml {configFile="app"}
+# Define a storage mount in your backend app's configuration.
+backend:
```
lolautruche, line 406: I think we should add the `applications` key on top, to avoid confusion

sites/platform/src/create-apps/app-reference.md
```diff
@@ -375,31 +377,61 @@ applications:
...
+    # The directory in the app container where you want your mount to be (mount path).
+    web/uploads:
+      # The mount type.
+      source: storage
+      # Grants access to the backend app's mount. 
+      service: backend
+      # The remote source_path in the backend app's mount to enable data sharing.
+      source_path: backend/uploads
+```
+
```
lolautruche, line 427: I think we should write one or two sentences explaining what the snippet does. It might not be totally self-explanatory

sites/platform/src/create-apps/app-reference.md
```diff
@@ -375,31 +377,61 @@ applications:
...
 
 Also, mounted directories aren't deleted when they're removed from `{{< vendor/configfile "app" >}}`.
-The files still exist on disk until manually removed (or, for `tmp` mounts, until the app container is moved to another host during a maintenance operation).
+The files still exist on disk until manually removed.
+Or, for `tmp` mounts, **until the app container is moved to another host during a maintenance operation.**
 
 {{% version/only "2" %}}
 
-### Mounts, instances, and Network Storage
+### Data sharing through mounts
```
lolautruche, line 392: I'm not a fan of the title. I don't suggest anything here since I don't have a proper idea.

themes/psh-docs/layouts/shortcodes/endpoint-description.md
```diff
@@ -162,10 +162,11 @@ Add the service to your app configuration:
...
 ```
 
-- `<TARGET_PATH>` is where you want your service to be, the path on your app container that has a writable mount.
+- `<DIRECTORY>` is the directory in the app container where you want your mount to be.
 - `<SERVICE_NAME>` is the name you [defined in step 1](#1-configure-the-service).
-- `<SOURCE_PATH>` is the path within the service that the mounts point to.
-  Usually the same as the `<SERVICE_NAME>`.
+- `<SOURCE_PATH>` is the path within the service that the mount points to.</br>
+  Usually the same as the `<SERVICE_NAME>`.</br>
+  If no `<SOURCE_PATH>` is defined or if an empty string is passed, the mount points to the entire `<DIRECTORY>`.
```
lolautruche, line 169: See my previous comment

lolautruche (changes requested):

sites/platform/src/create-apps/app-reference.md
```diff
@@ -311,21 +311,21 @@ See how to [troubleshoot the warning](./troubleshoot-mounts.md#overlapping-folde
...
-| `source_path` | `string`             | Yes      | The subdirectory within the mounted disk (the source) where the mount should point. If an empty string is passed, points to the entire directory. |
-| `service`     | `string`             |          | The name of the [network storage service](../add-services/network-storage.md). |
+| `source_path` | `string`             | No      | The subdirectory within the mounted disk (the source) where the mount should point.</br> If you don't define a `source_path`, or if an empty string is passed, the mount will point to [the entire {{< variable "DIRECTORY" >}}](#mounts). |
```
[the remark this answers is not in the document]
lolautruche: We need to also add an important warning because if a user changes the mount name while `source_path` was ignored, then the mount will be empty since the `source_path` will be changed accordingly with the new mount name, thus pointing a new empty shared directory.
The data won't be lost though. To ensure continuity, the user must specify the `source_path` in this case.

Example: I define a mount on `/var/storage`, then rename it to `/foo`. In order to still get access to my files, I need to define the `source_path` to `var/storage`.

Before:
```yaml
    mounts:
      "/var/storage":
        source: storage
```

After:
```yaml
    mounts:
      "/foo":
        source: storage
        source_path: var/storage
```

Note that the `source_path` is identical to the previous mount name, without the leading `/` (and potentially the trailing one if it's present)

sites/platform/src/create-apps/app-reference.md
```diff
@@ -362,7 +398,7 @@ applications:
                 source_path: uploads
             '/.tmp_platformsh':
                 source: tmp
-                source_path: files/tmp_platformsh
+                source_path: files/.tmp_platformsh
```
AnouckColson, line 401:
```suggestion
                source_path: files/tmp_platformsh
```

sites/platform/src/create-apps/app-reference.md
```diff
@@ -273,16 +273,32 @@ For more information, see how to [manage resources](/manage-resources.md).
 
 ## Mounts
 
-Mounts define directories that are writable after the build is complete.
-They aren't available during the build.
+After your app is built, its file system is read-only.
+You need to use Git as the principal way to make changes to your app.
+However, you may not want to rebuild and redeploy your app every time you need to make a small change to it.
```
lolautruche, line 278: It's not actually about making changes to the application. This must go through Git.
It's about writing files generated by the application (uploaded files, writing cache, writing logs, etc)

sites/platform/src/create-apps/app-reference.md
```diff
@@ -273,16 +273,32 @@ For more information, see how to [manage resources](/manage-resources.md).
...
+
+For enhanced flexibility, {{% vendor/name %}} allows you to define and use writable directories called mounts.
```
lolautruche, line 280:
```suggestion
For enhanced flexibility, {{% vendor/name %}} allows you to define and use writable directories called "mounts".
```

sites/platform/src/create-apps/app-reference.md
```diff
@@ -375,31 +412,122 @@ applications:
...
+ To ensure continuity, you need to explicitly define the `source_path` as the previous name of the mount, without leading or trailing slashes:
+ 
+ ```yaml {configFile="app"}
+mounts:
+    '/my/uploads/':
+        source: local
+        source_path: web/uploads
+```
+ 
+The `/my/uploads/` mount will point to the original `/web/uploads/` directory, maintaining access to all your existing files in that directory.
```
lolautruche, line 438: Is this chunk for PSH only, or also for Upsun?
Asking because I don't see the `<!-- Platform.sh -->` tag, and it mentions `local` mount which doesn't exist on Upsun.

sites/platform/src/create-apps/app-reference.md
```diff
@@ -375,31 +412,122 @@ applications:
...
 
 {{% version/only "2" %}}
 
-### Mounts, instances, and Network Storage
+### Share a mount between several apps
 
-`storage` mounts can be shared **between different instances of the same app**,
+By design, [`storage` mounts](#mounts) are shared **between different instances of the same app**,
```
lolautruche, line 444: Same question as above, as `storage` mounts are only available on Upsun (and thus so are shared mounts using the mechanism described here)

lolautruche (changes requested):

sites/platform/src/create-apps/app-reference.md
AnouckColson: My bad, changed it to the `tmp` type since it's supported by both products.

sites/platform/src/create-apps/app-reference.md
AnouckColson: This one has a {{% version/only "2" %}} shortcode applied to it so it's only shown in the Upsun output (version 1 = PSH, version 2 = Upsun). But yeah it can be confusing, and those two pieces of feedback from you only confirm to me that the separation must be clearer, so there can be no confusion and we get what's intended in prod.

sites/platform/src/create-apps/app-reference.md
```diff
@@ -273,16 +273,34 @@ For more information, see how to [manage resources](/manage-resources.md).
...
 ## Mounts
 
-Mounts define directories that are writable after the build is complete.
-They aren't available during the build.
+After your app is built, its file system is read-only.
+To make changes to your app's code, you need to use Git.
+
+For enhanced flexibility, {{% vendor/name %}} allows you to define and use writable directories called "mounts".
+Mounts give you write access to files generated by your app (such as cache and log files)
+and [uploaded files](/development/file-transfer.md#transfer-files-using-the-cli) without going through Git.
```
AnouckColson, line 281:
```suggestion
and uploaded files without going through Git.
```

sites/platform/src/create-apps/app-reference.md
```diff
@@ -273,16 +273,34 @@ For more information, see how to [manage resources](/manage-resources.md).
...
-They aren't available during the build.
+After your app is built, its file system is read-only.
+To make changes to your app's code, you need to use Git.
+
+For enhanced flexibility, {{% vendor/name %}} allows you to define and use writable directories called "mounts".
+Mounts give you write access to files generated by your app (such as cache and log files)
+and uploaded files without going through Git.
+
+When you define a mount, you are mounting an external directory to your app container,
+much like you would plug a hard drive into your computer to [transfer data](/development/file-transfer.md#transfer-files-using-the-cli).
```
AnouckColson, line 284:
```suggestion
much like you would plug a hard drive into your computer to transfer data.
```

lolautruche (approved):

chadwcarlson (approved):

PR #3701, opened 2024-01-22
Add ch-1 region to regions.yaml

AnouckColson:
## Why

Closes #{ISSUE_NUMBER}

### DIFF
--- a/shared/data/regions.yaml
+++ b/shared/data/regions.yaml
@@ -1,4 +1,18 @@
 europe:
+    ch-1:
+        name: Switzerland (ch-1)
+        label: ch-1
+        provider: GCP
+        available: true
+        outbound_ips: 
+            - "34.65.236.213"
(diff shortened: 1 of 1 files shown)


shared/data/regions.yaml
```diff
@@ -1,4 +1,18 @@
 europe:
+    ch-1:
+        name: Switzerland (ch-1)
+        label: ch-1
+        provider: GCP
```
remilejeune2, line 5: Google Cloud Platform

remilejeune2 (changes requested):

shared/data/regions.yaml
AnouckColson, line 5:
```suggestion
        provider: Google Cloud Platform
```

remilejeune2 (approved):

PR #3734, opened 2024-02-07
Upsun migrate security

gilzow:
## Why
Prepares security section for repository separation

## What's changed
* Copies the following from the platform docs src to the upsun docs src:
  * _index.md
  * backups.md
  * data-retention.md
* Updates the src and the copies above to remove the toggled sections between the vendors
* Updates Upsun config to exclude the entire security section when sourcing from platform/src


gilzow:
visually inspected the security sections of the PR version for both platform and upsun against their current production versions and everything appears to be identical.

chadwcarlson (approved):

PR #3737, opened 2024-02-08
Update Manage resources section

AnouckColson:
## Why

It is now possible to set a resource initialization strategy via the CLI or Console when branching, merging, activating an environment, and restoring a backup.

It is also possible to run a CLI command to sync resources (on top of code and/or data) when syncing 2 environments.

## What's changed

Full rework of the page to reorganize content and add sections where needed.

Closes https://github.com/platformsh/platformsh-docs/issues/3825


sites/friday/src/manage-resources/resource-init.md
```diff
@@ -106,79 +172,223 @@ platform integration:add \
...
+2. Click {{< icon merge >}} **Merge**.
+3. Select an initialization strategy from the proposed list.
+5. Click **Merge**.
+
+{{< /codetabs >}}
+
+### Environment activation
+
+When you activate an environment, {{% vendor/name %}} uses the same resource allocation as the parent environment.
+If there is no parent environment, the [`default` resource initialization strategy](#default-resources) applies.
```
AnouckColson, line 282: @nicogommen Is that correct?

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -106,79 +172,223 @@ platform integration:add \
...
+```bash {location="Terminal"}
+environment:activate --resources-init=minimum
+```
 
-### Environment sync
+### Backup restoration
+
+When you [restore a backup](/environments/restore.md),
+{{% vendor/name %}} uses the same resource allocation as the parent environment.
+If there is no parent environment, the [`default` resource initialization strategy](#default-resources) applies.
```
AnouckColson, line 308: @nicogommen Same question?

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -106,79 +172,223 @@ platform integration:add \
...
+## Environment sync
 
 When you [sync an environment](/glossary.md#sync),
-the target environment's disk size may be smaller that the source environment's disk size.
+changes in Production are copied to your Staging environment.
+You can synchronize only the code, only the data (databases, files), or both.
+
+As your Staging environment's disk size may be smaller than your Production environment's disk size,
+you can also sync resources.
+This ensures that there is enough disk on the Staging environment to successfully complete the sync.
```
AnouckColson, line 360: @nicogommen I'm unsure if I should talk about Production vs. Staging, or use more generic temrinology => source environment vs. target environment. WDYT?

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -106,79 +172,223 @@ platform integration:add \
...
+title=In the Console
++++
+
+1. Navigate to the environment where you want to restore a backup.
+2. Click **Backups**.
+3. Select the backup you want to restore and click **More** {{< icon more >}} next to it.
+4. Click **Restore**.
+5. Read through the consequences.
+6. Select an initialization strategy from the proposed list.
+7. Click **Backup**.
```
AnouckColson, line 348: @nicogommen Unsure about the actual step by step. Needs testing.

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -13,33 +13,102 @@ keywords:
...
+
+## Set a resource initialization strategy
+
+{{% vendor/name %}} provides the following resource initialization strategies:
 
 | Strategy | Description |
 | ---------| ----------- |
-| `default`  | Initializes the new containers using the [{{% vendor/name %}} default resources](/manage-resources/resource-init.md).</br>This is the strategy that applies when you first deploy your project or new container, unless you explicitly set another strategy.  |
+| `default`  | Initializes the new containers using the [{{% vendor/name %}} default resources](#default-resources).</br>This strategy applies when you first deploy your project (or new container) unless you explicitly set another strategy, or allocate resources manually via the `resources:set` CLI command.  |
 | `manual`   | With this strategy, the first deployment fails and you need to configure resources manually through [the Console](/manage-resources/adjust-resources.md), or using `resources:set` in the CLI. |
```
nicogommen, line 52: For the manual strategy, I think it would be nice to add also something explaining why this strategy could be a good option for the user, depending on their use case.
Because with the current description, with the failure etc., it doesn't sound very appealing 😅
The benefit of this strategy is that you can directly deploy with the desired resources. Whereas with the other strategies, if it's not exactly the resources that you want/need, you need to do a second deployment to change them. Which means that your env will run for a short moment with not exactly the wished resources. And the 2 deployments both involve some downtime (until we have real zero downtime deployments).

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -106,79 +172,223 @@ platform integration:add \
...
++++
+title=Using the CLI
++++
 
-When you [branch an environment](/glossary.md#branch) to create a new child environment,
-the child environment inherits all the resources from its parent.
+Run the following command:
+
+```bash {location="Terminal"}
+environment:branch --resources-init={{< variable "INITIALIZATION_STRATEGY" >}}
```
nicogommen, line 211: To avoid confusion, I think we should add the `upsun` command at the beginning.

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -106,79 +172,223 @@ platform integration:add \
...
+```
+
+For example, to use the `minimum` resource initialization strategy, run the following command:
+
+```bash {location="Terminal"}
+environment:branch --resources-init=minimum
```
nicogommen, line 217: To avoid confusion, I think we should add the `upsun` command at the beginning.

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -106,79 +172,223 @@ platform integration:add \
...
+{{< codetabs >}}
+
++++
+title=Using the CLI
++++
+
+Run the following command:
+
+```bash {location="Terminal"}
+environment:merge --resources-init={{< variable "INITIALIZATION_STRATEGY" >}}
```
nicogommen, line 257: To avoid confusion, I think we should add the `upsun` command at the beginning.

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -106,79 +172,223 @@ platform integration:add \
...
+```
+
+For example, to use the `manual` resource initialization strategy, run the following command:
+
+```bash {location="Terminal"}
+environment:merge --resources-init=manual
```
nicogommen, line 263: To avoid confusion, I think we should add the `upsun` command at the beginning.

sites/friday/src/manage-resources/resource-init.md
nicogommen: Yes, that's correct.

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -106,79 +172,223 @@ platform integration:add \
...
+{{% note theme="info" title="Prerequisite" %}}
+
+Make sure `sizing_api_enabled` is set to `true` in your `{{% vendor/configfile "app" %}}` file.
+
+{{% /note %}}
+
+Run the following command:
+
+```bash {location="Terminal"}
+environment:activate --resources-init={{< variable "INITIALIZATION_STRATEGY" >}}
```
nicogommen, line 295: To avoid confusion, I think we should add the `upsun` command at the beginning.

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -106,79 +172,223 @@ platform integration:add \
...
+```
+
+For example, to use the `minimum` resource initialization strategy, run the following command:
+
+```bash {location="Terminal"}
+environment:activate --resources-init=minimum
```
nicogommen, line 301: To avoid confusion, I think we should add the `upsun` command at the beginning.

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -106,79 +172,223 @@ platform integration:add \
...
+{{< codetabs >}}
+
++++
+title=Using the CLI
++++
+
+Run the following command:
+
+```bash {location="Terminal"}
+backup:restore --resources-init={{< variable "INITIALIZATION_STRATEGY" >}}
```
nicogommen, line 327: To avoid confusion, I think we should add the `upsun` command at the beginning.

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -106,79 +172,223 @@ platform integration:add \
...
+```
+
+For example, to use the `manual` resource initialization strategy, run the following command:
+
+```bash {location="Terminal"}
+backup:restore --resources-init=manual
```
nicogommen, line 333: To avoid confusion, I think we should add the `upsun` command at the beginning.

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -106,79 +172,223 @@ platform integration:add \
...
+{{< codetabs >}}
+
++++
+title=Using the CLI
++++
+
+Run the following command:
+
+```bash {location="Terminal"}
+{{% vendor/name %}}  sync code data resources
```
nicogommen, line 379: Here it gives `Upsun  `, with an uppercase U and 2 spaces after.
It should be `upsun`.

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -106,79 +172,223 @@ platform integration:add \
...
-| Redis persistent        | 0.1  | 352 MB | 256 MB  | 
-| Ruby                    | 0.1  | 64 MB  | 0 MB    |
-| Rust                    | 0.1  | 64 MB  | 0 MB    |
-| Solr                    | 0.1  | 448 MB | 256 MB  | 
-| Varnish                 | 0.1  | 448 MB | None    |
-| Vault KMS               | 0.1  | 448 MB | 256 MB  |
+{{% note theme="info" title="Prerequisite" %}}
 
-\* The disk size is set to `None` when the container never uses a disk, and to `0 MB` when the container doesn't require a disk but can use one.
+Make sure `sizing_api_enabled` is set to `true` in your `{{% vendor/configfile "app" %}}` file.
```
nicogommen, line 198: We don't need to mention this prerequisite to the user.
It's more something internal that the user doesn't need to know.

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -106,79 +172,223 @@ platform integration:add \
...
-you need to [allocate resources](/manage-resources/adjust-resources.md) to each of these newly added instances.
+By default, the resources from the child environment are also merged into the parent.
+
+However, you can set another resource initialization strategy.
+
+{{% note theme="info" title="Prerequisite" %}}
+
+Make sure `sizing_api_enabled` is set to `true` in your `{{% vendor/configfile "app" %}}` file.
+
+{{% /note %}}
```
nicogommen, line 246: We don't need to mention this prerequisite to the user.
It's more something internal that the user doesn't need to know.

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -106,79 +172,223 @@ platform integration:add \
...
+When you activate an environment, {{% vendor/name %}} uses the same resource allocation as the parent environment.
+If there is no parent environment, the [`default` resource initialization strategy](#default-resources) applies.
+
+You can also set another resource initialization strategy using the CLI.
+
+{{% note theme="info" title="Prerequisite" %}}
+
+Make sure `sizing_api_enabled` is set to `true` in your `{{% vendor/configfile "app" %}}` file.
+
+{{% /note %}}
```
nicogommen, line 290: We don't need to mention this prerequisite to the user.
It's more something internal that the user doesn't need to know.

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -106,79 +172,223 @@ platform integration:add \
...
+{{% vendor/name %}} uses the same resource allocation as the parent environment.
+If there is no parent environment, the [`default` resource initialization strategy](#default-resources) applies.
+
+You can also set another resource initialization strategy.
+
+{{% note theme="info" title="Prerequisite" %}}
+
+Make sure `sizing_api_enabled` is set to `true` in your `{{% vendor/configfile "app" %}}` file.
+
+{{% /note %}}
```
nicogommen, line 316: We don't need to mention this prerequisite to the user.
It's more something internal that the user doesn't need to know.

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -106,79 +172,223 @@ platform integration:add \
...
+you can also sync resources.
+This ensures that there is enough disk on the Staging environment to successfully complete the sync.
+
+To sync resources, follow these instructions.
+
+{{% note theme="info" title="Prerequisite" %}}
+
+Make sure `sizing_api_enabled` is set to `true` in your `{{% vendor/configfile "app" %}}` file.
+
+{{% /note %}}
```
nicogommen, line 368: We don't need to mention this prerequisite to the user.
It's more something internal that the user doesn't need to know.

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -106,79 +172,223 @@ platform integration:add \
...
+
+{{< /codetabs >}}
 
 ### Environment merge
 
 When you [merge](/glossary.md#merge) a child environment into a parent environment,
 any app or service on the child environment is merged into the parent.
-For the merge and deployment to be complete,
-you need to [allocate resources](/manage-resources/adjust-resources.md) to each of these newly added instances.
+By default, the resources from the child environment are also merged into the parent.
```
nicogommen, line 238: It's actually not exactly like that.
The resource init strategy is only used for new apps/services/workers.

For instance if you have a php app + a mariadb service, and you would like to add redis.
You will create a new branch, add redis, and once ready, merge this branch to the parent to deploy the new service to the parent.
It's only the new redis service that will be initialized with the chosen strategy. The php app and the mariadb service won't be touched, because we don't want to change the resources of these already running service, whereas redis is a new service that is being created.

We are actually working on a change in the Console, to show the resources init option only if we detect that a new container will be deployed with the merge.
So the modal won't show the option and stay very simple (as it is today in production), and the option will only be displayed if it is relevant, because we detected that a new container will be created with the merge.

Let me know if you have questions 🙂

sites/friday/src/manage-resources/resource-init.md
nicogommen: At the moment, it is correct, yes.
But, we are maybe changing this... but it's not sure yet that it will make it for the release... 😅  
The change would be that we are saving the resources configured at the moment the backup is created. So when a backup is restored, we restore the environment with the exact resources as when the backup was taken.
This would be like a new strategy ("backup resources", or something like that), and it would be the default behaviour when restoring a backup.
Then you can still choose another strategy, as you described it.
So maybe we can put this section on hold for a short time. We should know very soon if this will make it in Git 26 or not.
I will let you know. 
Sorry about this change.

sites/friday/src/manage-resources/resource-init.md
nicogommen: I think this is correct 👍

sites/friday/src/manage-resources/resource-init.md
nicogommen: Yes, I think a generic terminology would be more appropriate here.
We could use parent and child (it synchronizes from the parent to the child env).

And it is actually not exactly working like described here.

You can synchronize only the code, only the data, or only the resources, or any combination or these.
The resources sync will actually synchronize all the resources from the parent to the child (CPU&RAM, instances and disk).

Then, if you sync the data, even if you didn't choose to sync the resources, we will do what we described before:

> In this case, to ensure that the synchronization succeeds, Upsun automatically adjusts the target environment’s disk size so it matches the source environment’s disk size.

Let me know if it's not clear 🙂

sites/friday/src/manage-resources/resource-init.md
AnouckColson: @nicogommen Do you think this change in the Console might be ready for GA?

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -115,79 +171,202 @@ platform integration:add \
...
 
-```bash
+```bash {location="Terminal"}
 upsun integration:update --resources-init=minimum
 ```
 
 {{< /codetabs >}}
 
-### Minimum resources
+### Environment branch
```
nicogommen, line 189: Something new with the branch operation => even when using the another strategy than parent (such as `default` or `minimum`, since we removed `manual`), we are initializing the disk size with the same disk as the parent.
This is for the same reasons as when syncing the data from parent to the child. The child needs to have enough disk to clone the data from the parent. So even if the user choose the default or minimum strategy, the disk size will be the one from the parent.

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -13,42 +13,101 @@ keywords:
...
-| `manual`   | With this strategy, the first deployment fails and you need to configure resources manually through [the Console](/manage-resources/adjust-resources.md), or using `resources:set` in the CLI. |
-| `minimum`  | Initializes the new containers using the [{{% vendor/name %}} minimum resources](#minimum-resources). |
+| `default`  | Initializes the new containers using the [{{% vendor/name %}} default resources](#default-resources).</br>This strategy applies when you first deploy your project (or new container) unless you explicitly set another strategy, or allocate resources manually via the `resources:set` CLI command.  |
+| `manual`   | With this strategy, the first deployment fails and you need to configure resources manually through [the Console](/manage-resources/adjust-resources.md), or using `resources:set` in the CLI.</br></br> This strategy allows you to set the exact resources you want, with a single deployment. Other strategies may require fine-tuning, and therefore generate a second deployment. In this case, your environment would run for a short time with unwanted resources, and both deployments would generate downtime.|
```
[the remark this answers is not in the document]
tejapsh: We don't support `manual` in case of backup restore either. It suffers from the same problem as having manual for branch.

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -115,79 +171,202 @@ platform integration:add \
...
++++
+
+1. Navigate to the environment you want to merge into its parent.
+2. Click {{< icon merge >}} **Merge**.
+3. Select an initialization strategy from the proposed list.
+5. Click **Merge**.
+
+{{< /codetabs >}}
+
+### Environment activation
```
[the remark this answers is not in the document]
tejapsh: I am not sure if we are implying this elsewhere or if we even need to say it, but, if the deployment generated from a `resources:set` fails, in some cases, the resources configuration will not be saved and therefore, the customer will have to set the resources again.

This usually happens if either the yaml validation or the deployment validation fails.

sites/friday/src/manage-resources/resource-init.md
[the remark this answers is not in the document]
nicogommen: Manual has now been removed from env activate too.
I updated my comment above accordingly.

sites/friday/src/manage-resources/resource-init.md
AnouckColson: Note added at the end of the section.

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -115,79 +171,202 @@ platform integration:add \
...
+
+<--->
+
++++
+title=In the Console
++++
+
+1. Navigate to the environment you want to merge into its parent.
+2. Click {{< icon merge >}} **Merge**.
+3. Select an initialization strategy from the proposed list.
```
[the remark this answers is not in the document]
AnouckColson: Note added at step 3.

sites/friday/src/manage-resources/resource-init.md
AnouckColson: Backup restoration section updated.

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -115,79 +171,202 @@ platform integration:add \
...
 When you [merge](/glossary.md#merge) a child environment into a parent environment,
-any app or service on the child environment is merged into the parent.
-For the merge and deployment to be complete,
-you need to [allocate resources](/manage-resources/adjust-resources.md) to each of these newly added instances.
+all apps and services are copied from the child to the parent.
+
+Such merges don't affect resources allocated to apps and services that already existed on the parent environment.
+
+By default, when a **new** app or service is copied to the parent environment,
+it is given the same resources as on the child environment.
```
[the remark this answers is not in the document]
AnouckColson: Updated to apply your suggestions.

sites/friday/src/manage-resources/resource-init.md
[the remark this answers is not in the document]
AnouckColson: Added a table at the end of the **Specify a resource initialization strategy** section + an info block at the beginning of every sub-section in case users go straight to one of them without reading the intro.

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -115,79 +171,202 @@ platform integration:add \
...
+For example, to use the `manual` resource initialization strategy, run the following command:
+
+```bash {location="Terminal"}
+upsun backup:restore --resources-init=manual
+```
+
+<--->
+
++++
+title=In the Console
```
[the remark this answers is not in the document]
AnouckColson: **Backup and restoration** section updated.

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -115,79 +171,202 @@ platform integration:add \
...
+You can sync:
+
+- Only the code
+- Only the data (databases, files)
+- Only the resources (CPU and RAM, instances, and disk space)
+- Any combination of the three (code and data, data and resources, etc.)
+
+{{% note %}}
+
+When you only sync the data, {{% vendor/name %}} checks if the child environment's disk size is at least equivalent to the parent's.</br>
```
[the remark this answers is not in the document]
AnouckColson: Done!

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -115,79 +171,202 @@ platform integration:add \
...
+
+{{% /note %}}
+
+{{< codetabs >}}
+
++++
+title=Using the CLI
++++
+
+Run the following command:
```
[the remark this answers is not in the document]
AnouckColson: Updated the section with more examples of commands to give a better idea of what can be done/how.

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -13,42 +13,101 @@ keywords:
   - "scaling"
 ---
 
-When you first deploy your {{% vendor/name %}} project, and whenever you add a new app or service,
-{{% vendor/name %}} allocates the following default resources to every container:
+When you first deploy your project or add a new container to it,
+{{% vendor/name %}} uses [default resources](#default-resources)
+unless you set another [resource initialization strategy](#set-a-resource-initialization-strategy).
```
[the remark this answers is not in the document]
AnouckColson: Done!

sites/friday/src/manage-resources/resource-init.md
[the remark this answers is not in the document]
AnouckColson: I've added notes about this to the **Resource configuration** page.

sites/friday/src/manage-resources/resource-init.md
nicogommen: Thanks, the note is great.
We could maybe just put it below the tabs, because at the moment, it only appears in the `Using the CLI tab`, and not in the `In the Console` tab.

AnouckColson: You're absolutely right, I've moved the note up.

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -118,76 +184,288 @@ platform integration:add \
...
+
+When you [merge](/glossary.md#merge) a child environment into its parent environment,
+any apps and services you created on the child are merged and therefore created on the parent.
+
+When this happens, any new app or service container created on the parent environment is granted the same resources as on the child environment.
+However, you can specify a different resource initialization strategy.
+
+{{% note %}}
+
+Any other container already running on the parent environment keeps its resources.
```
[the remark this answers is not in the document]
nicogommen: Yes, that's correct.
The resources init strategy is only applied to initialize new containers. Apps/services already running on the target env are left untouched, in terms of resources.

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -13,42 +13,111 @@ keywords:
...
+
+{{% vendor/name %}} provides the following resource initialization strategies:
 
 | Strategy | Description |
 | ---------| ----------- |
-| `default`  | Initializes the new containers using the [{{% vendor/name %}} default resources](/manage-resources/resource-init.md). |
-| `manual`   | With this strategy, the first deployment fails and you need to configure resources manually through [the Console](/manage-resources/adjust-resources.md), or using `resources:set` in the CLI. |
-| `minimum`  | Initializes the new containers using the [{{% vendor/name %}} minimum resources](#minimum-resources). |
-| `parent`   | Initializes the new containers using the same resources as the parent environment.</br>If there is no parent environment, or if the container doesn't already exist on the parent, the `default` strategy applies instead. |
+| `default`  | Initializes new containers using the [{{% vendor/name %}} default resources](#default-resources).</br>This strategy applies when you first deploy your project (or new container) unless you explicitly specify a different strategy, or allocate resources manually via the `resources:set` CLI command.  |
```
gilzow, line 50: There is no anchor on this page with an id of `default-resources` so Hugo is interpreting this as an empty `href`

gilzow (changes requested):

sites/friday/src/manage-resources/resource-init.md
```diff
@@ -13,42 +13,111 @@ keywords:
...
+| `manual`   | With this strategy, the first deployment fails and you need to configure resources manually through [the Console](https://console.upsun.com/), or using `resources:set` in the CLI.</br></br> This strategy allows you to set the exact resources you want, with a single deployment. Other strategies may require fine-tuning, and therefore generate a second deployment. In this case, your environment would run for a short time with unwanted resources, and both deployments would generate downtime.|
+| `minimum`  | Initializes new containers using the {{% vendor/name %}} minimum resources (see below). |
+| `parent`   | Initializes new containers using the same resources as on the parent environment.</br>If there is no parent environment, or if the container doesn't already exist on the parent, the `default` strategy applies instead. |
+| `child`    | Initializes new containers using the same resources as on the child environment. |
```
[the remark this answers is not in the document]
AnouckColson: We decided to remove such information from this table to make it a bit lighter/easier to read, and introduced a second table a bit further down to show exactly which strategies are available when + info notes in every action section, in case readers jump straight to them without reading the intro.

Mupsi (approved):

PR #3747, opened 2024-02-09
Rework structure of Redirects page

### LINKED ISSUE #3700: restucture layout on partial redirects page

https://docs.platform.sh/define-routes/redirects.html

The current structure makes it appear as if `prefix` is a subpart of `regexp`

image
We have a code example, and a paragraph discussing `regexp` and then a sub bullet point which makes it appear that `prefix` is a part, or related to `regexp`. 

`prefix` and `append_suffix` are incompatible with `regexp`. It's an either or: if you use `regexp` you can't use either `prefix` or `append_suffix`. If you use `prefix` or `append_suffix` you can't also use `regexp`. 

### Additional context

Maybe we should add h3 headings beneath the `Partial redirects` h2 for each of the areas?
### Redirects using prefix/append_suffix

And then include information on how these two are equivalent to regex but a convenience. And then possibly the warning that they cant be used together.

AnouckColson:
## Why

Closes #3700


sites/platform/src/define-routes/redirects.md
```diff
@@ -141,11 +169,13 @@ Each rule under `paths` is defined by its key describing the expression to match
    ```
 {{< /version/specific >}}
 
-   with `prefix` set to `true`, `/from` redirects to `/to` and `/from/another/path` redirects to `/to/another/path`.
-   If `prefix` is set to `false` then `/from` triggers a redirect, but `/from/another/path` doesn't.
+With `prefix` set to `true`:
+- `/from` redirects to `/to`
+- `/from/another/path` redirects to `/to/another/path`
+
+If `prefix` were set to `false`, then `/from` would a redirect, but `/from/another/path` wouldn't.
```
chadwcarlson, line 176:
```suggestion
If `prefix` were set to `false`, then `/from` would be a redirect, but `/from/another/path` wouldn't.
```

sites/platform/src/define-routes/redirects.md
```diff
@@ -107,12 +121,26 @@ Each rule under `paths` is defined by its key describing the expression to match
    ```
 {{< /version/specific >}}
 
-   Note that special arguments in the `to` statement are also valid when `regexp` is set to `true`:
-    * `$is_args` evaluates to `?` or empty string
-    * `$args` evaluates to the full query string if any
-    * `$arg_foo` evaluates to the value of the query parameter `foo`
-    * `$uri` evaluates to the full URI of the request.
- * `prefix`: optional, specifies whether both the path and all its children or just the path itself should be redirected. Defaults to `true`, but not supported if `regexp` is `true`. For example,
+Note that special arguments in the `to` statement are also valid when `regexp` is set to `true`:
```
gilzow, line 124: "also valid" seems like they're valid in another context as well?

Maybe 
> the following special arguments in the `to` statement are available when `regexp` is set to `true`:

sites/platform/src/define-routes/redirects.md
```diff
@@ -141,11 +169,13 @@ Each rule under `paths` is defined by its key describing the expression to match
    ```
 {{< /version/specific >}}
 
-   with `prefix` set to `true`, `/from` redirects to `/to` and `/from/another/path` redirects to `/to/another/path`.
-   If `prefix` is set to `false` then `/from` triggers a redirect, but `/from/another/path` doesn't.
+With `prefix` set to `true`:
```
gilzow, line 172: see my comment in the main thread of comments

gilzow:
alternatively, or in combination with, we could add the regex explanations dmanz used.

gilzow (approved):

PR #3761, opened 2024-02-14
[Upsun migration] add-services section

gilzow:
For the`/add-services` section:

- adds src files from platformsh to upsun


gilzow:
False positives, platform side:
* `/add-services/mysql.html` <-- correct addition to side nav causing false positive. THAT SAID, the header that was picked up is fairly long and causing issues in the nav. This one might need more work.


Valid issues, platform side:
* `/add-services/mysql/mysql-replication.html` <-- there seems to be an issue with the `<snippet` shortcode that is embedded in the code block at the beginning of the page. Something is preventing hugo from seeing the block correctly? it's adding the wrong class to the elements inside the codeblock.
* `/add-services/opensearch.html` <-- same exact thing is happening with the `<snippet` shortcode at the "[use in app](https://upsun-migrate-add-4vcv5sq-ucq44jg6ofare.eu-5.platformsh.site/add-services/opensearch.html#use-in-app)" section. Happens again later in the authentication section (both blocks), and the plugins section.
* `/add-services/postgres.html` <-- same exact thing is happening with the `<snippet` shortcode at the "[multiple databases](https://upsun-migrate-add-4vcv5sq-ucq44jg6ofare.eu-5.platformsh.site/add-services/postgresql.html#multiple-databases)" section (all code blocks), and Extensions section. 
* `/add-services/rabbitmq.html` <-- same issue as the rest, but oddly, somehow *only* in the phone layout?
* `/add-services/redis.html` <-- same issue as the rest, all the way at the bottom of the page, last two blocks
* `/add-services/solr.html` <-- same issue, starting at solr 4 section to the end of the page for all blocks

chadwcarlson (approved):

chadwcarlson:
> False positives, platform side:
> 
> ...

CONFIRMED

PR #3767, opened 2024-02-14
Vendor: separate observability

chadwcarlson:
## Why

Separates observability section.


chadwcarlson:
Vrt results:

- Platform.sh: all false positives, either from missing image in search bar or cookie consent box.
- Upsun: Small errors on 404 pages and 0 search results text. I think we can ignore.

gilzow (approved):

PR #3770, opened 2024-02-14
Vendor: domains section

chadwcarlson:
## Why

Separates domains section.


chadwcarlson:
VRT results:
- Platform.sh: false positive wrt headers now rendered that didn't previously (because they were embedded in shortcode)
- Upsun: `No issues detected with upsun.`

gilzow (approved):

PR #3778, opened 2024-02-15
Vendor: separate create-apps section

gilzow:
For the `/create-apps` section:

- copies src files from platformsh to upsun
- updates upsun config to no longer source files for this section from platform


gilzow:
False-positives, platform side:
* `/create-apps/hooks/` <-- correct addition to side nav causing false positive. Also, sub sections of ordered lists are now properly nested beneath their sections also causing a false positive
* `/create-apps/app-reference.html` <-- correct addition to side nav causing false positive.
* `/create-apps/troubleshoot-disks.html` <-- correct addition to side nav causing false positive.

gilzow:
False-positives, upsun side:

- `/create-apps/hooks/` <--Sub sections of ordered lists are now properly nested beneath their sections causing a false positive
-

chadwcarlson:
vrt confirmed

chadwcarlson (approved):

PR #3816, opened 2024-02-29
[Django + Flask] Getting Started Frameworks section for Upsun

gilzow:
This is a WIP/demo PR 

Per conversation with Chad, demo PR of a reworked frameworks getting started section. 

The idea is that the Frameworks section should include the required steps for a framework that are not covered by the general Getting Started guide. This way, if a user navigates to this section (e.g. "Deploying Drupal on Upsun") they should have the steps/items they **need** to accomplish the goal of deploying on Upsun for that framework without being taken away from the docs to a blog post. The blog posts then can focus on more niche items that are not appropriate for the documentation (e.g. setting up celerey on Upsun for Django). 

What I don't like so far in this demo: we need a better, _consistent_ method for organizing remaining steps and structure on how much additional information we include. In this demo, there are six remaining steps. I chose to organize by _where_ the steps need to occur (`.environment` vs `.upsun/config.yaml`) but perhaps it needs to be an ordered list, one for each step?

For the first step (changes to`.environment`), I included information about why the environment variables are needed. Perhaps we don't need to include this information and we can instead assume that someone familiar with a framework already knows _what_ the items are and simply needs to know _where_ to include them?


sites/friday/src/get-started/stacks/django.md
```diff
@@ -9,21 +9,113 @@ description: |
...
+### Deploy Hook
+Just below that, find the `deploy` section and change the following lines from:
+```yaml
+      deploy: |
+        set -eux
+        # echo 'Put your deploy command here'
+```
+to:
+```yaml
+            deploy: |
```
flovntp, line 92: indentation changed?

flovntp: maybe you need to display all top level keys above, like; 
```
applications:
    app:
        hooks:
            deploy: |
```

flovntp: and same for all other YAML snippets?

sites/friday/src/get-started/stacks/django.md
```diff
@@ -9,21 +9,113 @@ description: |
...
+```yaml
+    # mounts:
+    #   "/.cache": # Represents the path in the app.
+    #     source: "local" # "local" sources are unique to the app, while "service" sources can be shared among apps.
+    #     source_path: "cache" # The subdirectory within the mounted disk (the source) where the mount should point.
+```
+to
+
+```yaml
+        mounts:
```
flovntp, line 111: indentation is different...

flovntp:
so, if it's an amendment to @chadwcarlson's Django blogpost, why don't you just change the snippets in the blogpost itself?
or, maybe add a note in the blogpost with the link to this page?

sites/friday/src/get-started/stacks/django.md
gilzow: that's the challenging part with yaml. do our code samples include the correct indentation according to where the keys we're changing are located so if someone copies and pastes they get the right indentation (assuming they're using 2 spaces per indent)? Or do we include the entire path (ie `applications:<app-name>:hooks`) and _assume_ the user knows that they cant copy and paste the whole block?

I don't have a good answer.

gilzow:
@flovntp this is a _demonstration_ of an idea i pitched to Chad. What is in this PR would *replace* the Django blog post. The blog post then would focus on adding celery to a django project, or some other topic that isn't a good candidate for the documentation. 

The challenge I see with the current structure of the Upsun docs is, as a user, if I navigate to "[Deploying Express on Upsun](https://docs.upsun.com/get-started/stacks/express.html)" in the documentation, there is *zero* documentation on the page on how to actually deploy an express app on Upsun. There are some generic getting started links, and general nodejs links, but nothing about Express. You have to scroll all the way to the bottom (which is "[below the fold](https://www.optimizely.com/optimization-glossary/below-the-fold/)") to find a link to a blog post about how to possibly host an express app on Upsun. 

IMO, most users are going to hit this page, see no relevant information immediately and bounce. Instead, my suggestion is to put the needed information directly on the page. Blog posts then should be reserved for topics that _aren't_ good candidates for documentation.

gilzow:
Added the [section for Flask](https://docs.upsun.com.getting-started-demo-c6quj6q-ucq44jg6ofare.eu-5.platformsh.site/get-started/stacks/flask.html).

Not included is the information on how to change to alternative python web servers. Though we do have a whole section on python web servers with full instructions (/languages/python/server.html) so we could just add a note/link on this flask page. 

I also don't like that i have both the `.environment` method for adding the environment variable as well as the config file method. We are inconsistent on their usage (I'm just as much at fault here as everyone else). On the one hand it demonstrates both are valid options and exposes the user to both options. On the other, it adds complexity and noise to the instructions. 

We might need to consider an internally strategy/guidelines on when to use each method, at least for the documentation.

sites/friday/src/get-started/stacks/django.md
```diff
@@ -9,21 +9,113 @@ description: |
...
 {{< /note >}}
 
-## Getting started
+Now that you have completed the [Getting started guide](/get-started/here/_index.md), there are **six remaining changes**
+required in order to have a successful deployment of Django on Upsun. Alternatively, you can read the blog post
+[_Up(sun) and running with Django_](https://upsun.com/blog/setting-up-django-on-upsun/) which walks through the steps
+required to run Django on Upsun.
+
+## `.environment`
+Django requires a few environment variables that have not been set. Open the `.environment` file and add the following:
```
chadwcarlson, line 24:
```suggestion
Your Django app, depending on your `settings.py`, may require environment variables that have not been set. Open the `.environment` file and add them:
```

chadwcarlson: Not all Django apps will allow for this override AFAIK - pulling them in has to be done explicitly in the `settings.py` file.

sites/friday/src/get-started/stacks/django.md
```diff
@@ -9,21 +9,113 @@ description: |
...
+to:
+```yaml
+    build: |
+      set -eux
+      pip install --upgrade pip
+      pip install -r requirements.txt
+      npm install
+      npm run build
+```
+### Deploy Hook
```
chadwcarlson, line 83:
```suggestion
If your project uses a different package manager, we also support [several other options](/languages/python/dependencies.md).

### Deploy Hook
```

sites/friday/src/get-started/stacks/flask.md
```diff
@@ -7,19 +7,145 @@ description: |
...
+ ```yaml {configFile="app"}
+ hooks:
+     build: |
+         set -eux
+         pip install --upgrade pip
+         pip install -r requirements.txt
+         npm install
+```
+
+### Deploy Hook
```
chadwcarlson, line 103:
```suggestion
If your project uses a different package manager, we also support [several other options](/languages/python/dependencies.md).

### Deploy Hook
```

chadwcarlson:
I couldn't comment directly, but we should update the `Django topics` to `Flask topics`, no?

PR #3818, opened 2024-03-01
Specify that cache is always disabled on proxy routes

ralt:
Also, there's a zero-whitespace character in there that makes the sentence look weird. Removing it.


gilzow (approved):
we might need to make this a warning instead of a note but we can determine that next week.

PR #3900, opened 2024-03-27
Add premium label

MPPlexor:
ElasticSearch is a premium feature only.

## Why

Closes #{ISSUE_NUMBER}


MPParsley:
Would be nice to have a trial for Elastic though.

Issue #3967, opened 2024-04-17
Duplicate tests for Upsun docs


chadwcarlson:
### Where on docs.platform.sh should be changed?

.github/tests for Upsun

### What exactly should be updated?

See above

Issue #3916, opened 2024-04-03
Document local Vale + add auto script to run all tests


chadwcarlson:
### What needs to be documented?

https://github.com/platformsh/platformsh-docs/pull/3754

PR #3968, opened 2024-04-17
Duplicate tests + update contribution docs

chadwcarlson:
## Why

Closes https://github.com/platformsh/platformsh-docs/issues/3916
Closes https://github.com/platformsh/platformsh-docs/issues/3967

## What's changed

- Adds docs + local script for running tests locally during contribution
- Duplicated actions to also check Upsun docs
- Adds fixes to make those tests pass now
- Removes `ga-staging` branch definitions in workflows, that had been set up for the GA release
- Updates `linting/Linting` workflow to run off `pull_request` instead of `pull_request_target`
- Updates `Get info on PR` so that comment left on PR does so for both Platform.sh and Upsun docs.

### DIFF
--- a/.github/workflows/get-pr-info.yaml
+++ b/.github/workflows/get-pr-info.yaml
@@ -1,7 +1,7 @@
 name: Get info on PR
 on:
   pull_request_target:
-    branches: [main, ga-staging]
+    branches: [main]
     types: [labeled,opened,reopened,synchronize]
 
 env:
--- a/.github/workflows/get-pr-info.yaml
+++ b/.github/workflows/get-pr-info.yaml
@@ -1,6 +1,6 @@
 name: Get info on PR
 on:
-  pull_request:
+  pull_request_target:
     branches: [main]
     types: [labeled,opened,reopened,synchronize]
 
--- a/.github/workflows/get-pr-info.yaml
+++ b/.github/workflows/get-pr-info.yaml
@@ -1,6 +1,6 @@
 name: Get info on PR
 on:
-  pull_request_target:
+  pull_request:
     branches: [main]
     types: [labeled,opened,reopened,synchronize]
 
@@ -9,6 +9,7 @@ env:
   PLATFORM_PROJECT: ${{ vars.PROJECT_ID }}
   PLATFORMSH_CLI_DEFAULT_TIMEOUT: 60 # Increase timeout for CLI commands
   BRANCH_TITLE: ${{ vars.BFF_PREFIX }}-${{ github.event.number }}
+  UPSUN_DOCS_PREFIX: "https://docs.upsun.com"
 jobs:
   get_pr_info:
     runs-on: ubuntu-latest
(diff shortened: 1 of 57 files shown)


.github/workflows/linting.yaml
```diff
@@ -1,7 +1,7 @@
 name: Linting
 on:
-  pull_request_target:
-    branches: [main, ga-staging]
+  pull_request:
```
gilzow, line 3:
```suggestion
  pull_request_target:
```

gilzow: we still need `_target` in this case because the workflow needs access to the `GITHUB_TOKEN` secret.

.github/workflows/linting.yaml
```diff
@@ -40,6 +43,9 @@ jobs:
       ) &&
       contains(github.event.pull_request.labels.*.name, vars.LBL_MRKDWNPROSE)
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        site: ['platform', 'upsun']
     steps:
       - name: Checkout
         uses: actions/checkout@v3
```
gilzow, line 51: we now need to add 
`ref: ${{ github.event.pull_request.head.ref }}` 
to the checkout action so we can pull the code from the PR, not from the target. 

To limit malicious injections, we're only running this job if the `build-fork` label has been added so we need to extra cautious when applying that label to PRs from forks.

.github/workflows/linting.yaml
```diff
@@ -14,6 +14,9 @@ jobs:
       ) &&
       contains(github.event.pull_request.labels.*.name, vars.LBL_MRKDWNPROSE)
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        site: ['platform', 'upsun']
     steps:
     - name: Checkout
       uses: actions/checkout@v3
```
gilzow, line 22: we now need to add 
`ref: ${{ github.event.pull_request.head.ref }}` 
to the checkout action so we can pull the code from the PR, not from the target. 

To limit malicious injections, we're only running this job if the `build-fork` label has been added so we need to extra cautious when applying that label to PRs from forks.

.github/workflows/get-pr-info.yaml
```diff
@@ -1,14 +1,15 @@
 name: Get info on PR
 on:
-  pull_request_target:
-    branches: [main, ga-staging]
+  pull_request:
```
gilzow, line 3:
```suggestion
  pull_request_target:
```
We still need `_target` here, otherwise the workflow will not access to the PSH token secret needed later. 
To limit malicious injections, we're only running this job if PR comes from this repo itself OR the build-fork label has been added so we need to extra cautious when applying that label to PRs from forks.

chadwcarlson:
workflow needs to be returned to use `pull_request_target` prior to merging

gilzow (changes requested):

Issue #3138, opened 2023-05-08
Search engine grabbing wrong description


gilzow:
### Where on docs.platform.sh should be changed?

Search Results:
https://docs.platform.sh/search.html?q=openssl 

### What exactly should be updated?

Not an issue, but more of an oddity. 
For the PHP | Extensions page, the search engine is indexing the embedded table on the page which causes the search results to look like this:

![image](https://user-images.githubusercontent.com/2332956/236840323-98675e1f-8c79-4500-9239-c2b42d7e76b7.png)

### Additional context

What's odd is that there is a meta description on the page:

```
 <meta name="description" content="See what PHP extensions are available with each PHP version on Platform.sh.">
 ```
 
 Not sure why the search engine isn't using it for the results description...

chadwcarlson:
There was an attempt, but this is really a bigger fix to come after improving the local search experience and upgrading everything.

chadwcarlson:
Closing this in favor of the broader search fix issue (#3092)

PR #3969, opened 2024-04-17
Fix meilisearch grabbing wrong description.

chadwcarlson:
## Why

Closes https://github.com/platformsh/platformsh-docs/issues/3138

## What's changed

Displayed attribute s/text/description


chadwcarlson:
On second thought this should only be displayed. Need to follow-up once passes

PR #4005, opened 2024-04-30
Update cmd to commands in cron configuration examples

### LINKED ISSUE #3935: Cron configuration changes

https://docs.platform.sh/create-apps/app-reference/single-runtime-image.html#cron-commands

Not sure since when, but the cron configuration used to be like:
```
crons:
    name:
        spec: ' * * * * *'
        cmd: 'echo hello'
```
but I found the following version in a project using a fresher psh template:
```
crons:
    name:
        spec: ' * * * * *'
        commands: 
            start: 'echo hello'
```

which I originally thought is wrong, but after checking psh template and docs, looks like the correct version. 
So what I am currently missing right now and would love to have available in the docs is:
- the "old way" seems to still work, but will this work forever or will it be deprecated at some point?

Thanks!

### Additional context

_No response_

AnouckColson:
## Why

Closes #3935

`cmd` has been deprecated for a while in favour of `commands`(as documented [here](https://docs.platform.sh/create-apps/upgrading.html#changes-in-version-202202)).

There were some snippets still using the deprecated syntax.

## What's changed

Updated snippets.

### DIFF
--- a/sites/platform/src/create-apps/app-reference/composable-image.md
+++ b/sites/platform/src/create-apps/app-reference/composable-image.md
@@ -1201,7 +1201,7 @@ crons:
   # Run Laravel's scheduler every 5 minutes.
   scheduler:
     spec: '*/5 * * * *'
-    cmd: 'php artisan schedule:run'
+    commands: 'php artisan schedule:run'
 {{< /snippet >}}
 ```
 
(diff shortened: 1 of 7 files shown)


sites/platform/src/create-apps/app-reference/composable-image.md
```diff
@@ -1201,7 +1209,7 @@ crons:
   # Run Laravel's scheduler every 5 minutes.
   scheduler:
     spec: '*/5 * * * *'
-    cmd: 'php artisan schedule:run'
+    commands: 'php artisan schedule:run'
```
chadwcarlson, line 1212: These examples should use `commands.start` rather than just `commands`.

chadwcarlson (approved):

PR #4046, opened 2024-05-31
Update info on mount types in migration tutorial

AnouckColson:
## Why

The tutorial wasn't updated when local mounts were reinstated as `instance` mounts on Upsun side.

## What's changed

Updated info on mount types in tutorial.

### DIFF
--- a/sites/upsun/src/learn/tutorials/migrating/from-psh.md
+++ b/sites/upsun/src/learn/tutorials/migrating/from-psh.md
@@ -195,11 +195,10 @@ mounts:
         source_path: private
 ```
 
-{{< note >}}
-{{< vendor/name >}} doesn't currently support the same `local` mounts as {{< vendor/psh_ref >}}.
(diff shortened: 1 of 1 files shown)


gilzow (approved):

sites/upsun/src/learn/tutorials/migrating/from-psh.md
```diff
@@ -195,23 +195,22 @@ mounts:
...
-For instance:
+For example:
 
 ```yaml {configFile="app"}
 applications:
     myapp:
         mounts:
             'web/uploads':
-                source: storage
+                source: instance
```
gilzow, line 210:
```suggestion
                source: storage
```

gilzow: this mount type needs to be shared between instances so will need to be the `storage` type

PR #4047, opened 2024-05-31
Erratum: web/uploads mount should be a storage mount

AnouckColson:
## Why

Error from #4046 

## What's changed

Fixed mount type in snippet and explanatory sentence.


gilzow (dismissed):
thanks for fixing my goof

gilzow (approved):

Issue #4030, opened 2024-05-22
Update Slack notifications


chadwcarlson:
### Where on docs.platform.sh should be changed?

- Custom bot user tokens are deprecated as of June 4.
- existing bot user tokens will not stop working
- all new tokens must be from a slack app

### What exactly should be updated?

- I created a Slack app: https://api.slack.com/apps
- Once created, went to OAuth & Permissions --> Scopes
- Under Bot Token Scopes, added the chat:write OAuth scope
- Installed and Authorized the app
- Added the app to a channel I created: [testing-stuff](https://platformsh.slack.com/archives/C073X5KKRK6)
- Retrieved the Bot User OAuth Token from OAuth & Permissions
- Used the instructions from https://docs.platform.sh/integrations/notifications.html#slack-notifications to run the command

platform integration:add --type health.slack --token YOUR_API_TOKEN --channel '#channelname'
and got the notifications in the Slack channel

### Additional context

- https://api.slack.com/legacy/custom-integrations/bot-users
- https://api.slack.com/quickstart
-

PR #4048, opened 2024-05-31
Describe new Slack app process.

chadwcarlson:
## Why

Closes https://github.com/platformsh/platformsh-docs/issues/4030


sites/platform/src/integrations/notifications.md
```diff
@@ -74,17 +74,42 @@ You can also configure a custom `--from-address`. The `--from-address` is whatev
...
+    {{% vendor/cli %}} integration:add --type health.slack --token {{% variable "USER_OAUTH_TOKEN" %}} --channel {{% variable "CHANNEL_NAME" %}} --project {{% variable "PROJECT_ID" %}}
+    ```
+
+    If the channel you're installing the app is called `project-notifications`, write the channel name in the command above as you would reference it within Slack. 
+    That is, 
+
+    ```bash
+    {{% vendor/cli %}} integration:add --type health.slack ... --channel '#project-notifications' ...
+    ```
+1. {{% vendor/name %}} will then send an initial message to the channel once the integration is successfully configured.
```
chadwcarlson, line 105:
```suggestion
{{% note theme="info" %}}
If you are _not_ a workspace admin, but an admin has approved your app for installation it may first be necessary to install the application to the desired channel within Slack prior to running the above command. 
{{% /note %}}


1. {{% vendor/name %}} will then send an initial message to the channel once the integration is successfully configured. 
```

AnouckColson (approved):

PR #4084, opened 2024-06-27
Update vertical scaling 

chadwcarlson:
Reverts platformsh/platformsh-docs#4083


sites/upsun/src/manage-resources/adjust-resources.md
```diff
@@ -60,6 +67,41 @@ you may need to set the resources again.
 After you've set resources, your environment is redeployed,
 which causes a short downtime.
 
+- **Manual commands:**
+
+Use the following CLI flags:
```
nicogommen, line 72: I find the word flags a bit confusing.
I would use options instead, and maybe a sentence like « Use the following options with the resources:set command: ».
And same in the table below.

But it depends if we are used to use the term flag elsewhere to describe options.

But this is a minor detail, otherwise this whole PR looks great to me.
Thanks.

sites/upsun/src/manage-resources/resource-init.md
```diff
@@ -397,18 +397,15 @@ title=In the Console
 
 When you [restore a backup](/environments/restore.md) using the Console, it is restored to your current environment.
 
-Each container already running on your current environment keeps its existing resources.
-Therefore, you don't need to specify a resource initialization strategy. 
+Containers are restored using the `backup` strategy, meaning:
+- The resources of every container running on the environment are reverted to their original state when the backup was taken.
```
nicogommen, line 401: This is not because the `backup` strategy is used.

So I would first explain above the default behaviour, that we restore resources to their original state when the backup was taken.

But that you can also use the resource initialization strategy to define how new containers will be restored, knowing that in the case of a backup restore, there is only 2 cases when new containers are initialized => 1) when you restore to a new environment, in that case, all containers are new, or 2) when you restore to an environment where you have removed some containers since the backup was taken, in that case, these removed containers will be restored using the resource initialization strategy.

nicogommen: I was also thinking that we might want to add a short sentence about this in these 2 pages:
https://docs.upsun.com/environments/restore.html#2-restore-from-a-backup
https://docs.upsun.com/environments/backup.html

sites/upsun/src/manage-resources/resource-init.md
```diff
@@ -52,7 +52,7 @@ Note that you can [monitor these costs](/administration/billing/monitor-billing.
 | `minimum`  | Initializes new containers using the {{% vendor/name %}} minimum resources (see below). |
 | `parent`   | Initializes new containers using the same resources as on the parent environment.</br>If there is no parent environment, or if the container doesn't already exist on the parent, the `default` strategy applies instead. |
 | `child`    | Initializes new containers using the same resources as on the child environment. Only relevant during merge activities. |
-| `backup`   | When restoring a backup, initializes new containers using the same resources as when the backup was taken. |
+| `backup`   | When restoring a backup, all containers are restored using the same resources as when the backup was taken. |
```
nicogommen, line 55: @AnouckColson, I would actually keep the previous version here. Because the init strategy is actually only for initialization of new containers. Even is the default will be to restore all the resources, but it's not necessarily because of the `backup` init strategy.

PR #4085, opened 2024-06-27
Adds e2e testing

gilzow:
## Why

Adds initial set of end-to-end testing to docs

## What's changed

- adds cypress as a dev dependency (for when running locally)
- Adds workflow to trigger cypress test when a PR is created.

Currently runs 3 tests against each vendorized doc (platform.sh / upsun.com) that
* searches for a term that should match on both sites
* searches for a term that should only match on platform.sh and not match on upsun.com
* searches for a term that should not match on platform.sh but should match on upsun.com

TODO: 

- Expand the local intercepts to include fixtures for the different searches (this pr only includes a fixture for "opensearch")
- Add support for Gherkins ([cucumber](https://www.npmjs.com/package)) feature files so non-devs can add e2e testing when adding docs
- Add support in the workflow so that each site tests from the current shared collection of spec files as well as individual site spec files


sites/platform/static/scripts/xss/src/containers/Search.js
```diff
@@ -7,6 +7,7 @@ import Suggestions from 'components/Suggestions'
 import SuggestionsPrimary from 'components/SuggestionsPrimary'
 
 const getConfig = async () => {
+  console.log('Retrieving config')
```
chadwcarlson, line 10: 4 `console.log` mentions to remove in this file.

chadwcarlson, line 10:
```suggestion
```

sites/platform/static/scripts/xss/src/containers/Search.js
```diff
@@ -15,6 +16,7 @@ const getConfig = async () => {
   // here if they are not yet set, but a file works just fine.
   // The mount `public/scripts/xss/dist/config` has been defined to support this.
   const response = await fetch(`/scripts/xss/dist/config/config.json?version=${Date.now().toString()}`);
+  console.log('returning config')
```
chadwcarlson, line 19:
```suggestion
```

sites/platform/static/scripts/xss/src/containers/Search.js
```diff
@@ -42,6 +44,14 @@ const Search = ({ fullPage }) => {
   const limit = fullPage ? maxResults : 7
 
   const getInfo = (infoConfig, infoQuery) => {
+    if (!Object.hasOwn(infoConfig, 'url')) {
+      console.log('Get info called but config object has missing properties!')
```
chadwcarlson, line 48:
```suggestion
```

sites/platform/static/scripts/xss/src/containers/Search.js
```diff
@@ -42,6 +44,14 @@ const Search = ({ fullPage }) => {
...
+      console.log('trying to get config props again')
```
chadwcarlson, line 49:
```suggestion
```

chadwcarlson (approved):

PR #4094, opened 2024-07-04
ruby: puma is default in latest RoR

spk:
## Why

Puma has been the default now on RoR.

## What's changed

This also simplify the first step to not be too verbose or intimidating for newcomers (full version kept at the end)


chadwcarlson:
Thanks @spk!

PR #4138, opened 2024-08-01
Fix malformed shell script example in set-variables.md

pjcdawkins:


flovntp (approved):
i also report the same update on Upsun side

chadwcarlson (approved):

PR #4150, opened 2024-08-19
Update NodeJS Custom Version Instructions

raupie:
## Why

The instructions need to be clear on how to make the custom nodejs version work during runtime.

## What's changed

I moved the "symlink note" further down into step 5 and stated the rsync command is required to be added to make this work during runtime.


gilzow (changes requested):

flovntp:
Hi @raupie 
Is it also valid for Upsun (cf. /sites/upsun/) ?

raupie:
> Hi @raupie Is it also valid for Upsun (cf. /sites/upsun/) ?

Yes, for sure. Apologies for skipping that.

Issue #4147, opened 2024-08-14
Update process for installing Python Poetry


hannesvz:
### Where on docs.platform.sh should be changed?

https://docs.platform.sh/languages/python/dependencies.html#poetry

### What exactly should be updated?

Instead of the `curl -sSL https://install.python-poetry.org/ | python3 -` step in the build hook, Poetry should be installed via the `dependencies` key in the app configuration.

Remove:

```
curl -sSL https://install.python-poetry.org/ | python3 -
```

And add:

```
    dependencies:
        python3:
            poetry: '1.8.3'
```

### Additional context

A recent update changed this behaviour and could cause `poetry` to fail certain builds. More context here:

https://platformsh.slack.com/archives/C8DNLHHL7/p1723641009762759?thread_ts=1723547937.839699&cid=C8DNLHHL7

PR #4206, opened 2024-09-23
updated code snippet

Kemi-Elizabeth:
Updated poetry python code with new version

## Why

Closes https://github.com/platformsh/platformsh-docs/issues/4147

## Where are changes

Updates are for:
[] platform (`sites/platform` templates)
[] upsun (`sites/upsun` templates)


chadwcarlson (approved):
All looks good, my only suggestion was to revert the `latest` shortcode. This will make sure that the `python3.X` call in the build hook always matches what we have in `type`.

chadwcarlson (approved):

PR #4208, opened 2024-09-25
postgresql: add info about local pg_dump with tunnel

spk:
## Why

Improve doc with postgresql pg_dump

## What's changed

Add a tips for user that want to generate the dump from local pg_dump command

## Where are changes

Updates are for:


Kemi-Elizabeth (approved):
@spk Hi! I just edited your sentence a little bit if that's okay.

PR #4278, opened 2024-11-29
:bug: Fix exemplary cron syntax.

biertz:
## Why

The current example does not work. Apparently, the cron syntax had been changed from `cmd` to `commands.start` and `commands.stop` (as is documented on [another page of the documentation](https://docs.platform.sh/create-apps/app-reference/single-runtime-image.html#cron-commands)) and when the Cronjobs page was updated the syntax got mixed-up.

## What's changed

I changed both examples for cron jobs so that they fit the new syntax. I also took the liberty of making one example use a simple string and the other one a multi-line block to demonstrate both usages.

## Where are changes

Updates are for:


chadwcarlson (approved):

chadwcarlson:
Thanks @biertz !

PR #4285, opened 2024-12-03
refactor: improve visibility of TXT affecting apex domain

hanoii:
## Why

Yesterday I wanted to migrate the apex domain to a new project, from an old project.

I completely forgot about the subdomains protection as this was done years ago.

I struggled yesterday with this, causing annoyance to one of my clients and not being able to re-add the apex domain again to either project. It wasn't until support was able to re-add (which I now know it was because eventually I removed this TXT record even without knowing about this and the DNS spread hit the API server).

## What's changed

This just make it visually more obvious.

## Where are changes

Updates are for:

### DIFF
--- a/sites/platform/src/domains/steps/subdomains.md
+++ b/sites/platform/src/domains/steps/subdomains.md
@@ -19,6 +19,15 @@ You need to add a DNS record to make it clear you explicitly allow multiple proj
 
 ## Enable subdomains across multiple projects
 
+{{< note theme="warning" title="Apex domain cannot be re-added" >}}
+
+With the `TXT` record, restrictions apply on the apex domain.
+For example, you can't add the apex domain to another project until you remove the `TXT` record.
+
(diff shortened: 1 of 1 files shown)


hanoii:
I have a couple complains/FR around this which I'd request you to raise them internally:

- Why block the APEX domain with this TXT record? I don't really see the benefit of it.
- If there's a good reason, the API error should be clear of what the issue is and point you to remove the record and WAIT for it to be spread. 
- Even further, the `cli` and web interface could be more clear about this, like if you remove an APEX domain from a platform project, it could check that this record is present and look for a user confirmation and warn that you won't be able to re-add again after you remove that and wait the DNS to spread.

hanoii:
Related support ticket: https://support.platform.sh/hc/en-us/requests/342984

chadwcarlson (approved):

chadwcarlson:
Thanks @hanoii!

PR #4301, opened 2024-12-12
make the apex domain error more visually important

hanoii:
This is a follow up to #4285 . Thanks for merging but I was truly annoyed about this which ended up with a 30m downtime on the client end.

I think it's very important that's why I rather have it as a warning, and also make it very clear that  removing the TXT is not an immediate fix, because until your API server doesn't *see* that the TXT was removed the error will still appear and you will not be able to add the apex domain to any project.

I can't emphasise enough that whatever protection this txt record relates to the apex domain should be handled differently.


hanoii:
BTW, I took the language of the DNS from https://docs.platform.sh/domains/steps.html#2-configure-your-dns-provider

chadwcarlson (approved):

chadwcarlson:
Makes sense @hanoii. Thank you again.

PR #4306, opened 2024-12-13
DO NOT MERGE

gilzow:
Now the redirection verification check is not failing. this adds some fake contracted anchor redirections to make sure we're catching them as we should.


gilzow:
works as expected

PR #4312, opened 2024-12-18
feat: swapped steps 5 and 6 around

### LINKED ISSUE #4311: Change order of operations of default environment pages

 we need to make a change to the order of operations listed on https://docs.upsun.com/environments/default-environment.html and https://docs.platform.sh/environments/default-environment.html
Basically, Step 5 and Step 6 need to be swapped around - in a lot of cases, you can't complete step 5 without completing step 6 first, which is obviously a little backwards.

- https://docs.upsun.com/environments/default-environment.html
- https://docs.platform.sh/environments/default-environment.html

### Additional context

_No response_

Kemi-Elizabeth:
Swapped steps around for betting understanding of process.

## Why

Closes #4311 

## What's changed

Steps 5 and 6 have been swapped around on the rename default environment pages on both Platform.sh and Upsun.

## Where are changes

Updates are for:


gilzow (approved):
Confirmed this works with the new order in both upsun and platform.sh

PR #4319, opened 2025-01-06
ALLOWED_HOSTS instructions incorrect

OriPekelman:
By default, Django's domain name pattern matching in ALLOWED_HOSTS only supports single-level wildcard subdomains (e.g., *.example.com). Multiple levels of wildcards (like *.*.example.com) are not supported. as such '.platformsh.site' will not match '*.us-2.platformsh.site'. Users may also set the ALLOWED_HOSTS to 'us-2.platformsh.site'. but that would require specifying the region in the configuration.

## Why

Closes #{ISSUE_NUMBER}

## Where are changes

Updates are for:


sites/upsun/src/get-started/stacks/django.md
```diff
@@ -40,18 +40,13 @@ See the following sections to see how other common settings should be set on {{%
 
 ## 2. Configure `ALLOWED_HOSTS`
 
-Your `settings.py` file may not allow you to use an environment variable like `DJANGO_ALLOWED_HOSTS`.
-If so, to configure allowed hosts, update your `settings.py` file to include `{{< vendor/urlraw "hostname" >}}`:
+By default other than `localhost` Django only allows hosts listed in `settings.ALLOWED_HOSTS` to be accessed. By default Django does not allow for wildcard hosts than span multiple levels so in order to support our dynamic preview environments you will want to dynamically add those to the list.
```
chadwcarlson, line 43:
```suggestion
By default, other than `localhost`, Django only allows hosts listed in `settings.ALLOWED_HOSTS` to be accessed. However, Django does not allow for wildcard hosts that span multiple levels by default. This becomes relevant in order to support our dynamic preview environments you will want to dynamically add to the list.
```

chadwcarlson (approved):

PR #4333, opened 2025-01-09
MFA term consistency

chadwcarlson:
## Why

Internal request to update docs to match other properties in referring to MFA. 

## Where are changes

Updates are for:


pjcdawkins:
Unfortunately we have this term in plenty of other places with the original format: at least the CLI and API

PR #4334, opened 2025-01-09
feat: corrected region names in table

### LINKED ISSUE #4331: Correct region names

All region names are wrong so they need to be corrected.

For example: [ch-1.upsun.com](http://ch-1.upsun.com/) should be ch-1.platform.sh as there is no difference between psh and Upsun here.

https://docs.upsun.com/development/regions.html

### Additional context

_No response_

Kemi-Elizabeth:
## Why

Closes #4331 

## What's changed

So I had to correct the region names in the table on [this page](https://docs.upsun.com/development/regions.html). I also had to correct the region names in the list below the `Where REGION is one of the impacted regions` section. 

The region names were displayed as `us-3.upsun.com` for example, when they should be `us-3.platform.sh` regardless of whether it's on Upsun or Platform.sh.

## Where are changes

I went into the `regions.md` file and replaced the shortcode with text so regions would change from [ch-1.upsun.com](http://ch-1.upsun.com/) to ch-1.platform.sh. 

I also went into the `regions.html` file and changed the `code class="!bg-snow">`. In particular, I removed the code that was pulling in the site url automatically and just replaced that with text that said "platform.sh". It seemed like the easiest way to ensure it's hard coded and consistent across Upsun and Platform.sh. 

I didn't change the Inbound IPs column in the table as that seemed correct. It points to an url ending `platformsh.site`. Can I confirm that's correct please? 

Updates are for:


gilzow (approved):

gilzow:
This "situation" is a remnant of when the two docs were intertwined, and missed when they were separated. Since we're planning on restructuring docs anyway, I think hard-coding the `platform.sh` parts is perfectly acceptable.

PR #4373, opened 2025-02-06
Accessibility: Fix aria labels on section buttons

Menelion:
## Why

In current platform.sh docs, there are quite a number of accessibility errors. Most of them are just annoying, but there was one thing that prevented really from using the docs with a screen reader correctly.
Each section has a button to expand it, and this button has an `aria-label` of "Toggle section". So imagine if each button had the same text: "Toggle section". that's how I see it on the currently released pages.

## What's changed

If a button contains an SVG image (correctly hidden with `aria-hidden="true"` btw), then the `aria-label` is changed to have the title of the section in question. If however the section title was the text of the button itself (simplified code: `<button>{{ .Title }}</button>`), then the `aria-label` was removed as it is not needed anymore: the text between the `button` tags is read by all screen readers.

## Where are changes

Updates are for:


flovntp (approved):
Thanks for your contribution

PR #4386, opened 2025-02-13
feat: added paragraph about resources

### LINKED ISSUE #4287: Clarify how resources work on Upsun

On Upsun documentation website (https://docs.upsun.com/), we'd like to better describe how resources work on our Grid infrastructure.
A bit similarly to what we are doing in the new Dedicated pages (https://docs.platform.sh/dedicated-environments/overview.html#grid-vs-dedicated).
For example, mention wherever relevant that the projects on Upsun are running on containers with dedicated resources on top of a shared redundant infrastructure.

Not sure yet which pages would be the most relevant.
We first need to analyze where exactly this information would fit best.

### Additional context

We might offer more types of resources on Upsun at some point (dedicated resources, guaranteed resources, performance CPU, ...etc.), so we will need to introduce and explain the differences between the new type and the current standard resources. Adding this clarification about the current resources now already would be helpful in anticipation for when will have more types available.

Kemi-Elizabeth:
**Added resources paragraph**

## Why

Closes #4287 

## What's changed
I've added a paragraph in the Manage Resources index that clarifies how resources work on Upsun. I'm not sure if it should go on this page or if it should go on /manage-resources/adjust-resources.html. 

## Where are changes
/manage-resources.html

Updates are for:


sites/upsun/src/manage-resources/_index.md
```diff
@@ -16,4 +16,16 @@ keywords:
 {{% vendor/name %}} allows you to configure resources (CPU, RAM, and disk) per environment for each of your apps and services.
 You can also add instances for each of your apps depending on your needs.
 
-For information on costs related to resource usage, see the [{{% vendor/name %}} pricing page](https://upsun.com/pricing/).
\ No newline at end of file
+For information on costs related to resource usage, see the [{{% vendor/name %}} pricing page](https://upsun.com/pricing/).
+
+## How resources work on Upsun
+
+Projects on Upsun run on a Grid host where processing power is shared among customers. This means that the amount of processing power allocated to your project is dependent on the resources other customers are using.
```
nicogommen, line 23: Hey @Kemi-Elizabeth , this is what I'm suggestion for this `How resources work on Upsun` section:

> Projects on Upsun run on shared Grid hosts, which means that the processing power allocated to your containers can be dependent on the resources other customers are using on the same Grid hosts. Our orchestration layer works to ensure stable access to resources when possible, however some variability should be expected due to the shared infrastructure of our Grid regions.

We can then keep the note below.
Before publishing it, I would like to get a final review from some stakeholders. So, if you agree with the above, it would be great if you can update the preview env, so we can share it directly with other people for this final review.
Let me know if that makes sense.
Thank you.

fabpot (approved):

flovntp (approved):

PR #4399, opened 2025-02-14
Fix link checker

gilzow:
## Why

We need to make sure links are not broken

## What's changed

The link checker was being flagged by our WAF and blocking connections. This change lowers the number of concurrent threads and time between requests to keep from being blocked. It'll be slower but given this runs on a schedule and not in our CI process, it shouldn't interfere with daily work.


gregqualls (dismissed):
everything looks good and makes sense. :D

gregqualls (approved):
Looks good.

PR #4423, opened 2025-02-25
Upsun: removing sectionBefore for Drupal Guide

flovntp:
## Why

Closes #{ISSUE_NUMBER}

## Where are changes

Updates are for:

### DIFF
--- a/sites/upsun/src/get-started/stacks/drupal.md
+++ b/sites/upsun/src/get-started/stacks/drupal.md
@@ -3,7 +3,7 @@ title: Deploy Drupal on {{% vendor/name %}}
 sidebarTitle: Drupal
 sectionBefore: PHP
 #layout: single
-weight: -66
+weight: -67
 description: |
     Complete the last required steps to successfully deploy Drupal on {{% vendor/name %}}.
 ---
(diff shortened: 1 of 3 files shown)


gilzow:
@flovntp unfortunately this bumps it into the Python section which is the issue I was having previously 
image

sites/upsun/src/get-started/stacks/drupal.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: Deploy Drupal on {{% vendor/name %}}
 sidebarTitle: Drupal
-sectionBefore: PHP
+#sectionBefore: PHP
 #layout: single
 weight: -67
```
gilzow, line 6:
```suggestion
weight: -57
```

gilzow: this is how Kemi has it set. let's do this and then we can merge this one while we're finishing the magento guide.

gilzow (changes requested):

flovntp:
@gilzow i told you i would find.
That's a bit weird but why not: from the entire list of PHP template, only the first one can have the param `sectionBefore`. 
removing `sectionBefore` from Laravel guide did the trick ;)

Kemi-Elizabeth:
Hi @flovntp! Is it okay for me to close this PR as it's been solved now? :)

PR #4446, opened 2025-02-28
limits job to run only if PR workflow completed successfully

### LINKED ISSUE #4443: Add conditional logic to comment on pr workflow

N/A

The [Comment on PR ](https://github.com/platformsh/platformsh-docs/blob/main/.github/workflows/comment-on-pr.yaml) workflow is currently triggered via `workflow_run` from the completion of the `Get PR URL, start dependent jobs` workflow. We need to add a conditional for the job so that it only runs if that workflow's conclusion is success:

```
if: ${{ github.event.workflow_run.conclusion == 'success' }}
```

### Additional context

_No response_

gilzow:
## Why

Closes #4443 

## What's changed

sets the comment on PR workflow job to only run if the PR url workflow completes _successfully_.

### DIFF
--- a/.github/workflows/comment-on-pr.yaml
+++ b/.github/workflows/comment-on-pr.yaml
@@ -12,6 +12,7 @@ on:
 jobs:
   comment:
     runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
     steps:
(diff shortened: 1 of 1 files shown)


gregqualls (approved):
Thats a nice line of code right there.

PR #4455, opened 2025-03-06
Remove the lisp image and references to it

danmilon:
See [IMG-1954: fully drop the lisp image](https://linear.app/platformsh/issue/IMG-1954/fully-drop-the-lisp-image).

The image is not used at all, so lets sunset it to reduce maintenance cost.


flovntp:
Hey @danmilon 
i think you miss the recent split between single-runtime and Composable image. 
This [page](https://bffpr-4455-pbevw5i-ucq44jg6ofare.eu-5.platformsh.site/create-apps/app-reference/composable-image.html#container-profile-reference) still reference Lisp (and same on Upsun side).  
Could you please also modify the composable-image.md files on both Upsun and P.sh? 
https://github.com/platformsh/platformsh-docs/blob/main/sites/platform/src/create-apps/app-reference/composable-image.md?plain=1#L121

I would be happy to contribute to your branch, but it's a fork and i can't :D 

ps: 
- [Upsun preview env](https://docs.upsun.com.bffpr-4455-pbevw5i-ucq44jg6ofare.eu-5.platformsh.site/)
- [P.sh preview env](https://bffpr-4455-pbevw5i-ucq44jg6ofare.eu-5.platformsh.site/)

flovntp:
or maybe i'm wrong and using Composable image, Common Lisp (SBCL) is still supported.
If that's the case, i would gladly approve you PR ;)

danmilon:
yeah at the moment there's one production use of lisp on the composable image, and it's supposed to be supported. I think it makes sense to keep this one for now.

flovntp (approved):

flovntp:
Last thing, could you please remove the link in the composable-image.md files (Upsun and P.sh) to the lisp page?

danmilon:
ah I see. so the composable image pages link to the single-runtime pages? But if we remove the link, it's as if the composable image doesn't support lisp. Anyways, it's OK to undocument it completely then. what do you think?

flovntp:
https://devcenter.upsun.com/posts/deploying-with-lisp/

Kemi-Elizabeth (approved):

pjcdawkins:
Note I think Lisp will also need to be dropped [from `platformify`](https://github.com/platformsh/platformify/blob/c37c6a6e84e0af54bd775fbc84da5fd1ed924177/internal/question/models/runtime.go#L12) (the `project:init` CLI command) and potentially other places like templates

PR #4564, opened 2025-04-23
Add python 3.13 and ruby 3.4

spk:
## Why

Add python 3.13 and ruby 3.4 images are in production

## Where are changes

Updates are for:


gmoigneu:
💚

PR #4565, opened 2025-04-23
fix: latest fetch on partials

spk:
The result is kind of random currently, for example on https://docs.upsun.com/languages/python/dependencies.html it currently shows python:3.9 where it should be [3.12](https://github.com/platformsh/platformsh-docs/blob/6de635dd2393479383300a34e02e5908e053f10a/shared/data/registry.json#L945)

image

## Why

It doesn't show the latest version currently since it a list of strings

## What's changed

Just fetch for the first version defined on the registry since its already sorted

## Where are changes

Updates are for:


gmoigneu (approved):
💚

PR #4566, opened 2025-04-23
docs: remove composable disclaimer

spk:
## Why

Since composable composable image is not going to get out from beta status let stop promoting it on all [runtimes ex for ruby](https://docs.upsun.com/languages/ruby.html).

Screenshot from 2025-04-22 10-47-06

## Where are changes

Updates are for:


gmoigneu:
@fabpot Would love your opinion on this.

PR #4580, opened 2025-04-29
Update single-runtime-image.md

### LINKED ISSUE #4504: Improve mounts documentation - add pros and cons for different types of mounts

Expand on the pros and cons of each type of mounts on both Upsun and Platform, but especially. For example, we need to clarify the differences between local and remote storage. We also need to explain the difference between storage, service, temporary and instance. 

**Proposal:**
Add a section called "Different kinds of Mounts" or "Mount types" where each mount type is broken down, explained and compared to the other with a pros and cons table and an example of how to implement. 

https://docs.upsun.com/create-apps/app-reference/single-runtime-image.html#mounts
https://docs.upsun.com/create-apps/app-reference/composable-image.html#mounts (Clarify with Fabien) 

https://docs.platform.sh/create-apps/app-reference/single-runtime-image.html#mounts
https://docs.platform.sh/create-apps/app-reference/composable-image.html#mounts

### Additional context

https://linear.app/platformsh/project/configurable-and-billable-instance-storage-on-upsun-894840e10085/overview

Kemi-Elizabeth:
## Why

Closes #4504 

## What's changed
Added section to single runtime image page.

## Where are changes

Updates are for:

### DIFF
--- a/sites/upsun/src/create-apps/app-reference/single-runtime-image.md
+++ b/sites/upsun/src/create-apps/app-reference/single-runtime-image.md
@@ -374,6 +374,51 @@ much like you would plug a hard drive into your computer to transfer data.
 
 {{% /note %}}
 
+### Mount types
+
+Each mount type has its specific use case, whether for shared storage across instances, local storage for individual instances, or temporary storage for ephemeral data. Understand how to use each type effectively to optimize your application's performance.
+
+#### `storage` mount
+
+The `storage` mount is a **shared network storage** service that is automatically added by the platform. It is designed for **sharing files between multiple instances** of an application when scaling horizontally—such as images, uploaded assets, or other shared resources. Users can configure the size of this mount directly through the {{% vendor/name %}} UI, providing flexibility based on the project’s specific storage needs. This mount type ensures consistency and accessibility of files across all instances of a scaled application.
+
+**Pros:**
+- Shared across all instances – great for horizontal scaling.
+- Size is configurable via the UI.
+- Persistent – data survives restarts and redeployments.
+
+**Cons:**
+- Network latency may be slightly higher than local storage.
+- Less suitable for high-throughput temporary file operations.
+
+
+#### `instance` mount (`local`)
+
+The `instance` mount, also known as `local`, provides a **dedicated disk per application instance**. It is backed by a Ceph-mounted volume and is **specific to a single instance**, meaning the data stored here is not shared between scaled instances. By default, each `instance` mount has a size of **8GB**, and users currently **cannot adjust this size** through self-service. This mount is ideal for storing files that should remain local to an individual instance, such as **application logs** or instance-specific temporary data.
+
+**Pros:**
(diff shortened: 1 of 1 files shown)


sites/platform/src/create-apps/app-reference/composable-image.md
```diff
@@ -598,6 +598,61 @@ much like you would plug a hard drive into your computer to transfer data.
...
+### Mount types
+
+Each mount type has a specific use case and understanding how to use each type effectively will help optimize your application.
+
+#### `storage` mount
+
+The `storage` mount is a shared network storage that is automatically added by the platform. It is required for applications that need to share files across multiple instances—such as user-uploaded media or shared assets. You can configure the size of this mount via the {{% vendor/name %}} UI.
+
+Pros:
+- Shared across all app and worker instances.
```
ralt, line 610: The worker instances do not get to share the same mount as the app instances for this kind of storage. You need to declare an explicit network-storage service and use that if you want to share a mount between app and workers. The default `storage` mount is not shareable. The purpose is for horizontally-scaled applications to use the same files.

sites/platform/src/create-apps/app-reference/composable-image.md
```diff
@@ -598,6 +598,61 @@ much like you would plug a hard drive into your computer to transfer data.
...
+- Configurable size via the UI.
+- Persistent across restarts and redeployments.
+
+Cons:
+- Lower performance due to network-based access.
```
ralt, line 615: Technically, the `instance` storage is also a network-based volume :sweat_smile: 

The `storage` mount is an NFS volume and not a "block storage" kind of mount, so it has a lot more synchronization overhead. Not sure how to express that in this doc without getting too much into details though...

sites/platform/src/create-apps/app-reference/composable-image.md
```diff
@@ -598,6 +598,61 @@ much like you would plug a hard drive into your computer to transfer data.
...
+
+Cons:
+- Requires explicit setup and service configuration.
+- Slightly more complex than default `storage` mounts.
+
+For setup instructions and more information, see the [Network Storage service documentation](/add-services/network-storage).
+
+#### Scaling
+
+When horizontally scaling your app (multiple instances of the same app), using the appropriate mount type is critical. For shared data, a `storage` or `service` mount is recommended. For local-only data, use `instance` or `tmp` mounts.
```
ralt, line 650: It's a bit weird to mention `instance` mounts here given that they're not defined

ralt (commented):
Overall the text is correctly explaining the spirit of the mounts! Very happy to see this. I left a few comments that you could almost describe as nitpicky. Feel free to ignore.

sites/platform/src/create-apps/app-reference/composable-image.md
Kemi-Elizabeth, line 610:
```suggestion
```

PR #4595, opened 2025-05-07
enables robots.txt for both sites

gilzow:
Enables robots.txt in Hugo config for both sites. 
## Where are changes

Updates are for:


gregqualls (approved):
code looks good.  As long as there are no security concerns lets do it.

PR #4597, opened 2025-05-07
4596 guide wonkiness

### LINKED ISSUE #4596: Guides section is wonky

https://docs.platform.sh/guides.html

and by "wonky" i mean:

Image

Someone put anchors in the description meta for Django, symfony, laravel, etc. When this page was changed to use "cards", the card is surrounded with an `<a` this means we now have nested `<a` elements which you can't do. 

We need to remove the anchors in the description meta.

### Additional context

This is identical to what happened in  #4464

gilzow:
## Why

Closes #4596

## What's changed

Removes the anchor from the description meta data for several guides, and makes that description information static in the guide itself. This is required so that we don't end up with nested anchor elements.

## Where are changes

Updates are for:

### DIFF
--- a/sites/platform/src/guides/django/_index.md
+++ b/sites/platform/src/guides/django/_index.md
@@ -3,10 +3,10 @@ title: "Django"
 weight: -200
 sectionBefore: Python
 description: |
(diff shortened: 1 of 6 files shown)


gregqualls (approved):
looks good

PR #4709, opened 2025-06-25
Expand MariaDB/MySQL External Replication info [4700]

### LINKED ISSUE #4700: Expand MariaDB/MySQL External Replication page with new section about disabling external replication

**Insert section in [this page](https://docs.platform.sh/add-services/mysql/mysql-replication.html) that explains the following process:**

`When replication is disabled, the binary logs will stay on the filesystem as the mariadb service will not manage them anymore. These will have to be removed on our side. Depending on if the remote replica was not reachable for a while, these binary logs can end up using up a significant amount of storage.`

**Suggested change**
Add a section before [the 'starting the replica' section](https://docs.platform.sh/add-services/mysql/mysql-replication.html#starting-the-replica) that discusses how replication works and sets the stage for what happens after it's running or if it stops. 

**Suggested section (feel free to adjust the text or H3)**

> ### Binary Log Retention and Cleanup (or Disabling External Replication)
> 
> When replication is disabled, the binary logs will remain on the filesystem, as the `MariaDB` service will no longer manage them. These logs must be manually removed. If the remote replica has been unreachable for some time, the accumulated binary logs can consume a significant amount of storage.

### Additional context

https://platformsh.slack.com/archives/C022N5Q5LDP/p1750753386242109

catplat:
## Why

Closes #4700  

## What's changed
New:
- Binary log retention and cleanup section

Changed:
- Set up SSH tunneling section: Changed the organization. Please verify the technical meaning is intact.
- Create a replication user section: changed paragraph to a bulleted list for easier scanning
- Assorted minor style changes

## Where are changes
/add-services/mysql/mysql-replication.html 

Updates are for:

- [ X] platform (`sites/platform` templates)

### DIFF
--- a/sites/platform/src/add-services/mysql/mysql-replication.md
+++ b/sites/platform/src/add-services/mysql/mysql-replication.md
@@ -6,16 +6,15 @@ description: In rare cases, it may be useful to maintain a replica instance of y
 
 In rare cases, it may be useful to maintain a replica instance of your MySQL/MariaDB database outside of {{% vendor/name %}}.
 
-Normally an automated backup is better for short-term usage and a `mysqldump` for longer term storage, but in some cases the data set is large enough that `mysqldump` is prohibitive.
+Typically, an automated backup is better for short-term usage and a `mysqldump` for longer term storage, but in some cases the data set is large enough that `mysqldump` is prohibitive.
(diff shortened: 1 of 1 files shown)


gregqualls (approved):
Looks great.  adds even more clarity to the page.

PR #4713, opened 2025-06-26
Prepare docs for api docs

gilzow:
## Why

Prepares the docs sites for migration of api docs

## What's changed

* refactors api workflow to save api docs file into shared/pages
* adds stylesheet for api docs
* adds location for both docs to allow for /api.
*  adds steps to build hook to copy api docs page into new location

## Where are changes

Updates are for:

### DIFF
--- a/.platform/applications.yaml
+++ b/.platform/applications.yaml
@@ -24,11 +24,23 @@
         # Build hooks can modify the application files on disk but not access any services like databases.
         build: |
           set -e
+          #copy API docs into correct location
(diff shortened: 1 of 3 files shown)


.platform/applications.yaml
```diff
@@ -135,7 +160,10 @@
                     '^/favicon.ico$':
                         allow: true
                         passthru: '/images/favicon.ico'
-
+            '/api':
+              root: 'sites/platform/public/api'
```
flovntp: should be 
```
              root: 'sites/upsun/public/api'
```

flovntp (changes requested):

PR #4794, opened 2025-07-23
Fix a vulnerability where we would redirect to an external site

gmoigneu:


gmoigneu:
This patch can't work as is.

PR #4908, opened 2025-09-05
Add services as applications explanation and example

gmoigneu:
## Why

As we want to promote running services as applications, this PR adds a section in `/add-services` to explain how to do this and list the available examples.

## What's changed

`/add-services/index.md`

## Where are changes

Updates are for:

### DIFF
--- a/sites/upsun/src/add-services/_index.md
+++ b/sites/upsun/src/add-services/_index.md
@@ -215,6 +215,29 @@ Add them to the `type` key of the [service configuration](#1-configure-the-servi
 <!-- To update the versions in this table, use docs/data/registry.json -->
 {{% supported-services %}}
 
+### Running other services as applications
+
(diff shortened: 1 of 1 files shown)


sites/upsun/src/add-services/_index.md
```diff
@@ -215,6 +215,29 @@ Add them to the `type` key of the [service configuration](#1-configure-the-servi
 <!-- To update the versions in this table, use docs/data/registry.json -->
 {{% supported-services %}}
 
+### Running other services as applications
+
+While Upsun publishes and maintains service images for the list above, you can also run any other service as an application in your project as long as they rely on a http-based interface for their respective clients.
```
fabpot:
```suggestion
While Upsun publishes and maintains service images for the list above, you can also run any other service as an application in your project as long as they rely on an HTTP-based interface for their respective clients.
```

fabpot (changes requested):
I would only explain that we also have HTTP-based service that are configured via our multi-app feature and not explain how on can do it for their own (which should be documented separately).

PR #4958, opened 2025-09-19
Replace psh in static fetch files

gilzow:
- Updates all instances of Platform.sh --> Upsun Fixed in the `./sites/platform/static/files/fetch/` location
- Updates the image used in social related meta elements


gregqualls (approved):
looks great

PR #5113, opened 2025-10-20
Composable Image: add deprecated section + composable versions in YAML Config

flovntp:
## Why

Closes #{ISSUE_NUMBER}

## Where are changes

Updates are for:


gilzow (approved):
we should rename the `composable_versions.yaml` file to be more generic but we can do that in a later iteration.

PR #5288, opened 2025-12-17
[Meta-bot] Update service versions + PHP extensions

flovntp:
This PR is generated from Meta Version updater cached data.

- Ingest stores GitLab + upstream metadata (endoflife.date or custom) into each service.
- Export reads cache once and updates registry + php_extensions grid.

Files updated:
- `shared/data/registry.json`
- `shared/data/php_extensions.yaml`


shared/data/registry.json
```diff
@@ -1321,7 +1150,7 @@
     "docs": {
       "relationship_name": "vault-kms",
       "service_name": "vault-kms",
-      "url": "/add-services/vault.html"
+      "url": "/add-services/vault-kms.html"
```
gilzow: the file name/location for vault is /add-services/vault.html. See https://docs.upsun.com/add-services/vault.html

gilzow:
```suggestion
      "url": "/add-services/vault.html"
```

flovntp: good catch, just updated the config file and update the PR ;)

flovntp: same

PR #5375, opened 2026-01-27
Network storage updates

### LINKED ISSUE #5374: network-storage: upgrade/change/downgrade info
### Priority

3 - Not urgent - can be addressed as team bandwidth permits

### Where should the changes be made?

https://docs.upsun.com/add-services/network-storage.html
https://fixed.docs.upsun.com/add-services/network-storage.html

Version 1.0 is the recommended version. Provide a description of the different versions so customers know which one to select and understand when they should change versions.

### Optional: Additional context

_No response_

catplat:
## Why

Closes #5374 

## What's changed

- Adds more context about the network-storage service
- Updates supported/deprecated versions
- Elaborates on the warning and advises user to contact Support if they plan to migrate to 1.0

## Where are changes
https://docs.upsun.com/add-services/network-storage.html

Updates are for:

- [] platform (`sites/platform` templates)

### DIFF
--- a/shared/pages/openapispec-upsun-internal.json
+++ b/shared/pages/openapispec-upsun-internal.json
@@ -1477,6 +1477,44 @@
         }
       }
     },
+    "/subscriptions/{subscriptionId}/can-afford": {
+      "post": {
+        "tags": [
+          "Subscriptions"
+        ],
+        "summary": "Checks if the user can afford the requested resources.",
+        "operationId": "can-afford-subscription",
+        "parameters": [
+          {
+            "$ref": "#/components/parameters/subscription_id"
+          }
+        ],
+        "requestBody": {
+          "content": {
+            "application/json": {
+              "schema": {
+                "properties": {
+                  "resources": {
+                    "description": "A map of resource keys to quantities (both strings).",
+                    "type": "object",
+                    "additionalProperties": {}
+                  }
+                },
+                "type": "object"
+              }
+            }
+          }
+        },
+        "responses": {
+          "200": {
+            "description": "Proposed resource change is affordable."
+          },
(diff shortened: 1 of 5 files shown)


sites/upsun/src/add-services/network-storage.md
```diff
@@ -3,30 +3,39 @@ title: "Network Storage"
...
 When you deploy your app, you always get the latest available patches.
 
-{{< image-versions image="network-storage" status="supported" environment="grid" >}}
+- 1.0 
+
+<!-- TODO: reinstate this line when v2.0 is officially deprecated < image-versions image="network-storage" status="supported" environment="grid" > -->
+
+{{< note theme="warning" title="Migrating to version 1.0" >}}
+
+To minimize application downtime and data corruption, Upsun recommends using only version 1.0 of this service. Because **changing versions will wipe your persistent data**, we ask that you **[create a Support ticket](/learn/overview/get-support.md)** so we can help you through a smooth transition and avoid unintended data loss.
```
nicogommen: We could give a little more details imo. I think some experienced customers would manage to do the migration autonomously without creating a support ticket.
We could say something like:

> Changing the version from 2.0 to 1.0 requires a migration. Simply updating the version in your configuration will wipe your persistent data. To migrate, you need to add a second network storage service with the new version (1.0), manually copy your data to it, then remove the previous service (2.0). If you are not confortable with this process, please create a support ticket so we can help you through a smooth transition and avoid unintended data loss. 

I think this could save some tickets for our support team.

catplat: Excellent comment - thank you. Addressed - marking as resolved.

PR #5484, opened 2026-03-13
fix(mercure): use precise wording when describing routes

danmilon:
The term `APP_NAME` was misleading cause previously it refers to mercure with `SERVICE_NAME`.

## Why

The wording was a bit misleading previously, suggesting to the customer to use the app instead of the mercure service.

## Where are changes

Updates are for:


catplat (approved):

sites/upsun/src/add-services/mercure.md
flovntp, line 174:
```suggestion
      <SERVICE_NAME>: 
```

PR #5499, opened 2026-03-19
docs: add webhook authentication documentation

romainneutron:
Document the --shared-key option for webhook integrations, explaining the JWS (RFC 7515/7797) signing mechanism and providing verification examples in Python, Node.js, PHP, and Go.

## Why

Closes #{ISSUE_NUMBER}

## Where are changes

Updates are for:


sites/platform/src/integrations/activity/webhooks.md
```diff
@@ -23,6 +23,166 @@ Leave at the default values to get all events on all environments in a project.
...
+
+Note the **two dots with nothing between them** — the payload slot is empty because the payload is sent in the POST body rather than embedded in the JWS token.
+
+### Verifying the signature
+
+To verify a webhook request:
+
+1. Read the raw POST body (the JSON payload).
+2. Read the `X-JWS-Signature` header.
+3. Re-attach the payload into the JWS token (insert it between the two dots).
```
catplat:
```suggestion
3. Parse the JWS protected header and signature from the header value (the string in the form `<base64url-encoded-header>..<base64url-encoded-signature>`). Then, do one of the following:
   - Use a JWS library that supports RFC 7797 detached, unencoded payloads (`b64:false`) and pass the raw body bytes from step 1 as the detached payload.
   - Manually compute the JWS signing input as `<base64url-encoded-header>.<raw-body-bytes>` and verify the HS256 MAC over this signing input with your shared key.
```

sites/platform/src/integrations/activity/webhooks.md
```diff
@@ -23,6 +23,166 @@ Leave at the default values to get all events on all environments in a project.
...
+4. Verify the signature using HS256 with your shared key.
```
catplat:
```suggestion
4. Verify the signature using HS256 with your shared key, treating the raw body bytes from step 1 as the payload. Do **not** try to build a new compact JWS string by inserting the raw body between the two dots.
```

catplat (dismissed):
Incorporate review changes from the equivalent Flex topics in the Flex repo.

PR #5604, opened 2026-06-01
Remove tier gating dkim

matthiaz:
## Why

Having it tier gated causes more frustration from the customers and more work from support. Reasons and argumentation was discussed internally.

## What's changed

Removed the note that says this is tier gated to enterprise/elite
As discussed with @nicogommen 

## Where are changes

Updates are for:


catplat (approved):
lgtm

PR #5605, opened 2026-06-03
docs: add object storage page

ovv:
## Why

Documents the object storage feature for Upsun Flex. Users can request an
S3-compatible bucket for an app by setting an object-storage size on its
resources, but there was no user-facing page explaining how to enable, size,
connect to, or remove it.

## What's changed

Adds a new page at `sites/upsun/src/manage-resources/object-storage.md` covering:

- Enabling and sizing a bucket with `resources:set --object-storage` (in MB),
  including the step / minimum / maximum rules.
- Connecting from an app through the auto-injected `object-storage` relationship
  (`PLATFORM_RELATIONSHIPS`, path-style addressing, no credentials), with a boto3
  example.
- Sharing a bucket with workers, services, and tasks via the
  `<app>:object-storage` relationship syntax.
- Resizing and removing storage, with a note that removal is destructive.
- Current limitations (one bucket per app, internal-only endpoint, unsupported
  S3 features).

The page lives under `manage-resources` because the feature is enabled through
the same `resources:set` flow as CPU, RAM, and disk.

## Where are changes

Updates are for:

### DIFF
--- a/sites/upsun/src/manage-resources/object-storage.md
+++ b/sites/upsun/src/manage-resources/object-storage.md
@@ -1,0 +1,176 @@
+---
+title: Object storage
+description: Give an app an S3-compatible bucket by setting an object-storage size on its resources.
+weight: -40
+keywords:
+  - "object storage"
+  - "bucket"
+  - "S3"
+  - "storage"
+  - "resources"
+---
+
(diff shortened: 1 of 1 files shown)


catplat:
Closing this PR - see PR 571 in new Flex docs repo.

PR #5615, opened 2026-06-15
Fix: repository-s3 plugin supported on DG2 only

catplat:
## Why

The repository-s3 plugin on both Elasticsearch and OpenSearch is not supported on Grid/GridHA/DG3 and only possible to configure on DG2.

## What's changed

Append info that indicates the plugin is supported on DG2 only. 

## Where are changes
https://fixed.docs.upsun.com/add-services/opensearch.html#available-plugins
https://fixed.docs.upsun.com/add-services/elasticsearch.html#available-plugins

Updates are for:


catplat:
Duplicate of #5612.

PR #5685, opened 2026-07-29
Fix/preview env route urls

catplat:
## Why

Routes: Update information about how URLs are generates for preview environments. 

## Where are changes

Updates are for:


catplat:
Approved via Slack (equivalent [Flex PR](https://github.com/upsun/devcenter-mintlify/pull/662/changes))

PR #5692, opened 2026-08-07
Update DKIM request instructions in email.md

matthiaz:
Customers get asked to send a support email but not how and what they should request. I've written up a simple example so we can avoid a back and forth.


catplat (approved):
lgtm

